PR #33, opened 2021-05-03
tune data model for unified flow

ylwu-amzn:

### Description
Update data model to support unified flow
 
 
### Check List

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTask.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTask.java
@@ -30,6 +30,8 @@ import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedT
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
 
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
@@ -66,6 +68,10 @@ public class ADTask implements ToXContentObject, Writeable {
     public static final String COORDINATING_NODE_FIELD = "coordinating_node";
     public static final String WORKER_NODE_FIELD = "worker_node";
     public static final String DETECTOR_FIELD = "detector";
+    public static final String DETECTION_DATE_RANGE_FIELD = "detection_date_range";
+    public static final String ENTITY_FIELD = "entity";
+    public static final String PARENT_TASK_ID_FIELD = "parent_task_id";
+    public static final String ESTIMATED_MINUTES_LEFT_FIELD = "estimated_minutes_left";
     public static final String USER_FIELD = "user";
 
     private String taskId = null;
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTask.java
```diff
@@ -87,6 +93,10 @@
 
     private String coordinatingNode = null;
     private String workerNode = null;
+    private DetectionDateRange detectionDateRange = null;
+    private List<Entity> entity = null;
```
ohltyler, line 97: should this be called something like `entities` or `entityList`?

ylwu-amzn: Good question, this is to keep consistent with current code, check this one https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyResult.java#L91

We can rename this for all places with a dedicated PR to make the change clean and easy to review.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -179,6 +180,66 @@ public AnomalyDetector(
...
+            throw new IllegalArgumentException("Indices should be set");
+        }
+        if (detectionInterval == null) {
+            throw new IllegalArgumentException("Detection interval should be set");
+        }
+        if (shingleSize != null && shingleSize < 1) {
+            throw new IllegalArgumentException("Shingle size must be a positive integer");
+        }
+        if (categoryFields != null && categoryFields.size() > CATEGORY_FIELD_LIMIT) {
+            throw new IllegalArgumentException(CommonErrorMessages.CATEGORICAL_FIELD_NUMBER_SURPASSED + CATEGORY_FIELD_LIMIT);
```
ohltyler, line 218: minor - if we have some constant `CommonErrorMessages` used for the category field here, should we do the same for the above fields, rather than hardcoded strings? I'm thinking this could help when adding new constructors for the detector, and to make sure we're consistent & grammatically correct since some of these may be propagated on the frontend through toasts or callouts

ylwu-amzn: Good point, there is some overlap between this PR and Yizhe's validation PR  https://github.com/opensearch-project/anomaly-detection/pull/20/files#diff-ab4bf6eb04e845b21b1182e30afb84c73b282bae5cc26dbcfef2e846211bcdecR213 . I will add these error message to `CommonErrorMessages`, @yizheliu-amazon , you can make similar change in your PR or merge after this PR merged, ,

yizheliu-amazon: sure. I can merge it after this PR merged.

yizheliu-amazon (approved):
thanks for the change

ohltyler (approved):
LGTM

PR #35, opened 2021-05-05
Renaming RestAPIs while supporting backwards compatibility.

saratvemulapalli:
Signed-off-by: Sarat Vemulapalli <vemsarat@amazon.com>

### Description
Renaming RestAPIs from `_opendistro` to `_plugins`. 

- This PR replaces all existing REST APIs starting with `_opendistro` to `_plugins`.
Example:
`_opendistro/_anomaly_detection/*` -> `_plugins/_anomaly_detection/*`
- It supports backwards compatibility as old APIs still work, but a warning message in the logs is spit out explaining the path is deprecated. 
- Tested APIs with both prefixes `_opendistro` and `_plugins`. Added integration tests to verify.
  - Ideal way to test these APIs is to have Parameterized tests, i.e pass both old and new URI's to make sure all the APIs work and are backwards compatible. But there is a problem with Parameterized tests in OpenSearch Rest Test framework. https://github.com/opensearch-project/OpenSearch/issues/695

 
### Check List

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -201,8 +201,8 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
 
     private static final Logger LOG = LogManager.getLogger(AnomalyDetectorPlugin.class);
 
-    public static final String AD_BASE_LEGACY = "/_opendistro/_anomaly_detection";
-    public static final String AD_BASE_LEGACY_DETECTORS_URI = AD_BASE_LEGACY + "/detectors";
+    public static final String LEGACY_AD_BASE = "/_opendistro/_anomaly_detection";
+    public static final String LEGACY_OPENDISTRO_AD_BASE_URI = LEGACY_AD_BASE + "/detectors";
     public static final String AD_BASE_URI = "/_opensearch/_anomaly_detection";
     public static final String AD_BASE_DETECTORS_URI = AD_BASE_URI + "/detectors";
     public static final String AD_THREAD_POOL_PREFIX = "opendistro.ad.";
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -201,7 +201,9 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
 
     private static final Logger LOG = LogManager.getLogger(AnomalyDetectorPlugin.class);
 
-    public static final String AD_BASE_URI = "/_opendistro/_anomaly_detection";
+    public static final String AD_BASE_LEGACY = "/_opendistro/_anomaly_detection";
+    public static final String AD_BASE_LEGACY_DETECTORS_URI = AD_BASE_LEGACY + "/detectors";
+    public static final String AD_BASE_URI = "/_opensearch/_anomaly_detection";
     public static final String AD_BASE_DETECTORS_URI = AD_BASE_URI + "/detectors";
     public static final String AD_THREAD_POOL_PREFIX = "opendistro.ad.";
     public static final String AD_THREAD_POOL_NAME = "ad-threadpool";
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -203,7 +203,7 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
 
     public static final String LEGACY_AD_BASE = "/_opendistro/_anomaly_detection";
     public static final String LEGACY_OPENDISTRO_AD_BASE_URI = LEGACY_AD_BASE + "/detectors";
-    public static final String AD_BASE_URI = "/_opensearch/_anomaly_detection";
+    public static final String AD_BASE_URI = "/_plugins/_anomaly_detection";
     public static final String AD_BASE_DETECTORS_URI = AD_BASE_URI + "/detectors";
     public static final String AD_THREAD_POOL_PREFIX = "opendistro.ad.";
     public static final String AD_THREAD_POOL_NAME = "ad-threadpool";
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
@@ -70,14 +70,14 @@ public abstract class AbstractSearchAction<T extends ToXContentObject> extends B
     private final String index;
     private final Class<T> clazz;
     private final List<String> urlPaths;
-    private final List<Pair<String, String>> replacedPaths;
+    private final List<Pair<String, String>> deprecatedPaths;
     private final ActionType<SearchResponse> actionType;
 
     private final Logger logger = LogManager.getLogger(AbstractSearchAction.class);
 
     public AbstractSearchAction(
         List<String> urlPaths,
-        List<Pair<String, String>> replacedPaths,
+        List<Pair<String, String>> deprecatedPaths,
         String index,
         Class<T> clazz,
         ActionType<SearchResponse> actionType
@@ -85,7 +85,7 @@ public abstract class AbstractSearchAction<T extends ToXContentObject> extends B
         this.index = index;
         this.clazz = clazz;
         this.urlPaths = urlPaths;
-        this.replacedPaths = replacedPaths;
+        this.deprecatedPaths = deprecatedPaths;
         this.actionType = actionType;
     }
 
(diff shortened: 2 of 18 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -201,7 +201,9 @@
 
     private static final Logger LOG = LogManager.getLogger(AnomalyDetectorPlugin.class);
 
-    public static final String AD_BASE_URI = "/_opendistro/_anomaly_detection";
+    public static final String AD_BASE_LEGACY = "/_opendistro/_anomaly_detection";
```
dblock, line 204: I was wondering whether these should be `AD_BASE_LEGACY` or `LEGACY_AD_BASE` and I decided in https://github.com/opensearch-project/job-scheduler/pull/19/files to make it `LEGACY_AD_BASE` because I wanted those to be grouped together. Furthermore I think these should include OPENDISTRO in them. So to be consistent I propose that `AD_BASE_URI` becomes `LEGACY_OPENDISTRO_AD_BASE_URI`. 

Whether we do  `LEGACY_OPENDISTRO_AD_BASE_URI` or `AD_BASE_URI_LEGACY_OPENDISTRO` is not important to me, but let's decide the convention now?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
```diff
@@ -68,15 +68,15 @@
...
     private final String index;
     private final Class<T> clazz;
-    private final String urlPath;
+    private final List<String> urlPaths;
     private final ActionType<SearchResponse> actionType;
 
     private final Logger logger = LogManager.getLogger(AbstractSearchAction.class);
 
-    public AbstractSearchAction(String urlPath, String index, Class<T> clazz, ActionType<SearchResponse> actionType) {
+    public AbstractSearchAction(List<String> urlPaths, String index, Class<T> clazz, ActionType<SearchResponse> actionType) {
```
dblock, line 76: Add a constructor that just takes `String` now to avoid future callers adding new routes to have to pass a list.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestIndexAnomalyDetectorAction.java
```diff
@@ -152,6 +152,13 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
     public List<Route> routes() {
         return ImmutableList
             .of(
+                // Create (legacy)
```
dblock, line 155: I think we want a new `Route` constructor that takes multiple paths here to make it easier on the plugin authors.

So 

```
new Route(
  RestRequest.Method.PUT,
  ImmutableList.of(
    String.format(Locale.ROOT, "%s/{%s}", AnomalyDetectorPlugin.AD_BASE_LEGACY_DETECTORS_URI, DETECTOR_ID),
    String.format(Locale.ROOT, "%s/{%s}", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, DETECTOR_ID)
), 
 ...
```

dblock (commented):
This works, but I think it's a bit too much copy-paste. I suggest modifying/wrapping/extending `Route` to either allow multiple paths in the constructor, or extending `Route` with a static method that takes a method, and a list of paths, to return a list of routes.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
saratvemulapalli: I actually like prefixing `LEGACY` it makes more sense rather than postfixing it :). 
Sure definitely, lets do `LEGACY_OPENDISTRO_AD_BASE_URI` it makes more sense to me.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
saratvemulapalli: Sure, I like that!

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestIndexAnomalyDetectorAction.java
saratvemulapalli: I understand we are duplicating code.
Looking at `Route` its seems like it was built to handle only a single path. It would need some changes across OpenSearch to accept multiple paths and also probably it could be confusing to have two different paths in a single Route. 

May what I could do is have util function which will take a list of base path's and API paths, return a list of routes.
What do you think?

dblock: Totally agree. A static method on `Route` that returns a list of routes sounds from a matrix of paths and methods sounds right!

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
```diff
@@ -68,15 +69,23 @@
 
     private final String index;
     private final Class<T> clazz;
-    private final String urlPath;
+    private final List<String> urlPaths;
+    private final List<Pair<String, String>> replacedPaths;
```
dblock, line 73: These should be called `deprecatedPaths` I think.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestDeleteAnomalyDetectorAction.java
```diff
@@ -78,12 +78,19 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
 
     @Override
     public List<Route> routes() {
+        return ImmutableList.of();
```
dblock, line 81: Can we sink this into a base class and avoid the copy-paste?

dblock (commented):
I think "replaced" is not a great name, maybe "deprecated"? Otherwise looks ok.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestDeleteAnomalyDetectorAction.java
saratvemulapalli: Thats a great idea. I can add this in OpenSearch in the base class. 
I'll do a quick follow up after this PR.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
dblock: And these are also `urlPaths`, right? So maybe `deprecatedUrlPaths`?

saratvemulapalli: Yeah definitely, nice catch :)

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
saratvemulapalli: After looking at the updated changes, I think future callers can use a list but pass in one item if needed. What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestDeleteAnomalyDetectorAction.java
saratvemulapalli: Created an issue to take care of this: https://github.com/opensearch-project/OpenSearch/issues/694

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
```diff
@@ -126,6 +135,26 @@ public RestResponse buildResponse(SearchResponse response) throws Exception {
...
+        List<Route> routes = new ArrayList<>();
+        for (String path : urlPaths) {
+            routes.add(new Route(RestRequest.Method.POST, path));
+            routes.add(new Route(RestRequest.Method.GET, path));
+        }
+        return routes;
+    }
+
+    @Override
+    public List<ReplacedRoute> replacedRoutes() {
```
ylwu-amzn, line 147: I think we will deprecate the old `opendistro` path in future. Should we put the old paths in `deprecatedRoutes` ? Haven't used the `replacedRoutes` and `deprecatedRoutes`. Just asking to understand the differences.

saratvemulapalli: Yeah the difference between them is that `deprecatedRoute` is only for routes which are deprecated but have no replacement. `replacedRoute` Is used for routes which are replaced with a new path while deprecating the old one.

ylwu-amzn: Got it. Will user see some warning if they use the old path?

dblock: https://github.com/opensearch-project/opensearch-docs/blob/f1f6e7109f2b5312eb5dd47c4025b7373683db7a/server/src/main/java/org/elasticsearch/rest/RestHandler.java#L89

dblock: Yes, you'll see a deprecation warning from [here](https://github.com/opensearch-project/opensearch-docs/blob/f1f6e7109f2b5312eb5dd47c4025b7373683db7a/server/src/main/java/org/elasticsearch/rest/RestController.java#L150).

dblock:
> I think "replaced" is not a great name, maybe "deprecated"? Otherwise looks ok.

I wrote this when I didn't know there was support for this already in OpenSearch.

dblock (approved):

dblock:
LGTM, but would wait for @ylwu-amzn to re-review

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
ylwu-amzn: Nice. Thanks. Have no permission to check these two links shared by @dblock. But it's ok if we have such warning.

ylwu-amzn (approved):
LGTM. Thanks for the change.

PR #43, opened 2021-05-14
    Rename namespaces from opendistro to opensearch.

spbjss:
Signed-off-by: Alex <pengsun@amazon.com>

### Description
Rename namespaces & other references of opendistro to opensearch
 
### Issues Resolved
Update the namespaces to adopt the opensearch GA.
 
### Check List
- [Y] Commits are signed per the DCO using --signoff 


src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -56,23 +73,6 @@
...
-import com.amazon.opendistroforelasticsearch.ad.model.IntervalTimeConfiguration;
-import com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings;
-import com.amazon.opendistroforelasticsearch.ad.transport.AnomalyResultAction;
-import com.amazon.opendistroforelasticsearch.ad.transport.AnomalyResultRequest;
-import com.amazon.opendistroforelasticsearch.ad.transport.AnomalyResultResponse;
-import com.amazon.opendistroforelasticsearch.ad.transport.AnomalyResultTransportAction;
-import com.amazon.opendistroforelasticsearch.ad.transport.handler.AnomalyIndexHandler;
-import com.amazon.opendistroforelasticsearch.ad.transport.handler.DetectionStateHandler;
-import com.amazon.opendistroforelasticsearch.ad.util.ClientUtil;
 import com.amazon.opendistroforelasticsearch.commons.InjectSecurity;
```
ohltyler, line 76: Should we wait for the AD dependencies (`common-utils`, `job-scheduler`) to make the corresponding namespace changes first? Then we can remove all references to `opendistroforelasticsearch`?

spbjss: Good point! 
But considering the tight launch schedule, I prefer to change as much as we could before the AD dependencies were ready. Once the dependencies were ready, we could make another small change to remove all references to opendistroforelasticsearch.

ylwu-amzn: line 76 to 83 of this file still using `com.amazon.opendistroforelasticsearch.*`  as `common-utils` and `job-scheduler` not changed yet. 
I think it's ok to change it later when `common-utils` and `job-scheduler` ready. So we can make sure AD build can pass.

ylwu-amzn (approved):
LGTM. Thanks for the change!

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
ohltyler: Cool, it can be fixed after. Created #45 to track.

ohltyler (approved):
Thanks for the changes!

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
dblock: https://github.com/opensearch-project/common-utils/issues/19 and https://github.com/opensearch-project/job-scheduler/issues/22 are being taken of

Issue #40, opened 2021-05-12
Ensure backwards compatibility with ODFE


dblock:
Coming from [opensearch-plugins#12](https://github.com/opensearch-project/opensearch-plugins/issues/12), enable upgrade to OpenSearch, while supporting backward compatibility.

- [x] Rename Namespaces (e.g. `com.amazon.opendistroforelasticsearch` to `org.opensearch`)
- [x] Rename Classes, Methods, Variables 
- [x] Rename Remaining Identifiers (e.g. `Opendistro` to `OpenSearch`).
- [x] Rename, and Backwards Compatibility for Rest APIs
- [x] Rename, and Backwards Compatibility for Settings 
- [ ] Rename, and Backwards Compatibility for Indices
- [ ] Run in a backwards compatible way on top of OpenSearch 1.0 that has joined an ES 7.10.x cluster 
- [ ] Run in a backwards compatible way on top of OpenSearch 1.0 that has joined an ODFE 1.13.x cluster 
- [ ] Drop in replacement for the Opendistro version 1.13 of the plugin

PR #49, opened 2021-05-18
WIP: Alias AD index.

dblock:
Signed-off-by: dblock <dblock@amazon.com>

### Description

In progress #40 on indices. 

I am not in love with where this is going. Too much cruft that needs to be refactored so that it's a lot easier to reference an index.
  
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
+++ b/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
@@ -54,6 +54,8 @@ import org.apache.logging.log4j.message.ParameterizedMessage;
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.admin.cluster.state.ClusterStateRequest;
 import org.opensearch.action.admin.indices.alias.Alias;
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest;
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest.AliasActions;
 import org.opensearch.action.admin.indices.alias.get.GetAliasesRequest;
 import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.admin.indices.create.CreateIndexResponse;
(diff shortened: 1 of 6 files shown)


skkosuri-amzn:
@dblock To achive zero downtime, this approach looks good. One additional thing we need to manage: plugin indices are configured as [system-indices](https://opendistro.github.io/for-elasticsearch-docs/docs/security/configuration/system-indices/) In `hybrid` mode these systems must point to legacy indices names, in complete new opensearch install these need to point to new indices names. We can't use alias for this configuration. In new install these are deployed to opensearch.yml from here: https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L376

dblock:
> @dblock To achive zero downtime, this approach looks good. One additional thing we need to manage: plugin indices are configured as [system-indices](https://opendistro.github.io/for-elasticsearch-docs/docs/security/configuration/system-indices/) In `hybrid` mode these systems must point to legacy indices names, in complete new opensearch install these need to point to new indices names. We can't use alias for this configuration. In new install these are deployed to opensearch.yml from here: https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L376

Are you saying that we cannot use alias for system indices because the alias doesn't follow the same permissions model? Does it not? Or is it something else?

skkosuri-amzn:
> > @dblock To achive zero downtime, this approach looks good. One additional thing we need to manage: plugin indices are configured as [system-indices](https://opendistro.github.io/for-elasticsearch-docs/docs/security/configuration/system-indices/) In `hybrid` mode these systems must point to legacy indices names, in complete new opensearch install these need to point to new indices names. We can't use alias for this configuration. In new install these are deployed to opensearch.yml from here: https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L376
> 
> Are you saying that we cannot use alias for system indices because the alias doesn't follow the same permissions model? Does it not? Or is it something else?

In hybrid (odfe + opensearch clulster) mode system-indices config must have LEGACY_INDICES and not alias. 
OpenSearch only config must have NEW_INDICES_NAMES. (no alias are created on this mode anyways).

dblock:
Abandoning this approach, closing.

PR #51, opened 2021-05-20
support preview with detector directly

ylwu-amzn:

### Description
In unified flow, user can preview detector before creating. That mean user can pass detector configuration to preview without detector id.

This PR add new URL path `_opendistro/_anomaly_detection/detectors/_preview`. User can preview existing detector as well.

Examples
```
# preview detector directly without detector id
POST _opendistro/_anomaly_detection/detectors/_preview
{
    "period_start": 1612982516000,
    "period_end": 1614278539000,
    "detector": {
        "name": "test-detector",
        "description": "test",
        "time_field": "timestamp",
        "indices": [
            "test_data"
        ],
        "detection_interval": {
            "period": {
                "interval": 1,
                "unit": "Minutes"
            }
        },
        "feature_attributes": [
            {
                "feature_name": "feature1",
                "feature_enabled": true,
                "aggregation_query": {
                    "feature1": {
                        "sum": {
                            "field": "value"
                        }
                    }
                }
            }
        ]
    }
}

# preview existing detector with detector id
POST _opendistro/_anomaly_detection/detectors/_preview
{
    "detector_id": "sYkUvHcBiZv51f-Lv8QN",
    "period_start": 1612982516000,
    "period_end": 1614278539000
}
```

To support backward compatibility, we still keep the old url `POST _opendistro/_anomaly_detection/detectors/<detector_id>/_preview`. So user can preview existing detector in the old way, for example 
```
POST _opendistro/_anomaly_detection/detectors/sYkUvHcBiZv51f-Lv8QN/_preview
{
    "period_start": 1612982516000,
    "period_end": 1614278539000
}
```

 
### Check List

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
@@ -31,7 +31,7 @@ import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedT
 import java.io.IOException;
 import java.time.Instant;
 
-import org.apache.logging.log4j.util.Strings;
+import org.opensearch.common.Strings;
 import org.opensearch.common.xcontent.ToXContentObject;
 import org.opensearch.common.xcontent.XContentBuilder;
 import org.opensearch.common.xcontent.XContentParser;
@@ -107,7 +107,7 @@ public class AnomalyDetectorExecutionInput implements ToXContentObject {
                     break;
             }
         }
-        if (Strings.isNotEmpty(adId)) {
+        if (!Strings.isNullOrEmpty(adId)) {
             detectorId = adId;
         }
         return new AnomalyDetectorExecutionInput(detectorId, periodStart, periodEnd, detector);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
@@ -31,6 +31,7 @@ import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedT
 import java.io.IOException;
 import java.time.Instant;
 
+import org.apache.logging.log4j.util.Strings;
 import org.opensearch.common.xcontent.ToXContentObject;
 import org.opensearch.common.xcontent.XContentBuilder;
 import org.opensearch.common.xcontent.XContentParser;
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
```diff
@@ -31,6 +31,7 @@
 import java.io.IOException;
 import java.time.Instant;
 
+import org.apache.logging.log4j.util.Strings;
```
spbjss, line 34: Are we using the log4j util.Strings for not empty check? I guess it suppose to be org.opensearch.common.Strings?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
```diff
@@ -98,6 +107,9 @@ public static AnomalyDetectorExecutionInput parse(XContentParser parser, String
                     break;
             }
         }
+        if (Strings.isNotEmpty(adId)) {
+            detectorId = adId;
+        }
```
spbjss, line 112: this looks a little bit wired, if the adId == "", then we will apply null to detectorId. Could you explain the difference of "" and null here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
ylwu-amzn: Good catch, will fix it

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorExecutionInput.java
ylwu-amzn: No difference, this is to prevent some user wrong input. The detector id should not be null or "".  If the user input "" , we still consider that no detector id provided.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -117,8 +117,8 @@ private AnomalyDetectorExecutionInput getAnomalyDetectorExecutionInput(RestReque
     }
 
     private String validateAdExecutionInput(AnomalyDetectorExecutionInput input) {
-        if (StringUtils.isBlank(input.getDetectorId())) {
-            return "Must set anomaly detector id";
+        if (StringUtils.isBlank(input.getDetectorId()) && input.getDetector() == null) {
```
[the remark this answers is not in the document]
ylwu-amzn: As said in the description 
> In unified flow, user can preview detector before creating. That mean user can pass detector configuration to preview without detector id.

spbjss (approved):
Looks good to me.

PR #59, opened 2021-05-22
Add missed settings update consumer

spbjss:
### Description
Add missed settings update consumer
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/40
 
### Check List


src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -75,9 +76,10 @@
     private CheckpointDao checkpointDao;
     private CacheProvider cache;
     private final NodeStateManager stateManager;
-    private final int coolDownMinutes;
+    private int coolDownMinutes;
```
ylwu-amzn, line 79: Will the update consumer work by using "private int" ? I think we should use "private volatile int"

spbjss: Thanks! will update in another commit.

ylwu-amzn (approved):
LGTM. Thanks for the change!

src/main/java/org/opensearch/ad/transport/BackPressureRouting.java
```diff
@@ -45,18 +46,23 @@
...
 
-    public BackPressureRouting(String nodeId, Clock clock, Settings settings) {
+    public BackPressureRouting(String nodeId, Clock clock, Settings settings, ClusterService clusterService) {
         this.nodeId = nodeId;
         this.clock = clock;
         this.backpressureCounter = new AtomicInteger(0);
         this.maxRetryForUnresponsiveNode = MAX_RETRY_FOR_UNRESPONSIVE_NODE.get(settings);
         this.mutePeriod = BACKOFF_MINUTES.get(settings);
         this.lastMuteTime = 0;
+        this.clusterService = clusterService;
```
kaituo, line 62: Have you tried to verify dynamically changing max retry for unresponsive node works?  You need to simulate a node is not responsive by changing the code to reject every incoming request and check if the node will be filtered out after the set unresponsive times?

kaituo: This applies to all settings you have changed.

PR #65, opened 2021-05-24
Add queues for concurrent requests

kaituo:
Note: since there are a lot of dependencies, I only list the main class and test code to save reviewers' time. The build will fail due to missing dependencies. I will use that PR just for review. will not merge it. Will have a big one in the end and merge once after all review PRs get approved.  Now the code is missing unit tests.  Posting PRs now to meet the cutoff date (June 1).  Will add unit tests, run performance tests, and fix bugs before the official release.

### Description
This PR adds the ability for us to run concurrent requests (either batch or single request).  The concurrency is configurable. The callers use the put method to put requests in and the queue tries to execute them if there are concurrency slots.

Testing done:
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ratelimit/BatchQueue.java
+++ b/src/main/java/org/opensearch/ad/ratelimit/BatchQueue.java
@@ -1,0 +1,138 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.ratelimit;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.util.List;
+import java.util.Random;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.action.ActionListener;
+import org.opensearch.action.support.ThreadedActionListener;
+import org.opensearch.ad.AnomalyDetectorPlugin;
+import org.opensearch.ad.NodeStateManager;
+import org.opensearch.ad.breaker.ADCircuitBreakerService;
+import org.opensearch.ad.util.ClientUtil;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.threadpool.ThreadPool;
+
+/**
+ *
+ * @param <RequestType> Individual request type that is a subtype of ADRequest
+ * @param <BatchRequestType> Batch request type like BulkRequest
+ * @param <BatchResponseType> Response type like BulkResponse
+ */
+public abstract class BatchQueue<RequestType extends QueuedRequest, BatchRequestType, BatchResponseType> extends
+    ConcurrentQueue<RequestType> {
+    private static final Logger LOG = LogManager.getLogger(BatchQueue.class);
+    protected int batchSize;
+
+    public BatchQueue(
+        String queueName,
+        long heapSize,
+        int singleRequestSize,
+        Setting<Float> maxHeapPercentForQueueSetting,
+        ClusterService clusterService,
+        Random random,
+        ADCircuitBreakerService adCircuitBreakerService,
+        ThreadPool threadPool,
+        Settings settings,
+        float maxQueuedTaskRatio,
+        Clock clock,
+        float mediumSegmentPruneRatio,
+        float lowSegmentPruneRatio,
+        int maintenanceFreqConstant,
+        ClientUtil clientUtil,
+        Setting<Integer> concurrencySetting,
+        Duration executionTtl,
+        Setting<Integer> batchSizeSetting,
+        Duration stateTtl,
+        NodeStateManager nodeStateManager
+    ) {
+        super(
+            queueName,
+            heapSize,
+            singleRequestSize,
+            maxHeapPercentForQueueSetting,
+            clusterService,
+            random,
+            adCircuitBreakerService,
+            threadPool,
+            settings,
+            maxQueuedTaskRatio,
+            clock,
+            mediumSegmentPruneRatio,
+            lowSegmentPruneRatio,
+            maintenanceFreqConstant,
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    @Override
+    protected void triggerProcess() {
+        threadPool.executor(AnomalyDetectorPlugin.AD_THREAD_POOL_NAME).execute(() -> {
+            if (permits.tryAcquire()) {
+                try {
+                    lastExecuteTime = clock.instant();
+                    execute(() -> {
+                        permits.release();
+                        process();
```
jmazanec15, line 145: I am a little confused here. The call to `process()` will call `triggerProcess` again and this will continue until the permit can no longer be acquired?

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Override
+    public void maintenance() {
+        super.maintenance();
+
+        if (lastExecuteTime.plus(executionTtl).isBefore(clock.instant())) {
+            LOG.warn("previous execution has been running for too long.  Maybe there are bugs.");
+
+            // release one permit
+            permits.release();
```
jmazanec15, line 130: Why do we release one permit here?

src/main/java/org/opensearch/ad/ratelimit/BatchQueue.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Used by subclasses to creates customized logic to send batch requests.
+     * After everything finishes, the method should call listener.
+     * @param request Batch request to execute
+     * @param listener customized listener
+     */
+    protected abstract void executeBatchRequest(BatchRequestType request, ActionListener<BatchResponseType> listener);
+
+    /**
+     * We convert from queued requests understood by AD to batchRequest understood by ES.
```
jmazanec15, line 99: ES -> OpenSearch

kaituo: changed

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
kaituo: Please check the comment at line 127.

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
kaituo: This is a callback after processing one request.  So we keep looking for new requests if there is any after finishing one request.  Otherwise, just release (the 2nd callback).  Added comment to explain.

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
jmazanec15: Right, but why 1 permit and not multiple?

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
jmazanec15: I see that makes sense.

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
```diff
@@ -0,0 +1,162 @@
...
+            clock,
+            mediumSegmentPruneRatio,
+            lowSegmentPruneRatio,
+            maintenanceFreqConstant,
+            stateTtl,
+            nodeStateManager
+        );
+
+        this.permits = new Semaphore(concurrencySetting.get(settings));
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(concurrencySetting, it -> permits = new Semaphore(it));
```
jmazanec15: Is there a possibility where the same concurrencySetting is passed in for 2 separate queue? If so, is this call safe?

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
kaituo: because one request uses one permit.  After the request finished, we need to release one.

src/main/java/org/opensearch/ad/ratelimit/ConcurrentQueue.java
kaituo: Our queues are singletons. So it is safe here.  It seems everybody did it this way.  Not sure how ES handles the race conditions of setting change.  There is a complicated protocol to ensure consistency and roll back if not from my read of the related code briefly.

jmazanec15: I see. So the concurrency setting will be unique for each queue?

kaituo: yes

jmazanec15 (approved):

src/main/java/org/opensearch/ad/ratelimit/ConcurrentWorker.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /**
+     * try to execute queued requests if there are concurrency slots and return right away.
+     */
+    @Override
+    protected void triggerProcess() {
+        threadPool.executor(AnomalyDetectorPlugin.AD_THREAD_POOL_NAME).execute(() -> {
+            if (permits.tryAcquire()) {
```
jngz-es, line 142: Just a question, the number of semaphores should be less than or equal to the number of threads, right?

jngz-es (approved):

src/main/java/org/opensearch/ad/ratelimit/ConcurrentWorker.java
kaituo: seems not. I have seen some well-known project using 10k semaphores to represent total number of tokens that account for total heap memory available. Each mb is 1 semaphore.

I can have 100 semaphores with 1 thread.  If all of the code are asynchronous, those 100 operations (protected by the semaphores) can still run concurrently as long as they don't use the thread at the same time.

PR #67, opened 2021-05-25
Add a queue for HCAD model cold start.

kaituo:

### Description
This PR adds a queue for HCAD model training (a.k.a. cold start). As model training is a pretty expensive operation, we pull cold start requests from the queue in a serial fashion. Each detector has an equal chance of being pulled. The equal probability is achieved by putting model training requests for different detectors into different segments and pulling requests from segments in a round-robin fashion.

To accommodate the model training queue, we need to add a listener as an input to EntityColdStarter's trainModel method, the utility class for model training. The listener needs to be invoked whenever EntityColdStarter finishes training or encounters exceptions. The listener helps notify the cold start queue to pull another request (if any) to execute.

Some of the rate-limiting logic in EntityColdStarter is removed as we have a cold start queue now.

We have also changed to use DoorKeeper to limit the cold start an entity can trigger in a certain period.  DoorKeeper is a bloom filter that is used elsewhere in the package. Reusing the logic reduces the amount of code we have to maintain. The DoorKeeper needs to be constantly maintained. Thus, I let EntityColdStarter implement MaintenanceState and make CronTransportAction clean DoorKeeper regularly.

Testing done:
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -64,8 +64,8 @@ import org.opensearch.ad.feature.SearchFeatureDao;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.model.IntervalTimeConfiguration;
-import org.opensearch.ad.ratelimit.CheckpointWriteQueue;
-import org.opensearch.ad.ratelimit.SegmentPriority;
+import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
+import org.opensearch.ad.ratelimit.RequestPriority;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.threadpool.ThreadPool;
@@ -101,9 +101,10 @@ public class EntityColdStarter implements MaintenanceState {
     private int coolDownMinutes;
     // A bloom filter checked before cold start to ensure we don't repeatedly
     // retry cold start of the same model.
+    // keys are detector ids.
     private Map<String, DoorKeeper> doorKeepers;
     private final Duration modelTtl;
-    private final CheckpointWriteQueue checkpointWriteQueue;
+    private final CheckpointWriteWorker checkpointWriteQueue;
 
     /**
      * Constructor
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -94,9 +98,12 @@
     private final int shingleSize;
     private Instant lastThrottledColdStartTime;
     private final FeatureManager featureManager;
-    private final Cache<String, Instant> lastColdStartTime;
-    private final CheckpointDao checkpointDao;
     private int coolDownMinutes;
+    // A bloom filter checked before cold start to ensure we don't repeatedly
```
jmazanec15, line 102: Minor: The DoorKeeper is the bloom filter. Add comment saying keys are detector ids.

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartQueue.java
```diff
@@ -0,0 +1,130 @@
...
+            modelId.get(),
+            detectorId,
+            ModelType.ENTITY.getName(),
+            clock,
+            0
+        );
+
+        ActionListener<Void> failureListener = ActionListener.delegateResponse(listener, (delegateListener, e) -> {
+            if (ExceptionUtil.isOverloaded(e)) {
+                LOG.error("ES is overloaded");
```
jmazanec15, line 121: ES -> OpenSearch

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -121,10 +128,11 @@
      * @param thresholdDownsamples the number of samples to keep during downsampling
      * @param thresholdMaxSamples the max number of samples before downsampling
      * @param featureManager Used to create features for models.
-     * @param lastColdStartTimestampTtl max time to retain last cold start timestamp
-     * @param maxCacheSize max cache size
-     * @param checkpointDao utility to interact with the checkpoint index
      * @param settings ES settings accessor
+     * @param modelTtl time-to-leave before last access time of the cold start cache.
```
jmazanec15, line 132: Time to live, no?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: added

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: yeah, changed

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartQueue.java
kaituo: changed

jmazanec15 (approved):

jngz-es (approved):

PR #68, opened 2021-05-25
Add a queue for checkpoint read

kaituo:

### Description
This PR adds a queue for loading model checkpoint.  The read is a multi-get query. Possible results are:
* If a checkpoint is not found, we forward that request to cold start queue.
*When a request gets errors, the queue does not change its expiry time and puts that request to the end of the queue and automatically retries them before they expire.
* When a checkpoint is found, we load that point to memory and score the input data point and save the result if a complete model exists. Otherwise, we enqueue the sample.  If we can host that model in memory (e.g., there is enough memory), we put the loaded model to cache.  Otherwise (e.g., a cold entity), we write the updated checkpoint back to disk.

Testing done:

 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
+++ b/src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
@@ -1,0 +1,374 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.ratelimit;
+
+import static org.opensearch.ad.settings.AnomalyDetectorSettings.CHECKPOINT_READ_QUEUE_BATCH_SIZE;
+import static org.opensearch.ad.settings.AnomalyDetectorSettings.CHECKPOINT_READ_QUEUE_CONCURRENCY;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayDeque;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.Random;
+import java.util.Set;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
+import org.opensearch.action.ActionListener;
+import org.opensearch.action.get.MultiGetAction;
+import org.opensearch.action.get.MultiGetItemResponse;
+import org.opensearch.action.get.MultiGetRequest;
+import org.opensearch.action.get.MultiGetResponse;
+import org.opensearch.ad.NodeStateManager;
+import org.opensearch.ad.breaker.ADCircuitBreakerService;
+import org.opensearch.ad.caching.CacheProvider;
+import org.opensearch.ad.constant.CommonName;
+import org.opensearch.ad.indices.ADIndex;
+import org.opensearch.ad.indices.AnomalyDetectionIndices;
+import org.opensearch.ad.ml.CheckpointDao;
+import org.opensearch.ad.ml.EntityColdStarter;
+import org.opensearch.ad.ml.EntityModel;
+import org.opensearch.ad.ml.ModelManager;
+import org.opensearch.ad.ml.ModelManager.ModelType;
+import org.opensearch.ad.ml.ModelState;
+import org.opensearch.ad.ml.ThresholdingResult;
+import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.ad.model.AnomalyResult;
+import org.opensearch.ad.model.Entity;
+import org.opensearch.ad.util.ClientUtil;
+import org.opensearch.ad.util.ExceptionUtil;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
```diff
@@ -0,0 +1,374 @@
...
+                // retry all of them
+                super.putAll(toProcess);
+            } else {
+                LOG.error("Fail to restore models", exception);
+            }
+        });
+    }
+
+    private void processCheckpointIteration(
+        int i,
```
jmazanec15, line 227: I think this name be more descriptive.

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
```diff
@@ -0,0 +1,374 @@
...
+            if (checkpointResponse != null) {
+                // successful requests
+                Optional<Map<String, Object>> checkpointString = checkpointDao.processRawCheckpoint(checkpointResponse.getResponse());
+                if (checkpointString.isPresent()) {
+                    // checkpoint exists
+                    Optional<Entry<EntityModel, Instant>> checkpoint = checkpointDao
+                        .fromEntityModelCheckpoint(checkpointString.get(), modelId);
+
+                    if (false == checkpoint.isPresent()) {
+                        // checkpoint is too big
```
jmazanec15, line 263: Is this the only case where the checkpoint isnt present?

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
```diff
@@ -0,0 +1,374 @@
...
+
+    @Override
+    protected MultiGetRequest toBatchRequest(List<EntityFeatureRequest> toProcess) {
+        MultiGetRequest multiGetRequest = new MultiGetRequest();
+        for (EntityRequest request : toProcess) {
+            Optional<String> modelId = request.getModelId();
+            if (false == modelId.isPresent()) {
+                continue;
+            }
+            multiGetRequest.add(new MultiGetRequest.Item(CommonName.CHECKPOINT_INDEX_NAME, modelId.get()));
```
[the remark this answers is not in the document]
kaituo: It is imposed here: https://github.com/kaituo/anomaly-detection-1/blob/multiCat4/src/main/java/org/opensearch/ad/ratelimit/RateLimitedRequestWorker.java#L387

So toBatchRequest method can take the input and send the multi-get directly.

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
```diff
@@ -0,0 +1,374 @@
...
+                        // submit to cold start queue
+                        entityColdStartQueue.put(origRequest);
+                    }
+                }
+            }
+
+            processCheckpointIteration(0, toProcess, successfulRequests, retryableRequests);
+        }, exception -> {
+            if (ExceptionUtil.isOverloaded(exception)) {
+                LOG.error("too many get AD model checkpoint requests or shard not avialble");
```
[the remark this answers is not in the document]
kaituo: I am checking the exception is UnavailableShardsException.  I observed that can happen when the system cannot respond to requests in the performance tests.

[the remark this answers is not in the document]
kaituo: yes, fixed avialble.

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
```diff
@@ -0,0 +1,374 @@
...
+                setCoolDownStart();
+            } else if (ExceptionUtil.isRetryAble(exception)) {
+                // retry all of them
+                super.putAll(toProcess);
```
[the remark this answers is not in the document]
kaituo: yeah, removed

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
kaituo: Please check https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

"Note that certain universally-known abbreviations are OK, such as i for an iteration variable and T for a template parameter."

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadQueue.java
kaituo: yes it is so far

jmazanec15 (approved):

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadWorker.java
```diff
@@ -0,0 +1,383 @@
...
+            for (MultiGetItemResponse itemResponse : itemResponses) {
+                String modelId = itemResponse.getId();
+                if (itemResponse.isFailed()) {
+                    final Exception failure = itemResponse.getFailure().getFailure();
+                    if (failure instanceof IndexNotFoundException) {
+                        for (EntityRequest origRequest : toProcess) {
+                            // submit to cold start queue
+                            entityColdStartQueue.put(origRequest);
+                        }
+                        return;
```
jngz-es, line 189: I am not quite sure, is it return? not continue? I saw there is for loop outside.

kaituo: It is return.  If it is checkpoint index not found exception, I don't need to retry as checkpoint read is bound to fail. Just send everything to the cold start queue and return. Added comments to explain.

jngz-es (approved):

PR #69, opened 2021-05-25
Add queues for checkpoint and result write

kaituo:

### Description
This PR adds queues for writing model checkpoints and detection result.  The write is a batch request. 

Failure tolerance like retrying requests is handled. To retry failed result bulk requests, we also added the index request to retry in ADResultBulkResponse.  

Testing done:
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
@@ -26,24 +26,20 @@
 
 package org.opensearch.ad.ml;
 
+import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.time.Clock;
-import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.Arrays;
-import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -51,11 +47,7 @@ import org.apache.logging.log4j.message.ParameterizedMessage;
 import org.opensearch.ExceptionsHelper;
 import org.opensearch.ResourceAlreadyExistsException;
 import org.opensearch.action.ActionListener;
-import org.opensearch.action.DocWriteRequest;
-import org.opensearch.action.bulk.BulkAction;
 import org.opensearch.action.bulk.BulkItemResponse;
-import org.opensearch.action.bulk.BulkRequest;
-import org.opensearch.action.bulk.BulkResponse;
 import org.opensearch.action.delete.DeleteRequest;
 import org.opensearch.action.delete.DeleteResponse;
 import org.opensearch.action.get.GetRequest;
@@ -66,7 +58,7 @@ import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.ad.constant.CommonName;
 import org.opensearch.ad.indices.ADIndex;
 import org.opensearch.ad.indices.AnomalyDetectionIndices;
-import org.opensearch.ad.util.BulkUtil;
+import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.util.ClientUtil;
 import org.opensearch.client.Client;
 import org.opensearch.index.IndexNotFoundException;
@@ -78,7 +70,6 @@ import org.opensearch.index.reindex.ScrollableHitSource;
 
 import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.serialize.RandomCutForestSerDe;
-import com.google.common.util.concurrent.RateLimiter;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
@@ -116,15 +107,12 @@ public class CheckpointDao {
     private Gson gson;
     private RandomCutForestSerDe rcfSerde;
 
-    private ConcurrentLinkedQueue<DocWriteRequest<?>> requests;
-    private final ReentrantLock lock;
     private final Class<? extends ThresholdingModel> thresholdingModelClass;
-    private final Duration checkpointInterval;
-    private final Clock clock;
+
     private final AnomalyDetectionIndices indexUtil;
-    private final RateLimiter bulkRateLimiter;
-    private final int maxBulkRequestSize;
     private final JsonParser parser = new JsonParser();
+    // we won't read/write a checkpoint larger than a threshold
+    private final int maxCheckpointBytes;
 
     /**
      * Constructor with dependencies and configuration.
@@ -135,11 +123,8 @@ public class CheckpointDao {
      * @param gson accessor to Gson functionality
      * @param rcfSerde accessor to rcf serialization/deserialization
      * @param thresholdingModelClass thresholding model's class
-     * @param clock a UTC clock
-     * @param checkpointInterval how often we should save a checkpoint
      * @param indexUtil Index utility methods
-     * @param maxBulkRequestSize max number of index request a bulk can contain
-     * @param bulkPerSecond bulk requests per second
+     * @param maxCheckpointBytes max checkpoint size in bytes
      */
     public CheckpointDao(
         Client client,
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/ad/ratelimit/CheckpointWriteWorker.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    @Override
+    protected ActionListener<BulkResponse> getResponseListener(List<CheckpointWriteRequest> toProcess, BulkRequest batchRequest) {
+        return ActionListener.wrap(response -> {
+            for (BulkItemResponse r : response.getItems()) {
+                if (r.getFailureMessage() != null) {
+                    // maybe indicating a bug
+                    // don't retry failed requests since checkpoints are too large (250KB+)
+                    // Later maintenance window or cold start will retry saving
```
jmazanec15, line 152: Why not throw exception here?

kaituo: If throwing exceptions, I don't know how to handle it.  If we throw it to Opensearch, we risk bring the cluster down.  Log an exception so that we may be able to debug.

jmazanec15: Ah I see thanks

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -250,122 +224,37 @@ private void saveModelCheckpointAsync(Map<String, Object> source, String modelId
...
+     * @param modelState an entity model state
+     * @return serialized JSON map or empty map if the state is too bloated
+     * @throws IOException  when serialization fails
      */
-    public void write(ModelState<EntityModel> modelState, String modelId, boolean coldStart) {
-        Instant instant = modelState.getLastCheckpointTime();
-        // Instant.MIN is the default value. We don't save until we are sure.
-        if ((instant == Instant.MIN || instant.plus(checkpointInterval).isAfter(clock.instant())) && !coldStart) {
-            return;
+    public Map<String, Object> toIndexSource(ModelState<EntityModel> modelState) throws IOException {
```
jmazanec15, line 232: Why does this need to be public?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -250,122 +224,37 @@ private void saveModelCheckpointAsync(Map<String, Object> source, String modelId
...
+        if (serializedModel == null || serializedModel.length() > maxCheckpointBytes) {
+            logger
+                .warn(
+                    new ParameterizedMessage(
+                        "[{}]'s model empty or too large: [{}] bytes",
+                        modelState.getModelId(),
+                        serializedModel == null ? 0 : serializedModel.length()
+                    )
+                );
+            return source;
```
jmazanec15, line 245: Why return an empty map instead of throwing an exception?

src/main/java/org/opensearch/ad/ratelimit/CheckpointWriteWorker.java
```diff
@@ -0,0 +1,299 @@
...
+        this.indexUtil = indexUtil;
+        this.checkpoint = checkpoint;
+        this.indexName = indexName;
+        this.checkpointInterval = checkpointInterval;
+    }
+
+    @Override
+    protected void executeBatchRequest(BulkRequest request, ActionListener<BulkResponse> listener) {
+        if (indexUtil.doesCheckpointIndexExist()) {
+            clientUtil.<BulkRequest, BulkResponse>execute(BulkAction.INSTANCE, request, listener);
```
jmazanec15, line 116: Why doesnt this use the checkpointDao for putting checkpoints? Shouldnt all calls to put and get checkpoints go through the checkpointDao?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: CheckpointWriterWorker needs to call it :)

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: that can be an option.  It is two styles: https://docs.python.org/3/glossary.html#term-lbyl vs https://docs.python.org/3/glossary.html#term-eafp

For me, I am trying to avoid exceptions since it is more expensive than a conditional jump.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
jmazanec15: Right, but I am a little confused about the pattern here. Shouldn't CheckpointDao perform all checkpoint reads and writes? Why would the index source be needed outside of CheckpointDao.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
jmazanec15: Ah I see that makes sense.

Why then return an empty map instead of null?

src/main/java/org/opensearch/ad/ratelimit/CheckpointWriteWorker.java
kaituo: This line is a bulk call.  It does not involve any specifics of checkpoint.  CheckpointDao encapsulates inner working mechanisms of checkpoint index. Also, creating a one-line function inside CheckpointDao is wasteful.  So it seems fine to put this line here.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: Index source (the Map<String, Object> returned by toIndexSource) is part of the IndexRequest.  I need it to create an IndexRequest.  Are you suggesting I put creating index request inside CheckpointDao and CheckpointWriteWorker only get an IndexRequest from CheckpointDao?  CheckpointDao is a utility class that provides functions to deal with specifics of checkpoints.  CheckpointWriteWorkers invokes these functions to maintain a queue.  Are you suggesting I combine CheckpointWriterWorker and CheckpointDao?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: Best practices is not to return null.  Return an empty collection or Optional is recommended by the book Effective Java.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
jmazanec15: CheckpointWriterWorker only uses this function, `toIndexSource`, from CheckpointDao. It does not use any other function like [putModelCheckpoint](https://github.com/opensearch-project/anomaly-detection/pull/69/files#diff-ec6661209db81838626b3bc8ae378b8b20a7ab9e44ae6fce2d5c342ea498184aR181). 

So it seems to me that the DAO pattern is broken - checkpoint writes are happening outside of CheckpointDao. However, if you are okay with this, I am fine with it as well.

I understand that in `executeBatchRequest` the bulk request is a generic bulk request, so maybe it does not make sense to include in CheckpointDao, unless you can confirm it is a BulkRequest for checkpoint writes.

src/main/java/org/opensearch/ad/ratelimit/CheckpointWriteWorker.java
jmazanec15: Same discussion happening in above comment. Resolving

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: CheckpointWriterWorker can be viewed as a Dao for HCAD to write to the checkpoint index, similar to CheckpointReadWorker. I can put these two classes to CheckpointDao, but CheckpointDao would be so complex.  I don't know how much benefit we would gain.  It violates https://en.wikipedia.org/wiki/KISS_principle

Is there any place saying we should only have one Dao class for a single database responsibility?

jmazanec15: No, what I was saying is CheckpointWriteWorker already has a CheckpointDao as a member but it does not use it to interact with checkpoint persistence layer. It uses it as more of a utility. I was wondering if CheckpointWriteWorker should use the checkpointDao it already has to make calls to the checkpoint index.

From my understanding, the problem with using checkpointDao for all checkpoint operations inside CheckpointWriteWorker is that it does not really fit well in BatchWorker. `executeBatchRequest` takes and executes a bulk request that writes the checkpoint. If you were to delegate writing that bulkRequest to CheckpointDao, you would have to validate in CheckpointDao that the requests inside bulkRequest are actually CheckpointWriteRequests, or else checkpointDao could be used improperly. This validation is redundant.

On [Wikipedia](https://en.wikipedia.org/wiki/Data_access_object), they say the main benefit is the isolation between business logic and persistence layer. I think the benefit of having one class for Dao for a single database responsibility is that when changes need to be made with regards to persistence, only one class needs to be changed, supporting the single responsibility principle.

All of that being said, I am okay with this implementation.

kaituo: Got it.

I moved the read/write inside Worker classes to CheckpointDao.  Please check.

jmazanec15 (approved):

jngz-es (approved):

PR #70, opened 2021-05-25
Make dedicated cache’s size adjustable

kaituo:

### Description
Each detector has its dedicated cache that stores ten entities' states per node. A detector's hottest entities load their states into the dedicated cache. Other detectors cannot use space reserved by a detector's dedicated cache. Previously,  the size of the dedicated cache is not dynamically adjustable by users. This PR creates a dynamic setting AnomalyDetectorSettings.DEDICATED_CACHE_SIZE to make dedicated cache's size flexible. When that setting is changed, if the size decreases, we will release memory if required (e.g., when a user also decreased AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE, the max memory percentage that AD can use); if the size increases, we may reject the setting change if we cannot fulfill that request (e.g., when it will uses more memory than allowed for AD).

This PR also changes the behavior of the PriorityCache.get method. Previously, it will auto-load models if space is available or the external entity’s priority is high enough. Since we have moved the logic of loading models to rate-limited queues, we carry the loading models logic to a separate method and let the callers (i.e., queues) decide how to deal with models. PriorityCache.get will mostly return the existing state and update priority.

Testing done:
* manual testing by increasing/decreasing the setting value
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
@@ -47,8 +47,8 @@ import org.opensearch.ad.MemoryTracker.Origin;
 import org.opensearch.ad.ml.EntityModel;
 import org.opensearch.ad.ml.ModelState;
 import org.opensearch.ad.model.InitProgressProfile;
-import org.opensearch.ad.ratelimit.CheckpointWriteQueue;
-import org.opensearch.ad.ratelimit.SegmentPriority;
+import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
+import org.opensearch.ad.ratelimit.RequestPriority;
 
 /**
  * We use a layered cache to manage active entities’ states.  We have a two-level
@@ -87,7 +87,7 @@ public class CacheBuffer implements ExpiringState {
     private long reservedBytes;
     private final PriorityTracker priorityTracker;
     private final Clock clock;
-    private final CheckpointWriteQueue checkpointWriteQueue;
+    private final CheckpointWriteWorker checkpointWriteQueue;
     private final Random random;
 
     public CacheBuffer(
@@ -98,7 +98,7 @@ public class CacheBuffer implements ExpiringState {
         Clock clock,
         Duration modelTtl,
         String detectorId,
-        CheckpointWriteQueue checkpointWriteQueue,
+        CheckpointWriteWorker checkpointWriteQueue,
         Random random
     ) {
         this.memoryConsumptionPerEntity = memoryConsumptionPerEntity;
@@ -241,23 +241,18 @@ public class CacheBuffer implements ExpiringState {
         ModelState<EntityModel> valueRemoved = items.remove(keyToRemove);
 
         if (valueRemoved != null) {
-            if (reserved) {
-                // release in reserved memory
-                memoryTracker.releaseMemory(memoryConsumptionPerEntity, true, Origin.HC_DETECTOR);
-            } else {
+            if (!reserved) {
                 // release in shared memory
                 memoryTracker.releaseMemory(memoryConsumptionPerEntity, false, Origin.HC_DETECTOR);
             }
 
             EntityModel modelRemoved = valueRemoved.getModel();
             if (modelRemoved != null) {
-                if (modelRemoved.getRcf() == null || modelRemoved.getThreshold() == null) {
-                    // only have samples. If we don't save, we throw the new samples and might
-                    // never be able to initialize the model
-                    checkpointWriteQueue.write(valueRemoved, true, SegmentPriority.MEDIUM);
-                } else {
-                    checkpointWriteQueue.write(valueRemoved, false, SegmentPriority.MEDIUM);
-                }
+                // null model has only samples. For null model we save a checkpoint
+                // regardless of last checkpoint time. whether If we don't save,
+                // we throw the new samples and might never be able to initialize the model
+                boolean isNullModel = modelRemoved.getRcf() == null || modelRemoved.getThreshold() == null;
+                checkpointWriteQueue.write(valueRemoved, isNullModel, RequestPriority.MEDIUM);
 
                 modelRemoved.clear();
             }
@@ -353,7 +348,7 @@ public class CacheBuffer implements ExpiringState {
             }
         });
 
-        checkpointWriteQueue.writeAll(modelsToSave, detectorId, false, SegmentPriority.MEDIUM);
+        checkpointWriteQueue.writeAll(modelsToSave, detectorId, false, RequestPriority.MEDIUM);
         return removedStates;
     }
 
--- a/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
@@ -29,9 +29,11 @@ package org.opensearch.ad.caching;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -240,16 +241,26 @@ public boolean canRemove() {
...
-            // if we releasing a shared cache item, release memory as well.
-            if (!reserved) {
-                memoryTracker.releaseMemory(memoryConsumptionPerEntity, false, Origin.MULTI_ENTITY_DETECTOR);
+            if (reserved) {
+                // release in reserved memory
+                memoryTracker.releaseMemory(memoryConsumptionPerEntity, true, Origin.HC_DETECTOR);
+            } else {
+                // release in shared memory
+                memoryTracker.releaseMemory(memoryConsumptionPerEntity, false, Origin.HC_DETECTOR);
             }
```
LiuJoyceC, line 250: Could be simplified to:
`memoryTracker.releaseMemory(memoryConsumptionPerEntity, reserved, Origin.HC_DETECTOR);`

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -240,16 +241,26 @@ public boolean canRemove() {
...
-            modelRemoved.clear();
+            EntityModel modelRemoved = valueRemoved.getModel();
+            if (modelRemoved != null) {
+                if (modelRemoved.getRcf() == null || modelRemoved.getThreshold() == null) {
+                    // only have samples. If we don't save, we throw the new samples and might
+                    // never be able to initialize the model
+                    checkpointWriteQueue.write(valueRemoved, true, SegmentPriority.MEDIUM);
+                } else {
+                    checkpointWriteQueue.write(valueRemoved, false, SegmentPriority.MEDIUM);
+                }
```
LiuJoyceC, line 260: [minor] Similar to above, could simply to:
```
// not sure what to name this variable since I can't find the class CheckpointWriteQueue and read the function signature
boolean isNullModel = modelRemoved.getRcf() == null || modelRemoved.getThreshold() == null;
checkpointWriteQueue.write(valueRemoved, isNullModel, SegmentPriority.MEDIUM);
```

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -40,13 +42,13 @@
...
 import org.opensearch.ad.ExpiringState;
-import org.opensearch.ad.MaintenanceState;
 import org.opensearch.ad.MemoryTracker;
 import org.opensearch.ad.MemoryTracker.Origin;
-import org.opensearch.ad.ml.CheckpointDao;
 import org.opensearch.ad.ml.EntityModel;
 import org.opensearch.ad.ml.ModelState;
 import org.opensearch.ad.model.InitProgressProfile;
+import org.opensearch.ad.ratelimit.CheckpointWriteQueue;
+import org.opensearch.ad.ratelimit.SegmentPriority;
```
LiuJoyceC, line 51: Are these classes defined somewhere? I can't find them in the current repo nor in this diff.

kaituo: Yes, they are defined somewhere and was included in separate PRs.  Please use this branch to check: https://github.com/kaituo/anomaly-detection-1/tree/multiCat

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
kaituo: yeah, changed using your version

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
kaituo: I realized I shouldn't release reserved memory.  Only keep the branch !reserved now.  Please check the new version.

LiuJoyceC (approved):

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -324,21 +335,21 @@ public void maintenance() {
...
-                    // we can have ConcurrentModificationException when serializing
-                    // and updating rcf model at the same time. To prevent this,
-                    // we need to have a deep copy of models or have a lock. Both
-                    // options are costly.
-                    // As we are gonna retry serializing either when the entity is
-                    // evicted out of cache or during the next maintenance period,
-                    // don't do anything when the exception happens.
-                    checkpointDao.write(modelState, entityModelId);
+                    removedStates.add(remove(entityModelId));
+                } else if (random.nextInt(6) == 0) {
```
ylwu-amzn, line 339: Why choose 6 ?  Add some comments?

kaituo: The comments/explanation are in the next few lines.

ylwu-amzn: oh, just see it. Sorry I jumped to a call when I reach this line. Continue reading this PR now.

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -324,21 +335,21 @@ public void maintenance() {
...
+                    // checkpoint is relatively big compared to other queued requests
+                    // save checkpoints with 1/6 probability as we expect to save
+                    // all every 6 hours statistically
```
ylwu-amzn, line 342: Any reason to use random to decide whether we should save model checkpoints rather than some deterministic way? 
For this case: If some model checkpoints missing in index and we need to analyze why it's not saved, will be hard to tell if it's not saved as the random was not 6 or failed to save.
And as you said "checkpoint is relatively big", so I guess there will be some spike on metrics like disk write/CPU/memory when we save checkpoint? If we need to analyze the metrics, the random spikes may be very hard to explain.

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -449,11 +460,19 @@ public String getDetectorId() {
...
-    public List<ModelState<?>> getAllModels() {
+    public List<ModelState<EntityModel>> getAllModels() {
         return items.values().stream().collect(Collectors.toList());
     }
 
     public PriorityTracker getPriorityTracker() {
         return priorityTracker;
     }
+
+    public void setMinimumCapacity(int minimumCapacity) {
```
ylwu-amzn, line 471: Seems the `minimumCapacity` means minimum entity count? Add some comments or change the variable name ?

src/main/java/org/opensearch/ad/caching/PriorityCache.java
```diff
@@ -26,105 +26,112 @@
...
-import com.google.common.util.concurrent.RateLimiter;
 
 public class PriorityCache implements EntityCache {
     private final Logger LOG = LogManager.getLogger(PriorityCache.class);
 
     // detector id -> CacheBuffer, weight based
     private final Map<String, CacheBuffer> activeEnities;
     private final CheckpointDao checkpointDao;
-    private final int dedicatedCacheSize;
+    private int dedicatedCacheSize;
```
ylwu-amzn, line 85: Add `volatile` ?

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
kaituo: We will save a checkpoint when
1) removing the model from cache.
2) cold start
3) no complete model only a few samples.  If we don't save new samples, we will never be able to have enough samples for a trained mode.
4) periodically save in case of exceptions.

This branch is doing 4). Previously, I will do it every hour for all in-cache models.  Consider we are moving to 1M entities, this will bring the cluster in a heavy payload every hour.  That's why I am doing it randomly (expected 6 hours for each checkpoint statistically).  

I am doing it random since maintaining a state of which one has been saved and which one hasn't are not cheap.  Also, the models in the cache can be dynamically changing.  Will have to maintain the state in the removing logic.  Random is a lazy way to deal with this as it is stateless and statistically sound.  

We will have a checkpoint, just that if a checkpoint does not fall into the 6-hour bucket in a particular scenario, the model is stale (i.e., we don't recover from the freshest model in disaster.).

Yes, this change is mostly due to performance and easy maintenance.  

Yes, the solution has flaws.  It is a tradeoff.  Any other suggestions?

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
kaituo: It means the reserved cache size.  So no matter how many entities there are, we will keep the size for minimum capacity entities.  Added the comments.

src/main/java/org/opensearch/ad/caching/PriorityCache.java
kaituo: sure.  Just curious, has a bug related to this happened?

ylwu-amzn: Before this PR we just use `final` value. We need to add `volatile` as we add setting update consumer in this PR.

kaituo: not particular to this PR.  Just wondering since you commented about this in other PRs too.

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
ylwu-amzn: Thanks, look good as tradeoff

ylwu-amzn (approved):

PR #71, opened 2021-05-25
Illustrate how we use queues

kaituo:

### Description
We have created multiple queues for rate-limiting expensive requests.  This PR illustrates how we actually use these queues.
1. We store as many frequently used entity models in a cache as allowed by the memory limit (10% heap). If an entity feature is a hit, we use the in-memory model to detect anomalies and record results using the result write queue.
2. If an entity feature is a miss, we check if there is free memory or any other entity's model can be evacuated. An in-memory entity's frequency may be lower compared to the cache miss entity. If that's the case, we replace the lower frequency entity's model with the higher frequency entity's model. To load the higher frequency entity's model, we first check if a model exists on disk by sending a checkpoint read queue request. If there is a checkpoint, we load it to memory, perform detection, and save the result using the result write queue. Otherwise, we enqueue a cold start request to the cold start queue for model training. If training is successful, we save the learned model via the checkpoint write queue.
3. We also have the cold entity queue configured for cold entities, and the model training and inference are connected by serial juxtaposition to limit resource usage.

Testing done:
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
@@ -61,12 +61,12 @@ import org.opensearch.ad.ml.ThresholdingResult;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.AnomalyResult;
 import org.opensearch.ad.model.Entity;
-import org.opensearch.ad.ratelimit.CheckpointReadQueue;
-import org.opensearch.ad.ratelimit.ColdEntityQueue;
+import org.opensearch.ad.ratelimit.CheckpointReadWorker;
+import org.opensearch.ad.ratelimit.ColdEntityWorker;
 import org.opensearch.ad.ratelimit.EntityFeatureRequest;
-import org.opensearch.ad.ratelimit.ResultWriteQueue;
+import org.opensearch.ad.ratelimit.RequestPriority;
 import org.opensearch.ad.ratelimit.ResultWriteRequest;
-import org.opensearch.ad.ratelimit.SegmentPriority;
+import org.opensearch.ad.ratelimit.ResultWriteWorker;
 import org.opensearch.ad.util.ExceptionUtil;
 import org.opensearch.ad.util.ParseUtils;
 import org.opensearch.common.inject.Inject;
@@ -74,6 +74,25 @@ import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportService;
 
+/**
+ * Entry-point for HCAD workflow.  We have created multiple queues for coordinating
+ * the workflow. The overrall workflow is:
+ * 1. We store as many frequently used entity models in a cache as allowed by the
+ *  memory limit (10% heap). If an entity feature is a hit, we use the in-memory model
+ *  to detect anomalies and record results using the result write queue.
+ * 2. If an entity feature is a miss, we check if there is free memory or any other
+ *  entity's model can be evacuated. An in-memory entity's frequency may be lower
+ *  compared to the cache miss entity. If that's the case, we replace the lower
+ *  frequency entity's model with the higher frequency entity's model. To load the
+ *  higher frequency entity's model, we first check if a model exists on disk by
+ *  sending a checkpoint read queue request. If there is a checkpoint, we load it
+ *  to memory, perform detection, and save the result using the result write queue.
+ *  Otherwise, we enqueue a cold start request to the cold start queue for model
+ *  training. If training is successful, we save the learned model via the checkpoint
+ *  write queue.
+ * 3. We also have the cold entity queue configured for cold entities, and the model
+ * training and inference are connected by serial juxtaposition to limit resource usage.
+ */
 public class EntityResultTransportAction extends HandledTransportAction<EntityResultRequest, AcknowledgedResponse> {
 
     private static final Logger LOG = LogManager.getLogger(EntityResultTransportAction.class);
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/ad/util/ExceptionUtil.java
```diff
@@ -116,4 +132,83 @@ public static String getErrorMessage(Exception e) {
...
+        // LimitExceededException may indicate circuit breaker exception
+        return cause instanceof RejectedExecutionException
+            || TransportActions.isShardNotAvailableException(cause)
+            || cause instanceof LimitExceededException;
+    }
+
+    public static boolean isRetryAble(Exception e) {
+        Throwable cause = ExceptionsHelper.unwrapCause(e);
+        RestStatus status = ExceptionsHelper.status(cause);
+        return RETRYABLE_STATUS.contains(status);
```
jmazanec15, line 152: Can this call 155, just in case 155 changes, you dont have to change 2 spots?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -162,65 +164,123 @@ protected void doExecute(Task task, EntityResultRequest request, ActionListener<
...
+            }
 
             Instant executionStartTime = Instant.now();
-            for (Entry<String, double[]> entity : request.getEntities().entrySet()) {
-                String entityName = entity.getKey();
-                // For ES, the limit of the document ID is 512 bytes.
-                // skip an entity if the entity's name is more than 256 characters
-                // since we are using it as part of document id.
-                if (entityName.length() > AnomalyDetectorSettings.MAX_ENTITY_LENGTH) {
+            // Map<String, EntityFeatureRequest> cacheMissEntities = new HashMap<>();
```
jmazanec15, line 174: Can this be removed?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -162,65 +164,123 @@ protected void doExecute(Task task, EntityResultRequest request, ActionListener<
...
 
-                double[] datapoint = entity.getValue();
-                String modelId = manager.getEntityModelId(detectorId, entityName);
-                ModelState<EntityModel> entityModel = cache.get().get(modelId, detector, datapoint, entityName);
+                String modelId = modelIdOptional.get();
+                double[] datapoint = entityEntry.getValue();
+                ModelState<EntityModel> entityModel = cache.get().get(modelId, detector);
                 if (entityModel == null) {
                     // cache miss
+                    cacheMissEntities.put(entityEntry.getKey(), entityEntry.getValue());
```
jmazanec15, line 189: Use the variable in 177 and 185.

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -162,65 +164,123 @@ protected void doExecute(Task task, EntityResultRequest request, ActionListener<
...
+                if (hotEntityValue == null) {
+                    LOG.error(new ParameterizedMessage("feature value should not be null: [{}]", hotEntity));
+                    continue;
+                }
+                hotEntityRequests
+                    .add(
+                        new EntityFeatureRequest(
+                            System.currentTimeMillis() + detector.getDetectorIntervalInMilliseconds(),
+                            detectorId,
+                            // will change once we know the anomaly grade. Set it to low since most of the entities should not count as
```
jmazanec15, line 244: Low or medium?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -162,65 +164,123 @@ protected void doExecute(Task task, EntityResultRequest request, ActionListener<
...
+                if (coldEntityValue == null) {
+                    LOG.error(new ParameterizedMessage("feature value should not be null: [{}]", coldEntity));
+                    continue;
+                }
+                coldEntityRequests
+                    .add(
+                        new EntityFeatureRequest(
+                            System.currentTimeMillis() + detector.getDetectorIntervalInMilliseconds(),
+                            detectorId,
+                            // will change once we know the anomaly grade. Set it to low since most of the entities should not count as
```
jmazanec15, line 265: These are cold entities. Does this comment still apply?

src/main/java/org/opensearch/ad/util/ExceptionUtil.java
kaituo: changed

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: yeah, removed

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: yes, fixed

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: yeah, fixed the comments.

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: it does not.  Thanks for the catch. Fixed.

jmazanec15 (approved):

jngz-es (approved):

PR #74, opened 2021-05-26
Only allocate models when circuit breaker is closed

kaituo:
Note: since there are a lot of dependencies, I only list the main class and test code to save reviewers' time. The build will fail due to missing dependencies. I will use that PR just for review. will not merge it. Will have a big one in the end and merge once after all review PRs get approved.  Now the code is missing unit tests.  Will add unit tests, run performance tests, and fix bugs before the official release.

### Description
Previously we may allocate memory when the circuit breaker is open but AD hasn’t used more than 10% of memory yet. This PR disallows any new model creation when the circuit breaker is open, which provides extra safety to prevent OOM.

This PR also differentiates isHostingAllowed and canAllocateReserved. Previously, both of them throw exceptions when allocation is not allowed. But most of the time it is fine to return false and let callers decide what to do if we cannot allocate. isHostingAllowed needs an exception thrown due to the special way the single-stream detector code is written (can be changed). This PR keeps the original exception thrown behavior of isHostingAllowed, but changes canAllocateReserved to return false instead.

Testing done:
1. Tested an open circuit breaker prevented model creation.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/MemoryTracker.java
+++ b/src/main/java/org/opensearch/ad/MemoryTracker.java
@@ -34,6 +34,7 @@ import java.util.Map;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.ad.breaker.ADCircuitBreakerService;
 import org.opensearch.ad.common.exception.LimitExceededException;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.cluster.service.ClusterService;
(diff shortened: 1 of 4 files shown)


ylwu-amzn (approved):
LGTM

src/main/java/org/opensearch/ad/MemoryTracker.java
```diff
@@ -50,7 +51,7 @@
 
     public enum Origin {
         SINGLE_ENTITY_DETECTOR,
-        MULTI_ENTITY_DETECTOR,
+        HC_DETECTOR,
```
ylwu-amzn, line 54: Seems this caused build failure. Make sure all places changed.

kaituo: sure, will make sure that in the last PR with all of the code.

ohltyler (approved):

PR #75, opened 2021-05-26
Add model id in anomaly result.

kaituo:
Note: since there are a lot of dependencies, I only list the main class and test code to save reviewers' time. The build will fail due to missing dependencies. I will use that PR just for review. will not merge it. Will have a big one in the end and merge once after all review PRs get approved.  Now the code is missing unit tests.  Will add unit tests, run performance tests, and fix bugs before the official release.

### Description
The front end needs to query for entities ordered by the descending order of anomaly grades and the number of anomalies. After supporting multi-category fields, it is hard to write such queries since the entity information is stored in a nested object array. Also, the front end has all code/queries/ helper functions in place to rely on a single key per entity combo. This PR adds model id to anomaly result to help the transition to multi-categorical field less painful.

Testing done:
1. Tested HCAD results have model id now.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
+++ b/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
@@ -597,7 +597,8 @@ public class ADBatchTaskRunner {
                     error,
                     null,
                     adTask.getDetector().getUser(),
-                    anomalyDetectionIndices.getSchemaVersion(ADIndex.RESULT)
+                    anomalyDetectionIndices.getSchemaVersion(ADIndex.RESULT),
+                    null
                 );
                 anomalyResults.add(anomalyResult);
             } else {
@@ -638,7 +639,8 @@ public class ADBatchTaskRunner {
                     null,
                     null,
                     adTask.getDetector().getUser(),
-                    anomalyDetectionIndices.getSchemaVersion(ADIndex.RESULT)
+                    anomalyDetectionIndices.getSchemaVersion(ADIndex.RESULT),
+                    null
                 );
                 anomalyResults.add(anomalyResult);
             }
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -597,7 +597,8 @@ private void detectAnomaly(
                     error,
                     null,
                     adTask.getDetector().getUser(),
-                    anomalyDetectionIndices.getSchemaVersion(ADIndex.RESULT)
+                    anomalyDetectionIndices.getSchemaVersion(ADIndex.RESULT),
+                    null
```
ohltyler, line 601: Why set to `null` here, can the model id not be pulled from the `rcf` variable? Where is the `model_id` actually added into the results exactly?

ohltyler: ^ I may have answered my own question - is this because this is related to historical detector results and `model_id` is not needed in that case? If so, how do we plan on handling the future case of historical HC detectors?

kaituo: yes, historical detector has no model id, bu task id.  Adding a model id may not be a good design as
* historical detector does not use it.  It uses task id. Front-end needs two sets of logic for real-time and historical detectors.
* it increases the size of result index doc.

We may be better off using the new result search API.  The only issue is the timeline.  We may not be able to finish appsec before 1.1 release.  Any suggestion?

ohltyler: I see. Makes sense to not include it, and we can use `task_id` instead of `model_id` for aggregating HC historical results, if the new result search API is still not ready. Thanks for the clarification

ohltyler (approved):

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: For historical results, we just store detector level task id in AD result index. I think we need to store model id in historical result as well to search entity's results.

ylwu-amzn (approved):
LGTM

PR #76, opened 2021-05-26
Add pagination to HCAD feature query

kaituo:

### Description
HCAD issues a query to fetch feature data for each entity regularly. HCAD v1 limit the number of entity values returned per query to 1000, both for reasons of efficiency (fetching the overall entity values is expensive and may not finish before the next scheduled query run starts) and for reasons of ease-of-load-shedding (the top 1000 limit, in turn, curtails the memory used to host models, the disk access needed to read/write model checkpoints and anomaly results, the CPU used for entity metadata maintenance and model training/inference, and the garbage collection for deleted models and metadata). Users can increase the maximum entity setting at the cost of a linear growth resource usage, which opens the door to cluster instability. However, such a heuristic “best-effort” limit is detrimental to the scalability of entity monitoring. Even if a user scales out, the number of monitored entities does not increase.

This PR changes to use pagination to fetch entities.  If there are more than 1000 entities, we will fetch them in the next page. We implement pagination with composite query. Results are decomposed into a set of 1000-entity pages. Each page encapsulates aggregated values for each entity and is sent to model nodes according to the hash ring mapping from entity model Id to a data node.

Testing done:
1. Manually tested pagination works.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
+++ b/src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
@@ -25,8 +25,6 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
-import org.opensearch.ad.common.exception.EndRunException;
-import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.model.Feature;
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -0,0 +1,346 @@
...
+
+        } catch (Exception e) {
+            listener
+                .onFailure(new EndRunException(anomalyDetector.getDetectorId(), CommonErrorMessages.INVALID_SEARCH_QUERY_MSG, e, false));
+        }
+    }
+
+    public class Page {
+        // a map from categorical field name to values (type: java.lang.Comparable)
+        Map<String, Object> afterKey;
```
penghuo, line 167: Does the listener need to know the afterKey and source? IMO, the listener should only consume the results.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -0,0 +1,346 @@
...
+            }
+
+            // In order to optimize the early termination it is advised to set track_total_hits in the request to false.
+            SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
+                .query(internalFilterQuery)
+                .size(0)
+                .aggregation(composite)
+                .trackTotalHits(false);
+
+            Page page = new Page(null, searchSourceBuilder, null);
```
penghuo, line 156: Does the iterable fit to this use case? e.g. class IterablePage extends Iterable<Result>

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -0,0 +1,346 @@
...
+    private final AnomalyDetector anomalyDetector;
+    private final NamedXContentRegistry xContent;
+    private final Client client;
+    private int totalResults;
+    private int maxEntities;
+    private final int pageSize;
+    private long expirationEpochMs;
+    private Clock clock;
+
+    public CompositeRetriever(
```
penghuo, line 73: Missing UT?

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: no, the listener does not need to know the afterkey and source.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: You meant https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html ?  I need parameters in the next method.  Does not seem fit.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: Will add UT in the following PRs :)

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -0,0 +1,346 @@
...
+            CompositeAggregationBuilder composite = AggregationBuilders
+                .composite(
+                    AGG_NAME_COMP,
+                    anomalyDetector
+                        .getCategoryField()
+                        .stream()
+                        .map(f -> new TermsValuesSourceBuilder(f).field(f))
+                        .collect(Collectors.toList())
+                )
+                .size(pageSize);
```
dai-chen, line 142: Is the page size hardcoding somewhere or configurable?

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -0,0 +1,346 @@
...
+            } else {
+                throw new IllegalArgumentException(
+                    String.format(Locale.ROOT, "Invalid client request; expected a composite builder but instead got {}", aggBuilder)
+                );
+            }
+        }
+
+        private boolean shouldRetryDueToEmptyPage(SearchResponse response) {
+            CompositeAggregation composite = getComposite(response);
+            // if there are no buckets but a next page, go fetch it instead of sending an empty response to the client
```
dai-chen, line 292: Not very familiar and wondering in which case this would happen?

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: I tried to separate page and pageIterator.  Please check the new commit.  Also, cannot extend JDK interface since the method signature next() does not match my need.  I need to return result using async listener callback.  Iterator.next() in JDK is a synchronous method.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: not sure, just a possibility :)

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: It is configurable via a dynamic setting.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -0,0 +1,364 @@
...
+            return composite != null && composite.getBuckets().isEmpty() && composite.afterKey() != null && !composite.afterKey().isEmpty();
+        }
+
+        CompositeAggregation getComposite(SearchResponse response) {
+            if (response == null || response.getAggregations() == null) {
+                return null;
+            }
+            Aggregation agg = response.getAggregations().get(AGG_NAME_COMP);
+            if (agg == null) {
```
penghuo, line 301: Consider using Optional to enforece the calller to handle the null case?

kaituo: yeah, changed to Optional.

penghuo (approved):
Thanks for the change!

PR #77, opened 2021-05-26
Combine CommonMessageAttributes and CommonName

kaituo:

### Description
Both CommonMessageAttributes and CommonName define some key or name used as in Json keys or log messages.  This PR combines these two class by moving the content from CommonMessageAttributes to CommonName.  This is so we can maintain such keys or attributes in a central place.

Testing done:
1. Build passes.
 
### Check List


src/test/java/org/opensearch/ad/AbstractADTest.java
```diff
@@ -212,7 +216,7 @@ protected static void setUpThreadPool(String name) {
                 AnomalyDetectorPlugin.AD_THREAD_POOL_NAME,
                 1,
                 1000,
-                "opendistro.ad." + AnomalyDetectorPlugin.AD_THREAD_POOL_NAME
+                "opensearch.ad." + AnomalyDetectorPlugin.AD_THREAD_POOL_NAME
```
ohltyler, line 219: Nice find. Can you grep in your final codebase that has all of the changes, that there are no more references of `opendistro`?

kaituo: There are quite a few of opendistro like

src/main/java/org/opensearch/ad/constant/CommonName.java
34:    public static final String CHECKPOINT_INDEX_NAME = ".opendistro-anomaly-checkpoints";
36:    public static final String DETECTION_STATE_INDEX = ".opendistro-anomaly-detection-state";
40:    public static final String ANOMALY_RESULT_INDEX_ALIAS = ".opendistro-anomaly-results";

According to infra team, our setting will have backward compatibility to opendistro, while our indices will use keep using opendistro.

ohltyler: Understood, agreed the BWC vars and the indices will remain `opendistro`, was more referring to other places that can be changed like the thread pool name above.

src/test/java/org/opensearch/ad/AbstractADTest.java
```diff
@@ -223,17 +227,33 @@ protected static void tearDownThreadPool() {
...
 
-    public void setupTestNodes(Settings settings, TransportInterceptor transportInterceptor) {
+    /**
+     *
+     * @param transportInterceptor Interceptor to for transport requests. Used
+     *  to mock transport layer.
+     * @param nodeSettings node override of setting
+     * @param setting the supported setting set.
+     */
+    public void setupTestNodes(TransportInterceptor transportInterceptor, final Settings nodeSettings, Setting<?>... setting) {
```
ohltyler, line 237: Can you explain why you've added the additional `Setting` set here to pass to `FakeNode`?

src/test/java/org/opensearch/ad/AbstractADTest.java
kaituo: right, fixed a few in our docs and comments.  Please check the new commit.

src/test/java/org/opensearch/ad/AbstractADTest.java
kaituo: Please see an example: https://github.com/kaituo/anomaly-detection-1/blob/multiCat4/src/test/java/org/opensearch/ad/transport/AnomalyResultTests.java#L437-L442

The added setting set is to list supported setting set.  It is used as part of testing setting changes.

Haven't sent out tests in PRs as I am still adding unit tests.

ohltyler: Got it, makes sense.

src/test/java/org/opensearch/ad/AbstractADTest.java
ohltyler: Not sure what commit you are referring to?

kaituo: this commit: https://github.com/opensearch-project/anomaly-detection/pull/77/commits/4cf9c7a24337dd27b29bbfc77596091c6281778d

ohltyler (approved):
LGTM

spbjss (approved):
looks good to me.

PR #78, opened 2021-05-26
Modify Entity to support multi-category field

kaituo:

### Description
This PR modified Entity to include multiple categorical fields.  One key change is to not use categorical values directly in the model document Id. The problem is that HCAD v1 uses the categorical value as part of the model document Id, but ES’s document Id can be at most 512 bytes. Categorical values are usually less than 256 characters, but can grow to 32766 in theory. HCAD v1 skips an entity if the entity's name is more than 256 characters. We cannot do that in v2 as that can reject a lot of entities. To overcome the obstacle, we hash categorical values to a 128-bit string (like SHA-1 that git uses) and use the hash as part of the model document Id.

We have choices to make regarding when to use the hash as part of a model document Id: for all HC detectors or a HC detector with multiple categorical fields. The challenge lies in providing backward compatibility of looking for a model checkpoint in the case of a HC detector with one categorical field. If using hashes for all HC detectors, we need two get requests to ensure that a model checkpoint exists. One uses the document Id without a hash, while one uses the document Id with a hash. The dual get requests are ineffective. If limiting hashes to a HC detector with multiple categorical fields, there is no backward compatibility issue. However, the code will be branchy. One may wonder if backward compatibility can be ignored; indeed, the old checkpoints will be gone after a transition period during upgrading. During the transition period, HC detectors can experience unnecessary cold starts as if the detectors were just started. Checkpoint index size can double if every model has two model documents. The transition period can be as long as 3 days since our checkpoint retention period is 3 days. There is no perfect solution. We prefer limiting hashes to an HC detector with multiple categorical fields as its customer impact is none.

Testing done:
1. Manually tested the model id of multi-category entity works, and the single-category entity does not break.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
@@ -37,6 +37,7 @@ import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.OpenSearchSecurityException;
 import org.opensearch.action.ActionListener;
 import org.opensearch.ad.constant.CommonValue;
 import org.opensearch.ad.feature.FeatureManager;
@@ -50,6 +51,7 @@ import org.opensearch.ad.model.EntityAnomalyResult;
 import org.opensearch.ad.model.Feature;
 import org.opensearch.ad.model.FeatureData;
 import org.opensearch.ad.util.MultiResponsesDelegateActionListener;
+import org.opensearch.common.util.concurrent.ThreadContext;
 
 /**
  * Runner to trigger an anomaly detector.
@@ -73,11 +75,18 @@ public final class AnomalyDetectorRunner {
      * @param detector  anomaly detector instance
      * @param startTime detection period start time
      * @param endTime   detection period end time
+     * @param context   stored thread context
      * @param listener handle anomaly result
      * @throws IOException - if a user gives wrong query input when defining a detector
      */
-    public void executeDetector(AnomalyDetector detector, Instant startTime, Instant endTime, ActionListener<List<AnomalyResult>> listener)
-        throws IOException {
+    public void executeDetector(
+        AnomalyDetector detector,
+        Instant startTime,
+        Instant endTime,
+        ThreadContext.StoredContext context,
+        ActionListener<List<AnomalyResult>> listener
+    ) throws IOException {
+        context.restore();
         List<String> categoryField = detector.getCategoryField();
         if (categoryField != null && !categoryField.isEmpty()) {
             featureManager.getPreviewEntities(detector, startTime.toEpochMilli(), endTime.toEpochMilli(), ActionListener.wrap(entities -> {
@@ -135,6 +144,9 @@ public final class AnomalyDetectorRunner {
         // TODO return exception like IllegalArgumentException to explain data is not enough for preview
         // This also requires front-end change to handle error message correspondingly
         // We return empty list for now to avoid breaking front-end
+        if (e instanceof OpenSearchSecurityException) {
+            listener.onFailure(e);
+        }
         listener.onResponse(Collections.emptyList());
     }
 
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
@@ -96,6 +96,7 @@ public final class AnomalyDetectorRunner {
                     // This also requires front-end change to handle error message correspondingly
                     // We return empty list for now to avoid breaking front-end
                     listener.onResponse(Collections.emptyList());
+                    return;
                 }
                 ActionListener<EntityAnomalyResult> entityAnomalyResultListener = ActionListener
                     .wrap(
@@ -146,6 +147,7 @@ public final class AnomalyDetectorRunner {
         // We return empty list for now to avoid breaking front-end
         if (e instanceof OpenSearchSecurityException) {
             listener.onFailure(e);
+            return;
         }
         listener.onResponse(Collections.emptyList());
     }
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
@@ -29,7 +29,6 @@ package org.opensearch.ad;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
@@ -111,7 +110,7 @@ public final class AnomalyDetectorRunner {
                             ActionListener.wrap(features -> {
                                 List<ThresholdingResult> entityResults = modelManager.getPreviewResults(features.getProcessedFeatures());
                                 List<AnomalyResult> sampledEntityResults = sample(
-                                    parsePreviewResult(detector, features, entityResults, Arrays.asList(entity)),
+                                    parsePreviewResult(detector, features, entityResults, entity),
                                     maxPreviewResults
                                 );
                                 multiEntitiesResponseListener.onResponse(new EntityAnomalyResult(sampledEntityResults));
@@ -143,7 +142,7 @@ public final class AnomalyDetectorRunner {
         AnomalyDetector detector,
         Features features,
         List<ThresholdingResult> results,
-        List<Entity> entity
+        Entity entity
     ) {
         // unprocessedFeatures[][], each row is for one date range.
         // For example, unprocessedFeatures[0][2] is for the first time range, the third feature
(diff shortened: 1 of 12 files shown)


src/main/java/org/opensearch/ad/model/Entity.java
```diff
@@ -29,66 +29,146 @@
...
+                    case ATTRIBUTE_NAME_FIELD:
+                        parsedName = parser.text();
+                        break;
+                    case ATTRIBUTE_VALUE_FIELD:
+                        parsedValue = parser.text();
+                        if (parsedName != null && parsedValue != null) {
+                            entities.put(parsedName, parsedValue);
+                        }
+                        parsedValue = null;
+                        parsedName = null;
```
ohltyler, line 164: Nit - why have this logic embedded in the `ATTRIBUTE_VALUE_FIELD` case? Can it be moved after the switch statement for readability? I see this can be more efficient if it is always guaranteed that the attribute name will come before the value, just wonder if it adds confusion to add this here.

src/main/java/org/opensearch/ad/model/Entity.java
```diff
@@ -99,28 +179,205 @@ public boolean equals(Object o) {
...
+    public static Optional<String> getModelId(String detectorId, SortedMap<String, String> attributes) {
+        if (attributes.isEmpty()) {
+            return Optional.empty();
+        } else if (attributes.size() == 1) {
+            for (Map.Entry<String, String> categoryValuePair : attributes.entrySet()) {
+                // For OpenSearch, the limit of the document ID is 512 bytes.
+                // skip an entity if the entity's name is more than 256 characters
+                // since we are using it as part of document id.
+                String categoricalValue = categoryValuePair.getValue().toString();
+                if (categoricalValue.length() > AnomalyDetectorSettings.MAX_ENTITY_LENGTH) {
```
ohltyler, line 255: What is the consequence for returning empty if the category field value is too long? How was this handled before?

src/main/java/org/opensearch/ad/transport/GetAnomalyDetectorRequest.java
```diff
@@ -42,7 +43,7 @@
     private String typeStr;
     private String rawPath;
     private boolean all;
-    private String entityValue;
+    private Entity entityValue;
```
ohltyler, line 46: nit - should this var be renamed to `entity` instead of `entityValue`, since it now includes a list of attributes rather than a single value?

src/main/java/org/opensearch/ad/transport/GetAnomalyDetectorTransportAction.java
```diff
@@ -159,7 +160,7 @@ protected void getExecute(GetAnomalyDetectorRequest request, ActionListener<GetA
         String detectorID = request.getDetectorID();
         String typesStr = request.getTypeStr();
         String rawPath = request.getRawPath();
-        String entityValue = request.getEntityValue();
+        Entity entityValue = request.getEntityValue();
```
ohltyler, line 163: nit - same thing here? Maybe rename to just `entity`?

src/test/java/org/opensearch/ad/TestHelpers.java
```diff
@@ -636,7 +637,7 @@ public static AnomalyResult randomMutlEntityAnomalyDetectResult(double score, do
             Instant.now().truncatedTo(ChronoUnit.SECONDS),
             Instant.now().truncatedTo(ChronoUnit.SECONDS),
             error,
-            Arrays.asList(new Entity(randomAlphaOfLength(5), randomAlphaOfLength(5))),
+            Entity.createSingleAttributeEntity(randomAlphaOfLength(5), randomAlphaOfLength(5), randomAlphaOfLength(5)),
```
ohltyler, line 640: Why create a single attribute entity for a function that looks like it should return a multi-entity result?

src/main/java/org/opensearch/ad/model/Entity.java
kaituo: If we moved this after the switch, it would reset for every field.  Here, I want to reset every time I have seen a name-value pair.

src/main/java/org/opensearch/ad/model/Entity.java
kaituo: The consequence is we skip the entity and not analyze it.  The logic is the same as before.

src/main/java/org/opensearch/ad/transport/GetAnomalyDetectorRequest.java
kaituo: right, changed

src/main/java/org/opensearch/ad/transport/GetAnomalyDetectorTransportAction.java
kaituo: yes, changed

src/test/java/org/opensearch/ad/TestHelpers.java
kaituo: Multi-entity means a detector with at least one category fields.  It is different form multi-attribute entity that means more than one category field. I changed the function name from randomMutlEntityAnomalyDetectResult to randomHCADAnomalyDetectResult to make it clearer.

src/main/java/org/opensearch/ad/model/Entity.java
ohltyler: I was thinking after the switch, but still within the `while` loop, so it would only reset after `parsedName` and `parsedField` were populated. Seems ok to leave as-is, just wanting to clarify

ylwu-amzn: Seems this logic can't work if `ATTRIBUTE_VALUE_FIELD` is before `ATTRIBUTE_NAME_FIELD`.  @ohltyler , do you mean change to this
```
        // Iterate every entity attribute pair
        while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
            ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
            
            String parsedValue = null;
            String parsedName = null;
            // parse current entity attribute pair
            while (parser.nextToken() != Token.END_OBJECT) {
                String fieldName = parser.currentName();
                // move to the field value
                parser.nextToken();
                switch (fieldName) {
                    case ATTRIBUTE_NAME_FIELD:
                        parsedName = parser.text();
                        break;
                    case ATTRIBUTE_VALUE_FIELD:
                        parsedValue = parser.text();
                    default:
                        break;
            }

            if (parsedName != null && parsedValue != null) {
                entities.put(parsedName, parsedValue);
            }
        }
```

ohltyler: @ylwu-amzn almost, except still clear the `parsedValue` and `parsedName` in the `if` block

kaituo: right.  Changed.

ohltyler: @ylwu-amzn I was thinking within the inner `while` loop still, like this:
```
        // Iterate every entity attribute pair
        while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
            ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
            
            // parse current entity attribute pair
            while (parser.nextToken() != Token.END_OBJECT) {
                String fieldName = parser.currentName();
                // move to the field value
                parser.nextToken();
                switch (fieldName) {
                    case ATTRIBUTE_NAME_FIELD:
                        parsedName = parser.text();
                        break;
                    case ATTRIBUTE_VALUE_FIELD:
                        parsedValue = parser.text();
                    default:
                        break;
               }
               if (parsedName != null && parsedValue != null) {
                   entities.put(parsedName, parsedValue);
                   parsedName = null;
                   parsedValue = null;
               }
            }
        }
```

ohltyler (approved):
LGTM

src/main/java/org/opensearch/ad/AnomalyDetectorRunner.java
```diff
@@ -136,14 +144,17 @@ private void onFailure(Exception e, ActionListener<List<AnomalyResult>> listener
         // TODO return exception like IllegalArgumentException to explain data is not enough for preview
         // This also requires front-end change to handle error message correspondingly
         // We return empty list for now to avoid breaking front-end
+        if (e instanceof OpenSearchSecurityException) {
+            listener.onFailure(e);
+        }
```
ylwu-amzn, line 149: If exception is `OpenSearchSecurityException`, will call `onFailure` first, then call `onResponse`, it will throw connection closed exception.

Add `return` in this `if` block, or put `listener.onResponse(Collections.emptyList());` in `else` block.

kaituo: yeah, fixed this one and another similar bug.

ylwu-amzn (approved):
LGTM

Issue #36, opened 2021-05-05
[BUG]Block creating/updating detector if user has no permission to read index


ylwu-amzn:
**Describe the bug**
If user has no permission to read index, we should block creating/updating detector with that index.

We should [check index](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportAction.java#L194) before [stashing thread context](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportAction.java#L102) .
The logic was correct in ODFE 1.11 check https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/opendistro-1.11/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportAction.java#L88

ylwu-amzn:
fixed

PR #80, opened 2021-05-27
check if user has permission to read index when create/update/preview detector

ylwu-amzn:
### Description
If user has no permission to read index, we should block creating/updating detector with that index.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/36
 
### Check List


saratvemulapalli (approved):
The change looks good to me.
Can we add an integration test to cover this?

ohltyler (approved):
Overall LGTM, 2 things though:
1. +1 to adding a test to cover this
2. Should this be targeted to add in the RC release?

PR #82, opened 2021-05-27
Remove error in the state when there is no error anymore.

kaituo:
Note: since there are a lot of dependencies, I only list the main class and test code to save reviewers' time. The build will fail due to missing dependencies. I will use that PR just for review. will not merge it. Will have a big one in the end and merge once after all review PRs get approved. Now the code is missing unit tests.  Will add unit tests, run performance tests, and fix bugs before the official release.

### Description
Previously, we won't reset error in the state when a detector changes from an error state to a non-error state. This PR fixes that.

Testing done:
* Tested on a cluster the problem is gone with the fix.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
@@ -468,12 +468,14 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
         String detectorId = jobParameter.getName();
         detectorEndRunExceptionCount.remove(detectorId);
         try {
+            // reset error if different from previously recorded one
+            detectionStateHandler.saveError(response.getError(), detectorId);
             // skipping writing to the result index if not necessary
             // For a single-entity detector, the result is not useful if error is null
             // and rcf score (thus anomaly grade/confidence) is null.
-            // For a multi-entity detector, we don't need to save on the detector level.
-            // We always return 0 rcf score if there is no error.
-            if (response.getAnomalyScore() <= 0 && response.getError() == null) {
+            // For a HCAD detector, we don't need to save on the detector level.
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -468,12 +468,14 @@ private void indexAnomalyResult(
...
+            detectionStateHandler.saveError(response.getError(), detectorId);
             // skipping writing to the result index if not necessary
             // For a single-entity detector, the result is not useful if error is null
             // and rcf score (thus anomaly grade/confidence) is null.
-            // For a multi-entity detector, we don't need to save on the detector level.
-            // We always return 0 rcf score if there is no error.
-            if (response.getAnomalyScore() <= 0 && response.getError() == null) {
+            // For a HCAD detector, we don't need to save on the detector level.
+            // We return 0 or Double.NaN rcf score if there is no error.
+            if ((response.getAnomalyScore() <= 0 || Double.isNaN(response.getAnomalyScore())) && response.getError() == null) {
```
ylwu-amzn, line 478: From the description
> Previously, we won't reset error in the state when a detector changes from an error state to a non-error state. This PR fixes that.

Should we set error as null in this if block?

kaituo: You meant call saveError under this branch?

spbjss (approved):
looks good to me.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
ylwu-amzn: Oh, I see your force pushed a new commit. Not sure what the original code looks like. But current code looks ok for me.

ylwu-amzn (approved):
LGTM

PR #83, opened 2021-05-27
Adjust settings for HCAD v2

kaituo:

### Description
This PR made a few changes on settings:
* replaced AnomalyDetectorSettings.AD_RESULT_HISTORY_MAX_DOCS with AD_RESULT_HISTORY_MAX_DOCS_PER_SHARD because the value of the AD_RESULT_HISTORY_MAX_DOCS is too low for rolling over the result index. Using AD_RESULT_HISTORY_MAX_DOCS, we rolled over the result index when the primary shards' total size is 20+ GB with three shards. We should allow the size to reach 65 GB per shard before rolling over.
* removed AnomalyDetectorSettings.MAX_CACHE_MISS_HANDLING_PER_SECOND as we changed the implementation of rate-limiting. The setting is not in effect anymore.
* Added settings like category_field_limit for the HCAD v2.

This PR also put low coverage tests in the jacoco exclusion list. I will add tests in the following PRs.

Testing done:
* test newly added and a few old setting. They work as expected.
 
### Check List


src/main/java/org/opensearch/ad/settings/AbstractSetting.java
```diff
@@ -0,0 +1,97 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ylwu-amzn, line 13: Remove this old license header

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+
+/**
+ * A container serving dynamic numeric setting.  The caller does not have to call
```
ylwu-amzn, line 41: Why only support `numeric` setting? Can we support other types of setting like boolean?

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Get setting value by key. Return default value if not configured explicitly.
+     *
+     * @param key   setting key.
+     * @param <T> Setting type
+     * @return T     setting value or default
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T getSettingValue(String key) {
+        return (T) latestSettings.getOrDefault(key, getSetting(key).getDefault(Settings.EMPTY));
```
ylwu-amzn, line 84: Can we put all setting's initial value into `latestSettings` map when `init` ? So we can just get setting values from `latestSettings`.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -100,14 +100,18 @@ private AnomalyDetectorSettings() {}
...
         .longSetting(
-            "plugins.anomaly_detection.ad_result_history_max_docs",
-            // Total documents in primary replica.
-            // A single feature result is roughly 150 bytes. Suppose a doc is
-            // of 200 bytes, 250 million docs is of 50 GB. We choose 50 GB
-            // because we have 1 shard at least. One shard can have at most 50 GB.
-            LegacyOpenDistroAnomalyDetectorSettings.AD_RESULT_HISTORY_MAX_DOCS,
+            "plugins.anomaly_detection.ad_result_history_max_docs_per_shard",
+            // Total documents in the primary shards.
+            // Note the count is for Lucene docs. Lucene considers a nested
```
ylwu-amzn, line 110: This number is used in `CreateIndexRequest#addMaxIndexDocsCondition` for adding condition to check if the index has at least numDocs. Does it mean the lucene doc count or ES index doc count?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -275,18 +279,52 @@ private AnomalyDetectorSettings() {}
...
+     * DEDICATED_CACHE_SIZE is a setting to make dedicated cache's size flexible.
+     * When that setting is changed, if the size decreases, we will release memory
+     * if required (e.g., when a user also decreased AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE,
+     * the max memory percentage that AD can use);
+     * if the size increases, we may reject the setting change if we cannot fulfill
+     * that request (e.g., when it will uses more memory than allowed for AD).
+     *
+     * With compact rcf, rcf with 30 trees and shingle size 4 is of 500KB.
+     * The recommended max heap size is 32 GB. Even if users use all of the heap
+     * for AD, the max number of entity model cannot surpass 3.2 * 10^10 / 5*10^5 = 6.4 * 10 ^4
```
ylwu-amzn, line 307: Takes some time to understand these numbers in `3.2 * 10^10 / 5*10^5`. I think you mean 3.2 GB/500KB  and 3.2 GB is from 10% memory limit of AD plugin?

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
kaituo: removed

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
kaituo: boolean setting is in the class EnabledSetting.  I am trying to follow https://en.wikipedia.org/wiki/Single-responsibility_principle

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
kaituo: If I understand you correctly, I have already done that.  init calls setSettingsUpdateConsumers that does the thing you asked.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: Lucene doc count. let me add comments for this.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: yes, added.

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
kaituo: changed according to your suggestion. Please check the new commit.

kaituo: turns out all integration tests fail due to the above change.  I think it is because our rest service has no access to ClusterService and thus we didn't call AbstractSetting,init in our rest test case set up.

This is the changed code:
https://gist.github.com/kaituo/13ed77f2d7e2c92bc66877cc55d2b373

Then I tried to move latestSetting initialization to the constructor and comment out settings.  But there are compile errors.  Notice setting is of type Map<String, Setting<?>>, while latestSettings is of type Map<String, Object>. 

https://gist.github.com/kaituo/2b15a8279dd8534add5e6c69ffbcac15

Please advise how to proceed.

build.gradle
```diff
@@ -1,487 +0,0 @@
-/*
```
kaituo, line 1: removed this file as changes have been reviewed in another PR

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
```diff
@@ -0,0 +1,82 @@
...
+    private final Map<String, Setting<?>> settings;
+
+    protected AbstractSetting(Map<String, Setting<?>> settings) {
+        this.settings = settings;
+    }
+
+    private void setSettingsUpdateConsumers() {
+        for (Setting<?> setting : settings.values()) {
+            clusterService.getClusterSettings().addSettingsUpdateConsumer(setting, newVal -> {
+                logger.info("[AD] The value of setting [{}] changed to [{}]", setting.getKey(), newVal);
```
spbjss, line 49: here any children classes of AbstractSetting will always use the logger defined with AbstractSetting 
private static Logger logger = LogManager.getLogger(AbstractSetting.class);

Following in an example on how to use different logger for the children classes:

```
public class TestA {
    protected String log = "logA";

    private void internalMethod() {
        System.out.println(log);
    }

    public void externalMethod() {
        internalMethod();
    }
}

public class TestB extends TestA {
    public TestB() {
        log = "logB";
    }
}

public class Test {
    public static void main(String[] args) {
        TestA test = new TestB();
        test.externalMethod();
    }
}
```

spbjss (commented):
Several small comments

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -396,6 +396,61 @@ private static Void initGson() {
...
+            clientUtil,
+            CommonName.CHECKPOINT_INDEX_NAME,
+            gson,
+            rcfSerde,
+            HybridThresholdingModel.class,
+            anomalyDetectionIndices,
+            AnomalyDetectorSettings.MAX_CHECKPOINT_BYTES
+        );
+
+        Random random = new Random(42);
```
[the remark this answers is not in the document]
kaituo: it can be any random number.  Doesn't have to be 42.

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -618,7 +746,9 @@ protected Clock getClock() {
                 new ScalingExecutorBuilder(
                     AD_THREAD_POOL_NAME,
                     1,
-                    Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) / 4),
+                    // HCAD can be heavy after supporting 1 million entities.
+                    // Limit to use at most half of the processors.
+                    Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) / 2),
```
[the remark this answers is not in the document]
kaituo: The threadpool size is determined at the beginning of node creation.  I don't know how to replace it with a new threadpool with another size dynamically.  Could you tell me how to do that?

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -681,15 +812,38 @@ protected Clock getClock() {
...
+                AnomalyDetectorSettings.COLD_ENTITY_QUEUE_CONCURRENCY,
+                AnomalyDetectorSettings.DEDICATED_CACHE_SIZE,
+                AnomalyDetectorSettings.COLD_ENTITY_QUEUE_MAX_HEAP_PERCENT,
+                AnomalyDetectorSettings.CHECKPOINT_READ_QUEUE_MAX_HEAP_PERCENT,
+                AnomalyDetectorSettings.CHECKPOINT_WRITE_QUEUE_MAX_HEAP_PERCENT,
+                AnomalyDetectorSettings.RESULT_WRITE_QUEUE_MAX_HEAP_PERCENT,
+                AnomalyDetectorSettings.ENTITY_COLD_START_QUEUE_MAX_HEAP_PERCENT,
+                AnomalyDetectorSettings.EXPECTED_COLD_ENTITY_EXECUTION_TIME_IN_SECS,
+                AnomalyDetectorSettings.MAX_ENTITIES_PER_INTERVAL,
+                AnomalyDetectorSettings.PAGE_SIZE
```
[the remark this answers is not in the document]
kaituo: yes, please check the new commit.

src/main/java/org/opensearch/ad/settings/NumericSetting.java
```diff
@@ -0,0 +1,85 @@
+/*
```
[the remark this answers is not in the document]
kaituo: yes

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
kaituo: the logger is a static field and cannot be set in constructor.  So I need to put the reassignment in a static block.

```
class TestA {
    protected static String log = "logA";

    private void internalMethod() {
        System.out.println(log);
    }

    public void externalMethod() {
        internalMethod();
    }
}
 class TestB extends TestA {
     {
         log = "logB";
     }
    public TestB() {
    }
}

public class Main {
    public static void main(String[] args) {
        TestA test = new TestB();
        test.externalMethod();
    }
}
```

I can use the setting name to differentiate which subclass it is.  Will use your technique in the future when needed :)

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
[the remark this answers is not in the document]
spbjss: Do we have a way for static config? something like the brazil config?
If yes, it's better to move it to the static config, otherwise, it's fine to keep the current code.

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
spbjss: Cool!

spbjss (approved):
looks good to me.

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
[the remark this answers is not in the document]
kaituo: We don't have a .config file :)

src/main/java/org/opensearch/ad/settings/AbstractSetting.java
ylwu-amzn: Let's keep as is. Seems not easy to change.

ylwu-amzn (approved):
LGTM

PR #88, opened 2021-06-02
Run integration test against docker image in CI

ylwu-amzn:
### Description
Pull docker image and run integration test in CI.
 
### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -81,6 +81,54 @@ jobs:
         run: |
           ./gradlew publishToMavenLocal -Dopensearch.version=1.0.0-rc1
 
+      - name: Multi Nodes Integration Testing
+        run: |
+          ./gradlew integTest  -PnumNodes=3
+
+      - name: Pull and Run Docker
+        run: |
+          ## plugin=`ls build/distributions/*.zip`
+          ## version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-3`
+          ## plugin_version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-4`
+          ## TODO: remove these two hard code versions below after GA release
+          version=1.0.0-rc1
+          plugin_version=1.0.0.0-rc1
+          echo $version
+          cd ..
+          if docker pull opensearchstaging/opensearch:$version
+          then
+            echo "FROM opensearchstaging/opensearch:$version" >> Dockerfile
+            ## The ESRestTest Client uses http by default.
+            ## Need to disable the security plugin to call the rest api over http.
+            echo "RUN if [ -d /usr/share/opensearch/plugins/opensearch-anomaly-detection ]; then /usr/share/opensearch/bin/opensearch-plugin remove opensearch-anomaly-detection; fi" >> Dockerfile
+            echo "ADD anomaly-detection/build/distributions/opensearch-anomaly-detection-$plugin_version.zip /tmp/" >> Dockerfile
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -81,6 +81,54 @@ jobs:
...
+          ## plugin_version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-4`
+          ## TODO: remove these two hard code versions below after GA release
+          version=1.0.0-rc1
+          plugin_version=1.0.0.0-rc1
+          echo $version
+          cd ..
+          if docker pull opensearchstaging/opensearch:$version
+          then
+            echo "FROM opensearchstaging/opensearch:$version" >> Dockerfile
+            ## The ESRestTest Client uses http by default.
```
saratvemulapalli, line 101: Super nit: 
```suggestion
            ## The OpenSearchRestTest Client uses http by default.
```

saratvemulapalli (approved):
Finally! Nice to have all the tests which were run in ODFE :)

.github/workflows/CI.yml
```diff
@@ -81,6 +81,54 @@ jobs:
...
+        if: env.imagePresent == 'true'
+        run: |
+          cd ..
+          docker run -p 9200:9200 -d -p 9600:9600 -e "discovery.type=single-node" opensearch-ad:test
+          sleep 90
+
+      - name: Run AD Test
+        if: env.imagePresent == 'true'
+        run: |
+          security=`curl -XGET https://localhost:9200/_cat/plugins?v -u admin:admin --insecure |grep opensearch-security|wc -l`
```
ohltyler, line 122: Since the current staging docker images are all bundled and will have security, should we change this to run CI against both security disabled and security enabled? For example, in AD Dashboards, we run the test suite for both scenarios. To test without security, we remove the security plugin from the image and run the tests (see [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/.github/workflows/e2e-tests-workflow.yml#L114))

ylwu-amzn: Good point. Will add test with security disabled.
I don't think it will fail security disabled case if security enabled case can pass. Anyway, it's good to cover more scenarios.

.github/workflows/CI.yml
```diff
@@ -81,6 +81,54 @@ jobs:
...
+          ## TODO: remove these two hard code versions below after GA release
+          version=1.0.0-rc1
+          plugin_version=1.0.0.0-rc1
+          echo $version
+          cd ..
+          if docker pull opensearchstaging/opensearch:$version
+          then
+            echo "FROM opensearchstaging/opensearch:$version" >> Dockerfile
+            ## The ESRestTest Client uses http by default.
+            ## Need to disable the security plugin to call the rest api over http.
```
kaituo, line 102: The comment does not apply  to the code.

.github/workflows/CI.yml
ylwu-amzn: removed these two comments

.github/workflows/CI.yml
ylwu-amzn: removed

.github/workflows/CI.yml
ylwu-amzn: @ohltyler Have no bandwidth now to work on the security disabled case. Will create an issue and add it later.

ylwu-amzn: https://github.com/opensearch-project/anomaly-detection/issues/91

spbjss (approved):
lgtm

PR #90, opened 2021-06-02
Fix empty range for cold start bug

kaituo:
### Description
While collecting training data, we continue searching even if no data range has data.  This bug can cause "No [ranges] specified for the [date_range] aggregation" exception printed out.  This PR fixed that by returning empty training data instead of logging exceptions.

Testing done:
1. testing on a cluster with sparse data.  Verified we don't log exceptions anymore.
 
### Check List


ylwu-amzn (approved):
LGTM

ohltyler (approved):

Issue #87, opened 2021-06-01
HCAD v2: support multiple category fields


kaituo:
**Is your feature request related to a problem? Please describe.**
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/404

**Describe the solution you'd like**
Supporting multiple categorical fields requires changes in, among other things, feature aggregations and model naming. We will switch to composite aggregations that natively support collecting entities from multiple fields in the same document.  Composite aggregations enable us to aggregate features over multiple categorical fields with minimal programming effort.

Another change is to not use categorical values directly in the model document Id.  The problem is that HCAD v1 uses the categorical value as part of the model document Id, but ES’s document Id can be at most 512 bytes. Categorical values are usually less than 256 characters, but can grow to 32766 (https://stackoverflow.com/questions/47177163/what-is-the-maximum-length-for-keyword-type-in-elasticsearch) in theory.  HCAD v1 skips an entity if the entity's name is more than 256 characters. We cannot do that in v2 as that can reject a lot of entities.  To overcome the obstacle, we hash categorical values to a 128-bit string (like SHA-1 that git uses) and use the hash as part of the model document Id. 

**Describe alternatives you've considered**
We have choices to make regarding when to use the hash as part of a model document Id: for all HC detectors or a HC detector with multiple categorical fields. The challenge lies in providing backward compatibility of looking for a model checkpoint in the case of a HC detector with one categorical field.  If using hashes for all HC detectors, we need two get requests to ensure that a model checkpoint exists.  One uses the document Id without a hash, while one uses the document Id with a hash.  The dual get requests are ineffective.  If limiting hashes to a HC detector with multiple categorical fields, there is no backward compatibility issue.  However, the code will be branchy.  One may wonder if backward compatibility can be ignored; indeed, the old checkpoints will be gone after a transition period during upgrading. During the transition period, HC detectors can experience unnecessary cold starts as if the detectors were just started. Checkpoint index size can double if every model has two model documents. The transition period can be as long as 3 days since our checkpoint retention period is 3 days.  There is no perfect solution.  We prefer limiting hashes to an HC detector with multiple categorical fields as its customer impact is none.

PR #93, opened 2021-06-08
Using a painless script to simulate multiple terms aggregation

kaituo:
### Description
We use composite aggregation for feature aggregation. But composite aggregation does not support ordering results based on doc count, which is required by preview API and historical detector components. We need to use terms aggregation. Terms aggregation does not support collecting terms from multiple fields in the same document. Scripts come to the rescue: With a script to retrieve terms from multiple fields, we can use terms aggregation to partition data. The script disables the global ordinals optimization and will be slower than collecting terms from a single field. Still, it gives us the flexibility to implement this option at search time. For a simple example, consider the following query about the number field’s sum aggregation on buckets partitioned by category_field_1 and category_field_2 from the index test.

Testing done:
* Verified preview works after the change.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/87
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/feature/ScriptMaker.java
+++ b/src/main/java/org/opensearch/ad/feature/ScriptMaker.java
@@ -1,0 +1,118 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package org.opensearch.ad.feature;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+
+import org.opensearch.script.Script;
+import org.opensearch.script.ScriptType;
+
+public class ScriptMaker {
+    private static final String template = "doc['%s'].value";
+
+    /**
+     * We use composite aggregation for feature aggregation.  But composite aggregation
+     * does not support ordering results based on doc count, which is required by
+     * preview and historical related components.  We need to use terms aggregation.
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/ad/feature/ScriptMaker.java
```diff
@@ -0,0 +1,118 @@
...
+        // in painless, a map is sth like [a:b, c:d]
+        format.append("[");
+        for (int i = 0; i < fields.size(); i++) {
+            if (i > 0) {
+                format.append(",");
+            }
+            format.append("\'");
+            format.append(fields.get(i));
+            format.append("\':");
+            format.append(String.format(Locale.ROOT, template, fields.get(i)));
```
ohltyler, line 113: Rather than handling only part of this via string template, can you change to handle the entire mapping as a template? Example template: `"\'%s\': doc['%s'].value"`

kaituo: yeah, done

src/main/java/org/opensearch/ad/feature/ScriptMaker.java
```diff
@@ -9,21 +9,6 @@
...
- * A copy of the License is located at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * or in the "license" file accompanying this file. This file is distributed
- * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
- * express or implied. See the License for the specific language governing
- * permissions and limitations under the License.
- */
-
```
ohltyler: Each file should still have both SPDX + Amazon license headers

ohltyler (approved):
LGTM.

src/main/java/org/opensearch/ad/feature/ScriptMaker.java
ohltyler: ^ as confirmed offline, only need SPDX header

src/main/java/org/opensearch/ad/feature/ScriptMaker.java
```diff
@@ -0,0 +1,100 @@
...
+                        }
+                    },
+              ...
+       *
+       * I put two categorical field in a map for parsing the results.  Otherwise,
+       * I won't know which categorical value is for which field.
+     * @param fields categorical fields
+     * @return script to use in terms aggregation
+     */
+    public static Script makeTermsScript(List<String> fields) {
```
ylwu-amzn, line 87: Just confirm as I don't see how this class will be used in this PR. will we use this script for single category filed case? I think painless script query's performance is not so good.

kaituo: It is used here: https://github.com/kaituo/anomaly-detection-1/blob/multiCat4/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java#L202

Only use it for multi category field case.

Yeah, that's the only option if I continue to use terms aggregation. It is only used in Preview API.  So the performance impact is not wide.

ylwu-amzn (approved):

jmazanec15 (approved):

Issue #85, opened 2021-06-01
HCAD v2: Better rate limiting


kaituo:
**Is your feature request related to a problem? Please describe.**
Our current rate limiting is too coarse and simple.  For example, HCAD v1 loads at most 100 model checkpoints from disks per second per node.  The 100 limit means HCAD throws away concurrent checkpoint reads beyond 100.  Having more than 100 reads per second is possible. HCAD can bombard ES with “thundering herd” traffic, in which many entities make requests that need similar ES reads/writes at approximately the same time. We chose 100 since it does not cause ES throttling during our performance testing. Still, we can observe 4~5 times query slowdown in esrally’s HTTP logs workload with the setting. 

**Describe the solution you'd like**
To amortize the extra operations imposed by the above change  (e.g., expensive requests like cache misses or writes),  we use in-memory queues to buffer pent-up requests for various work stages like reading checkpoints from disks.

PR #94, opened 2021-06-08
Refactor ModelManager.

kaituo:
### Description
First, we move maybeTrainBeforeScore to EntityColdStarter for code reuse in the entity cold start queue.
Second, we add logic to deal with null models and don’t assume the callers will ensure that.
Third, we make processEntityCheckpoint a pure function that does not change input parameters.

Testing done:
* Verified after refactoring, the basic workflow of HCAD still works.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/85
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/ModelManager.java
+++ b/src/main/java/org/opensearch/ad/ml/ModelManager.java
@@ -967,6 +967,7 @@ public class ModelManager implements DetectorModelSize {
                 modelState.setModel(entityModel);
             }
 
+            // trainModelFromExistingSamples may be able to make models not null
             if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
                 entityColdStarter.trainModelFromExistingSamples(modelState);
             }
--- a/src/main/java/org/opensearch/ad/ml/ModelManager.java
+++ b/src/main/java/org/opensearch/ad/ml/ModelManager.java
@@ -61,6 +61,7 @@ import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.feature.FeatureManager;
 import org.opensearch.ad.ml.rcf.CombinedRcfResult;
 import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.ad.model.Entity;
 
 import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.returntypes.DiVector;
@@ -149,7 +150,7 @@ public class ModelManager implements DetectorModelSize {
      * @param minPreviewSize minimum number of data points for preview
      * @param modelTtl time to live for hosted models
      * @param checkpointInterval interval between checkpoints
-     * @param entityColdStarter Used train models on input data
+     * @param entityColdStarter HCAD cold start utility
      * @param modelPartitioner Used to partition RCF models
      * @param featureManager Used to create features for models
      * @param memoryTracker AD memory usage tracker
@@ -178,7 +179,6 @@ public class ModelManager implements DetectorModelSize {
         FeatureManager featureManager,
         MemoryTracker memoryTracker
     ) {
-
         this.rcfSerde = rcfSerde;
         this.checkpointDao = checkpointDao;
         this.gson = gson;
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -944,102 +944,121 @@ public void getTotalUpdates(String modelId, String detectorId, ActionListener<Lo
...
+            if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
+                entityColdStarter.trainModelFromExistingSamples(modelState);
+            }
+
+            if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
+                return score(datapoint, modelId, modelState);
+            } else {
+                entityModel.addSample(datapoint);
+                return new ThresholdingResult(0, 0, 0);
+            }
```
ohltyler, line 979: These statements can be simplified:
```suggestion
            if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
                return score(datapoint, modelId, modelState);
            } else {
                entityColdStarter.trainModelFromExistingSamples(modelState);
                entityModel.addSample(datapoint);
                return new ThresholdingResult(0, 0, 0);
            }
```

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -944,102 +944,121 @@ public void getTotalUpdates(String modelId, String detectorId, ActionListener<Lo
...
         } else {
-            result = new ThresholdingResult(0, 0, 0);
+            return new ThresholdingResult(0, 0, 0);
         }
-
-        return result;
     }
 
-    private ThresholdingResult score(Queue<double[]> samples, String modelId, ModelState<EntityModel> modelState) {
+    public ThresholdingResult score(double[] feature, String modelId, ModelState<EntityModel> modelState) {
```
ohltyler, line 985: Can you explain why this method changed? Before it was taking the last feature from `samples` to generate a result, now it's taking `feature` directly?

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: trainModelFromExistingSamples may be able to make "entityModel.getRcf() != null && entityModel.getThreshold() != null" true.  Added comments to explain.

ohltyler: Ah, understood. Thanks

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: Changed because I want the caller to tell me which point to score instead the method making an assumption that the last point is the point to return score for.

ohltyler (approved):
LGTM

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -944,102 +944,122 @@ public void getTotalUpdates(String modelId, String detectorId, ActionListener<Lo
...
 
-            result = maybeTrainBeforeScore(modelState, entityName);
+            // trainModelFromExistingSamples may be able to make models not null
+            if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
+                entityColdStarter.trainModelFromExistingSamples(modelState);
+            }
+
+            if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
+                return score(datapoint, modelId, modelState);
+            } else {
```
jmazanec15, line 977: Else is unnecessary because if always returns. Any reason to keep it?

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -944,102 +944,122 @@ public void getTotalUpdates(String modelId, String detectorId, ActionListener<Lo
...
+                entityModel.addSample(datapoint);
+                return new ThresholdingResult(0, 0, 0);
+            }
         } else {
-            result = new ThresholdingResult(0, 0, 0);
+            return new ThresholdingResult(0, 0, 0);
```
jmazanec15, line 982: Else is unnecessary, because if always returns. Same as above.

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: Mostly for readability.  Some discussions here: https://stackoverflow.com/questions/3261849/why-is-else-rarely-used-after-if-x-then-return

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: Yeah, for readability's purpose.  There is no absolute rule which one is preferred.

jmazanec15 (approved):

PR #95, opened 2021-06-08
Reuse the error message of the multi-category limit surpassed error

kaituo:
### Description
We have a limit of the number of categorical fields and report an error if the limit is breached. This PR moves the message to CommonErrorMessage and reuses it in multiple places.

Testing done:
* Verified the basic workflow of HCAD works
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/87
 
### Check List
- [ Y ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #92, opened 2021-06-04
Refactoring related to result and checkpoint write worker.

kaituo:
### Description
We have added queues for writing model checkpoints and detection results. This PR contains refactoring related to those queues.
First, we have moved the index request clone from AnomalyIndexHandler to BulkUtil so that it can be reused.
Second, we have moved exception handling from MultiEntityResultHandler to queues to have a central place to handle exceptions.

Testing one:
* Verified queues can handle exceptions.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/85
 
### Check List
- [ Y ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #86, opened 2021-06-01
HCAD v2: add pagination


kaituo:
**Is your feature request related to a problem? Please describe.**
The major problem of HCAD is scalability. HCAD issues a query to fetch feature data for each entity regularly. HCAD v1 limit the number of entity values returned per query to 1000, both for reasons of efficiency (fetching the overall entity values is expensive and may not finish before the next scheduled query run starts) and for reasons of ease-of-load-shedding (the top 1000 limit, in turn, curtails the memory used to host models, the disk access needed to read/write model checkpoints and anomaly results, the CPU used for entity metadata maintenance and model training/inference, and the garbage collection for deleted models and metadata). Users can increase the maximum entity setting at the cost of a linear growth resource usage, which opens the door to cluster instability.  However, such a heuristic “best-effort” limit is detrimental to the scalability of entity monitoring.  Even if a user scales out, the number of monitored entities does not increase.

**Describe the solution you'd like**
We increase the default maximum entities returned by the feature query from one thousand to one million.  The huge number of entities are handled via pagination.

PR #84, opened 2021-05-27
Record all exceptions instead of just cold start exceptions in state

kaituo:
Note: since there are a lot of dependencies, I only list the main class and test code to save reviewers' time. The build will fail due to missing dependencies. I will use that PR just for review. will not merge it. Will have a big one in the end and merge once after all review PRs get approved. Now the code is missing unit tests. Will add unit tests, run performance tests, and fix bugs before the official release.

### Description
This PR record all exceptions instead of just cold start exceptions in node state.

Besides, this PR also merges exceptions if HCAD models send multiple exceptions. When merging an exception, EndRunException has priority since it can stop detectors. Our rate-limited queues need these changes to pass back exceptions asynchronously and make the exceptions visible to stats API and job scheduler.

Testing done:
* Tested our rate-limited queues can use NodeStateManager.setException to stop a detector.
 
### Check List
- [ Y ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #121, opened 2021-07-06
multi-category support, rate limiting, and pagination

kaituo:
### Description
This PR is a conglomerate of the following PRs.

https://github.com/opensearch-project/anomaly-detection/pull/60
https://github.com/opensearch-project/anomaly-detection/pull/64
https://github.com/opensearch-project/anomaly-detection/pull/65
https://github.com/opensearch-project/anomaly-detection/pull/67
https://github.com/opensearch-project/anomaly-detection/pull/68
https://github.com/opensearch-project/anomaly-detection/pull/69
https://github.com/opensearch-project/anomaly-detection/pull/70
https://github.com/opensearch-project/anomaly-detection/pull/71
https://github.com/opensearch-project/anomaly-detection/pull/74
https://github.com/opensearch-project/anomaly-detection/pull/75
https://github.com/opensearch-project/anomaly-detection/pull/76
https://github.com/opensearch-project/anomaly-detection/pull/77
https://github.com/opensearch-project/anomaly-detection/pull/78
https://github.com/opensearch-project/anomaly-detection/pull/79
https://github.com/opensearch-project/anomaly-detection/pull/82
https://github.com/opensearch-project/anomaly-detection/pull/83
https://github.com/opensearch-project/anomaly-detection/pull/84
https://github.com/opensearch-project/anomaly-detection/pull/92
https://github.com/opensearch-project/anomaly-detection/pull/94
https://github.com/opensearch-project/anomaly-detection/pull/93
https://github.com/opensearch-project/anomaly-detection/pull/95
https://github.com/kaituo/anomaly-detection-1/pull/1
https://github.com/kaituo/anomaly-detection-1/pull/2
https://github.com/kaituo/anomaly-detection-1/pull/3
https://github.com/kaituo/anomaly-detection-1/pull/4/
https://github.com/kaituo/anomaly-detection-1/pull/5
https://github.com/kaituo/anomaly-detection-1/pull/6
https://github.com/kaituo/anomaly-detection-1/pull/7
https://github.com/kaituo/anomaly-detection-1/pull/8
https://github.com/kaituo/anomaly-detection-1/pull/9
https://github.com/kaituo/anomaly-detection-1/pull/10

This spreadsheet contains the mappings from files to PR number (bug fix in my AD fork and tests are not included):
https://gist.github.com/kaituo/9e1592c4ac4f2f449356cb93d0591167

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/85 (rate limiting)
https://github.com/opensearch-project/anomaly-detection/issues/87 (multiple category fields)
https://github.com/opensearch-project/anomaly-detection/issues/86 (pagination)
 
### Check List


src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
```diff
@@ -129,4 +143,35 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+        if (Strings.isEmpty(detectorId)) {
+            throw new IllegalStateException(CommonErrorMessages.AD_ID_MISSING_MSG);
+        }
+
+        String entityName = request.param(CommonName.CATEGORICAL_FIELD);
+        String entityValue = request.param(CommonName.ENTITY_KEY);
+
+        if (entityName != null && entityValue != null) {
+            // single-stream profile request:
+            // GET _opendistro/_anomaly_detection/detectors/<detectorId>/_profile/init_progress?category_field=<field-name>&entity=<value>
```
ohltyler, line 157: nit: `_opendistro` -> `_plugins`

src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
```diff
@@ -129,4 +143,35 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
+            return Entity.createSingleAttributeEntity(detectorId, entityName, entityValue);
+        } else if (request.hasContent()) {
+            /* HCAD profile request:
+             * GET _opendistro/_anomaly_detection/detectors/<detectorId>/_profile/init_progress
```
ohltyler, line 161: nit: `_opendistro` -> `_plugins`

ylwu-amzn:
Some high level question which not covered in previous PR review.

Current design will dispatch entities to different work nodes with consistent hashing. Each work node will cache entities in queue and run RCF model, then bulk write anomaly result to index. If current detection interval timeout, we just drop all remaining entities in queue to avoid carry over to next interval. It's possible that some worker node may receive more entities which needs to run longer like need to load checkpoints. So the unbalanced load may cause some worker node drops more entities. But seems not a big concern as we don't guarantee all entities will be detected, there are other reasons can cause dropping entities like worker node queue is full. 

Have some concern about the horizontal scalability. For example, if user scale cluster from 10 data nodes to 100 data nodes, then will have 100 worker nodes running entities, that means we may send out 100 bulk write AD result requests concurrently. But the AD result index shard setting is still the same, that may bring higher write load and user's general bulk write request may impacted. Confirm with PM/SDM if the concern of horizontal scalable issue is reasonable.

src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
kaituo: fixed

src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
kaituo: fixed

kaituo:
> Some high level question which not covered in previous PR review.
> 
> Current design will dispatch entities to different work nodes with consistent hashing. Each work node will cache entities in queue and run RCF model, then bulk write anomaly result to index. If current detection interval timeout, we just drop all remaining entities in queue to avoid carry over to next interval. It's possible that some worker node may receive more entities which needs to run longer like need to load checkpoints. So the unbalanced load may cause some worker node drops more entities. But seems not a big concern as we don't guarantee all entities will be detected, there are other reasons can cause dropping entities like worker node queue is full.
> 
> Have some concern about the horizontal scalability. For example, if user scale cluster from 10 data nodes to 100 data nodes, then will have 100 worker nodes running entities, that means we may send out 100 bulk write AD result requests concurrently. But the AD result index shard setting is still the same, that may bring higher write load and user's general bulk write request may impacted. Confirm with PM/SDM if the concern of horizontal scalable issue is reasonable.

Will confirm with Sean about the horizontal scalability issue.  Result write is not problem, as we are gonna drop results when the index pressure is high.  But requests like reading checkpoints are impacted when
1)traffic is random (like ip address)
2)there are more entities than available memory can host

kaituo:
> > Some high level question which not covered in previous PR review.
> > Current design will dispatch entities to different work nodes with consistent hashing. Each work node will cache entities in queue and run RCF model, then bulk write anomaly result to index. If current detection interval timeout, we just drop all remaining entities in queue to avoid carry over to next interval. It's possible that some worker node may receive more entities which needs to run longer like need to load checkpoints. So the unbalanced load may cause some worker node drops more entities. But seems not a big concern as we don't guarantee all entities will be detected, there are other reasons can cause dropping entities like worker node queue is full.
> > Have some concern about the horizontal scalability. For example, if user scale cluster from 10 data nodes to 100 data nodes, then will have 100 worker nodes running entities, that means we may send out 100 bulk write AD result requests concurrently. But the AD result index shard setting is still the same, that may bring higher write load and user's general bulk write request may impacted. Confirm with PM/SDM if the concern of horizontal scalable issue is reasonable.
> 
> Will confirm with Sean about the horizontal scalability issue. Result write is not problem, as we are gonna drop results when the index pressure is high. But requests like reading checkpoints are impacted when
> 1)traffic is random (like ip address)
> 2)there are more entities than available memory can host

Asked Sean. He said creating an issue to track https://github.com/opensearch-project/anomaly-detection/issues/125 and don't block code commit.

kaituo:
@ylwu-amzn, please also comment on the issue I created with solutions.

ylwu-amzn (approved):

ohltyler (approved):

Issue #124, opened 2021-07-08
Don't replace detector's backend role when update


ylwu-amzn:
**Is your feature request related to a problem? Please describe.**
 
In current AD, when user create detector, we will set detector's backend role as user's backend role. When another user B update detector, will replace detector's backend role with this user B's backend role.

It is possible to deny a detector's creating user access through a series of detector update requests.  For example,
```
Users and Role Mappings: 
     CZ1 - czrole1
     CZ2 - czrole1, czrole2
     CZ3 - czrole2, czrole3
```
We can remove detector creator's access following these steps:
1. Create a detector with user CZ1. Then detector's backend role will be `czrole1`
2. Update detector description with user CZ2. Then detector backend role will change to  `czrole1, czrole2`
3. Update detector description with user CZ3. Then detector backend role will change to  `czrole2, czrole3`.
 
User CZ1 no longer has permission to access the detector they created. 

**Describe the solution you'd like**
One easy solution:
Don't replace detector's backend role when update detector. Detector's backend role will be fixed after created. 
User need to delete and recreate detector if needs to change detector's backend role.

**Describe alternatives you've considered**
User needs stricter permission control on monitor and detector, check https://github.com/opensearch-project/alerting/issues/82 and https://github.com/opensearch-project/anomaly-detection/issues/112

We can let user select backend role from his/her own backend role list when create detector. Admin should be able to update detector's backend role after creation.

ylwu-amzn:
@skkosuri-amzn (alerting plugin) @eirsep (async search) @davidcui-amzn (reporting plugin) @bowenlan-amzn (ISM) @spbjss (ML), please keep an eye on this issue and let's follow the same design.

davidcui-amzn:
@zhongnansu what's the Reporting back-end's current policy on this?

ylwu-amzn:
Close this one as we have done code change in AD.

PR #126, opened 2021-07-14
don't replace detector user when update

ylwu-amzn:

### Description
In current implementation, when user update detector,  we will replace detector user as current user. In some cases, it may cause detector creator lose access of detector. Check more details in #124 
Based on discussion with @skkosuri-amzn and PM, we plan to fix this issue quickly by blocking replacing detector user when update. User can't update detector user after creation.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/124
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
@@ -102,7 +102,7 @@ public class AnomalyDetectorJobTransportAction extends HandledTransportAction<An
                 detectorId,
                 filterByEnabled,
                 listener,
-                () -> executeDetector(listener, detectorId, seqNo, primaryTerm, rawPath, requestTimeout, user),
+                (anomalyDetector) -> executeDetector(listener, detectorId, seqNo, primaryTerm, rawPath, requestTimeout, user),
                 client,
                 clusterService,
                 xContentRegistry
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
```diff
@@ -102,7 +102,7 @@ protected void doExecute(Task task, AnomalyDetectorJobRequest request, ActionLis
                 detectorId,
                 filterByEnabled,
                 listener,
-                () -> executeDetector(listener, detectorId, seqNo, primaryTerm, rawPath, requestTimeout, user),
+                (anomalyDetector) -> executeDetector(listener, detectorId, seqNo, primaryTerm, rawPath, requestTimeout, user),
```
spbjss, line 105: Curiously, it seems the anomalyDetector was not used in the executeDetector method?

ylwu-amzn: Yes, actually the `executeDetector` will get detector with detectorId. This PR is mainly to fix the security issue. Will refactor code in separate PR to reuse detector rather than query it again.

spbjss: Thanks!

spbjss (approved):
looks good to me

src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
ylwu-amzn: Create an issue to track this https://github.com/opensearch-project/anomaly-detection/issues/127

saratvemulapalli (approved):
Had a chat with @ylwu-amzn , not really happy with the path from the customer perspective. 
But I agree security comes first. 

Thanks for the changes, they look good to me!

PR #128, opened 2021-07-14
Improve preview and exception handling in AnomalyResultTransportAction

kaituo:

### Description
This PR commits changes from the following PRs:
https://github.com/kaituo/anomaly-detection-1/pull/11
https://github.com/kaituo/anomaly-detection-1/pull/12

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/129

### Check List


src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -103,7 +102,8 @@
     private final Interpolator interpolator;
     private final ClientUtil clientUtil;
     private int maxEntitiesForPreview;
-    private final Gson gson;
+    private int pageSize;
```
ylwu-amzn, line 105: Add volatile ?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -182,6 +185,11 @@ public void getLatestDataTime(AnomalyDetector detector, ActionListener<Optional<
      * @param listener listener to return back the entities
      */
     public void getHighestCountEntities(AnomalyDetector detector, long startTime, long endTime, ActionListener<List<Entity>> listener) {
+        if (detector.getCategoryField() == null || detector.getCategoryField().isEmpty()) {
```
ylwu-amzn, line 188: Use `detector.isMultientityDetector()` ?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +198,200 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
+             *  on a 5 primary shard index. On a 15 data node cluster, terms  aggregation needs 217 seconds
+             *  on a 15 primary shard index. On a 30 data node cluster, terms aggregation needs 47 seconds
+             *  on a 30 primary shard index.
+             *
+             * Here we work around the problem using composite aggregation. Composite aggregation cannot
+             * give top entities without collecting all aggregated results. Paginated results are returned
+             * in the natural order of composite keys. This is fine for Preview API. Preview API needs the
+             * top entities to make sure there is enough data for training and showing the results. We
+             * can paginate entities and filter out entities that do not have enough docs (e.g., 256 docs).
+             * As long as we have collected the desired number of entities (e.g., 5 entities), we can stop
```
ylwu-amzn, line 227: How about add max pagination limit to avoid scanning all entities if no entity has enough docs?

src/main/java/org/opensearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
```diff
@@ -120,21 +132,44 @@ void previewExecute(
...
+                Instant endTime = request.getEndTime();
+                ActionListener<PreviewAnomalyDetectorResponse> releaseListener = ActionListener.runAfter(listener, () -> lock.release());
+                if (detector != null) {
+                    String error = validateDetector(detector);
+                    if (StringUtils.isNotBlank(error)) {
+                        listener.onFailure(new OpenSearchException(error, RestStatus.BAD_REQUEST));
+                        lock.release();
+                        return;
+                    }
+                    anomalyDetectorRunner
```
ylwu-amzn, line 159: Release semaphore once preview done?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: added

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: added

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: how about adding 1 minute expiration time as preview is time-sensitive? If we can finish within time, scanning all entities are acceptable.

src/main/java/org/opensearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
kaituo: it is released at line 151.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
ylwu-amzn: No, I always use volatile before fields which shared by multiple threads for system settings.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
ylwu-amzn: I'm ok for that

src/main/java/org/opensearch/ad/feature/AbstractRetriever.java
```diff
@@ -14,15 +14,20 @@
...
+import org.opensearch.search.aggregations.AggregationBuilder;
 import org.opensearch.search.aggregations.Aggregations;
 import org.opensearch.search.aggregations.bucket.MultiBucketsAggregation;
+import org.opensearch.search.aggregations.bucket.composite.CompositeAggregationBuilder;
 import org.opensearch.search.aggregations.metrics.InternalTDigestPercentiles;
 import org.opensearch.search.aggregations.metrics.NumericMetricsAggregation.SingleValue;
 import org.opensearch.search.aggregations.metrics.Percentile;
+import org.opensearch.search.builder.SearchSourceBuilder;
 
 public class AbstractRetriever {
```
jmazanec15, line 32: Out of curiousity, why is this class called abstract but not actually abstract?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +198,200 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
+             *   }
+             *
+             */
+            bucketAggs = AggregationBuilders
+                .composite(
+                    AGG_NAME_TOP,
+                    detector.getCategoryField().stream().map(f -> new TermsValuesSourceBuilder(f).field(f)).collect(Collectors.toList())
+                )
+                .size(pageSize);
+            bucketAggs
```
jmazanec15, line 303: nit: why not remove semi colon after line 302 and just call subAggregation?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +198,200 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
+            this.detector = detector;
+            this.topEntities = new ArrayList<>();
+            this.searchSourceBuilder = searchSourceBuilder;
+        }
+
+        @Override
+        public void onResponse(SearchResponse response) {
+            try {
+                Aggregations aggs = response.getAggregations();
+                if (aggs == null) {
```
jmazanec15, line 337: Is this okay to be null?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -323,68 +323,78 @@ public void onResponse(CompositeRetriever.Page entityFeatures) {
...
+                        LOG.error("Unexpetected exception", e);
+                        handleException(e);
+                    }
                 });
             }
         }
 
         @Override
         public void onFailure(Exception e) {
+            LOG.error("Unexpetected exception", e);
```
jmazanec15, line 393: nit: Unexpected not Unexpetected

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: please check the upcoming commit

README.md
```diff
@@ -1,6 +1,6 @@
 [![AD Test](https://github.com/opensearch-project/anomaly-detection/workflows/Build%20and%20Test%20Anomaly%20detection/badge.svg)](https://github.com/opensearch-project/anomaly-detection/actions?query=workflow%3A%22Build+and+Test+Anomaly+detection%22+branch%3A%22main%22)
 [![codecov](https://codecov.io/gh/opensearch-project/anomaly-detection/branch/main/graph/badge.svg?flag=plugin)](https://codecov.io/gh/opensearch-project/anomaly-detection)
-[![Documentation](https://img.shields.io/badge/doc-reference-blue)](https://docs-beta.opensearch.org/monitoring-plugins/ad/index/)
```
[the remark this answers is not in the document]
kaituo: this is to solve CI errors. Didn't plan to include here.  Next time I will write the diff in the overview and create a new PR.

src/main/java/org/opensearch/ad/feature/AbstractRetriever.java
kaituo: good question, changed to make it abstract.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: good catch.  removed the semicolon.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: It is not ok.  So I returned immediately without proceeding forward.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: yes, fixed

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
jmazanec15: Right, but then why isnt onFailure called?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +233,211 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
-                .flatMap(agg -> ((Terms) agg).getBuckets().stream())
-                .map(bucket -> bucket.getKeyAsString())
-                .collect(Collectors.toList())
-                .stream()
-                .map(entityValue -> parseCategoricalField(entityValue, detector))
-                .collect(Collectors.toList());
-            listener.onResponse(results);
-        }, listener::onFailure);
-        client.search(searchRequest, termsListener);
+        client
```
jmazanec15, line 351: For another PR (i.e. non-blocker): I realize this is not being changed in this PR, but why is the node client used here and not a separate threadpool? Isn't this call expensive and it will be run frequently?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +233,211 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
+                    for (int i = 0; i < amountToWrite && i < pageResults.size(); i++) {
+                        topEntities.add(pageResults.get(i));
+                    }
+                    Map<String, Object> afterKey = compositeAgg.afterKey();
+                    if (topEntities.size() >= maxEntitiesForPreview || afterKey == null || expirationEpochMs < clock.millis()) {
+                        listener.onResponse(topEntities);
+                    } else {
+                        updateSourceAfterKey(afterKey, searchSourceBuilder);
+                        client
+                            .search(
```
jmazanec15, line 425: For my understanding, why is another search call being made here conditionally, but not in line 405?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +233,211 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
+                    }
+                }
+            } catch (Exception e) {
+                onFailure(e);
+            }
+        }
+
+        @Override
+        public void onFailure(Exception e) {
+            logger.error("Fail to pageinate", e);
```
jmazanec15, line 438: Misspelling: pageinate --> paginate

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: This would indicate some bug or some opensearch core changes that we are not aware of (we don't keep up-to-date with the large amounts of changes there). For example, they may change to if there are results return it; otherwise return null instead of an empty Aggregations as they currently do.  I am ok with fail fast or recover from the failure, log it, and return whatever we have.  No strong preference.  What about you?

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: Sending a search request will land in Opensearch's search threadpool no matter what.  We cannot run it in other threadpool.  If preparing the search request is expensive, we can wrap that logic in the AD threadpool.  Here it is not the case.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: because at line 405, we are using terms aggregation and one call is enough.  Here we are using composite queries that are not a one-shot deal, like scrolling.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
kaituo: good catch.  Fixed.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -190,67 +233,211 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
...
+                        .filter(bucket -> bucket.getDocCount() >= minimumDocCountForPreview)
+                        .map(bucket -> Entity.createEntityByReordering(detector.getDetectorId(), bucket.getKey()))
+                        .collect(Collectors.toList());
+                    // we only need at most maxEntitiesForPreview
+                    int amountToWrite = maxEntitiesForPreview - topEntities.size();
+                    for (int i = 0; i < amountToWrite && i < pageResults.size(); i++) {
+                        topEntities.add(pageResults.get(i));
+                    }
+                    Map<String, Object> afterKey = compositeAgg.afterKey();
+                    if (topEntities.size() >= maxEntitiesForPreview || afterKey == null || expirationEpochMs < clock.millis()) {
```
ylwu-amzn, line 420: If timeout, should we use `onFailure` to show timeout warning on frontend ?

kaituo: how about if timeout and if there are entities found, log the timeout and send whatever we have? And if timeout and if there are no entities found, send failure as you said?

jmazanec15 (approved):

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
ylwu-amzn: That makes sense. We should only send out timeout failure if no entities found.

kaituo: changed. Please check the commit.

ylwu-amzn (approved):

jmazanec15 (approved):

PR #48, opened 2021-05-18
add realtime task

ylwu-amzn:
Signed-off-by: Yaliang Wu <ylwu@amazon.com>

### Description
Add realtime task for detector realtime job.
1.Create realtime task when start realtime detection.
2.Set realtime task state as stopped when realtime detection job stopped.
 
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #130, opened 2021-07-15
add init progress of realtime task

ylwu-amzn:
### Description
Add init progress of realtime task

 
### Check List


src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1500,6 +1501,8 @@ public void updateLatestADTask(
         getAndExecuteOnLatestDetectorLevelTask(detectorId, taskTypes, (adTask) -> {
             if (adTask.isPresent()) {
                 updateADTask(adTask.get().getTaskId(), updatedFields);
+            } else {
```
kaituo, line 1504: some generics issue for this method:
1) the parameter listener should have generic type.
2) getAndExecuteOnLatestDetectorLevelTask is a generic method.  Please add the type.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1538,6 +1541,41 @@ public void updateLatestRealtimeTask(
...
 
+    /**
+     * Update latest realtime task's init progress, estimated minutes left for initialization, state and error field.
+     *
+     * @param detectorId detector id
+     * @param totalUpdates total updates
+     * @param intervalMinutes interval minutes
+     * @param failure failure
+     */
+    public void updateLatestRealtimeTask(String detectorId, long totalUpdates, long intervalMinutes, Exception failure) {
```
kaituo, line 1552: Do we save no matter what?  If yes, can we save as needed, similar to what we do in DetectionStateHandler.saveError?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -736,6 +754,12 @@ private void handleRCFResults(int numFeatures) {
                     listener.onFailure(new InternalFailure(adID, NO_MODEL_ERR_MSG));
                     return;
                 }
+                try {
+                    long totalUpdates = rcfResults.get(0).getTotalUpdates();
+                    adTaskManager.updateLatestRealtimeTask(detector.getDetectorId(), totalUpdates, intervalMinutes, null);
```
kaituo, line 759: This can clear the detector's error, right? Current implementation would set error after observing rcf and threshold model and cold start error and create a final error in job scheduler runner.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -1145,11 +1168,16 @@ public void onFailure(Exception e) {
...
+            } else if (entityResultResponses.isEmpty()) {
                 listener.onFailure(new InternalFailure(adID, NO_ACK_ERR));
             } else {
                 listener.onResponse(new AnomalyResultResponse(0, 0, 0, new ArrayList<FeatureData>()));
             }
+            long maxTotalUpdates = 0;
+            while (!entityResultResponses.isEmpty()) {
+                maxTotalUpdates = Math.max(maxTotalUpdates, entityResultResponses.poll().getTotalUpdates());
+            }
+            adTaskManager.updateLatestRealtimeTask(adID, maxTotalUpdates, intervalMinutes, failure.get());
```
kaituo, line 1180: similar to the other comments, should we leave the setting error logic to job scheduler? This error can be overwritten by other errors (e.g., feature aggregation and other pages's entityResultResponse).

kaituo:
Current DetectionStateHandler.update still uses detector id to find a state.  We need to update it or move logic to ADTaskManager.  Considering ADTaskManager is already large, we can also move the real time part to DetectionStateHandler.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: Not sure why, seems some code change not synced from my Mac to desktop due to VPN issue. Now you should be able to see the latest change.

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -81,6 +81,9 @@
     // Use this field to cache all HC tasks. Key is detector id
     private Map<String, ADHCBatchTaskCache> hcTaskCaches;
 
+    // This field is to cache all realtime tasks. Key is detector id
+    private Map<String, ADRealtimeTaskCache> realtimeTaskCaches;
```
ylwu-amzn, line 85: https://github.com/opensearch-project/anomaly-detection/pull/48#discussion_r636310719

@spbjss hi, Alex, this change answers your concern about why we don't name the cache manager as `HistoricalTaskCacheManager`

src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
```diff
@@ -123,6 +127,9 @@ protected DeleteModelNodeResponse nodeOperation(DeleteModelNodeRequest request)
 
         cache.get().clear(adID);
 
+        // delete realtime task cache
+        adTaskCacheManager.removeRealtimeTaskCache(adID);
```
ylwu-amzn, line 131: TODO: cleanup stale realtime task cache in hourly cron. 

Will add this in another PR.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Yeah, agree. The first commit is some testing code to verify if can get total updates and update realtime init progress successfully. The second commit should fix it.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Same as the comment above

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1538,6 +1544,57 @@ public void updateLatestRealtimeTask(
...
+        String newState,
+        Long rcfTotalUpdates,
+        Long detectorIntervalInMinutes,
+        String error
+    ) {
+        Float initProgress = null;
+        String state = null;
+        if (detectorIntervalInMinutes != null && rcfTotalUpdates != null) {
+            state = ADTaskState.INIT.name();
+            if (rcfTotalUpdates <= NUM_MIN_SAMPLES) {
```
kaituo, line 1566: when rcfTotalUpdates == NUM_MIN_SAMPLES, the state is RUNNING.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1538,6 +1544,57 @@ public void updateLatestRealtimeTask(
...
+                initProgress = (float) rcfTotalUpdates / NUM_MIN_SAMPLES;
+            } else {
+                state = ADTaskState.RUNNING.name();
+                initProgress = 1.0f;
+            }
+        }
+        if (newState != null) {
+            state = newState;
```
kaituo, line 1574: What if rcfTotalUpdates < NUM_MIN_SAMPLES and newState is RUNNING?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -1145,10 +1194,14 @@ public void onFailure(Exception e) {
...
-            } else if (ackResponses.isEmpty()) {
+            } else if (entityResultResponses.isEmpty()) {
                 listener.onFailure(new InternalFailure(adID, NO_ACK_ERR));
             } else {
-                listener.onResponse(new AnomalyResultResponse(0, 0, 0, new ArrayList<FeatureData>()));
+                long maxTotalUpdates = 0;
+                while (!entityResultResponses.isEmpty()) {
+                    maxTotalUpdates = Math.max(maxTotalUpdates, entityResultResponses.poll().getTotalUpdates());
+                }
+                listener.onResponse(new AnomalyResultResponse(0, 0, 0, new ArrayList<FeatureData>(), maxTotalUpdates, intervalMinutes));
```
kaituo, line 1204: This may require a rewrite as current main branch has different logic.

kaituo:
Real time state doc is changed to not using detector id as doc id.  After upgrading, we will have two docs for real time task: one old doc using detector id and one new doc that does not.  Could you point me in code how you handle this case?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: updated in new commit

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: The new state has higher priority. It will always override state calculated with RCF total updates. Will add more comments.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Thanks for reminding. Will resolve in separate PR.

ylwu-amzn:
> Real time state doc is changed to not using detector id as doc id. After upgrading, we will have two docs for real time task: one old doc using detector id and one new doc that does not. Could you point me in code how you handle this case?

Create an issue to track backward compatibility issue. https://github.com/opensearch-project/anomaly-detection/issues/133

kaituo (approved):

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -767,4 +771,54 @@ public void clearPendingEntities(String detectorId) {
...
+     * 2. If any field value changed, will consider the realtime task changed.
+     * 3. If realtime task cache not found, will consider the realtime task changed
+     *    and put new realtime task into cache.
+     * @param detectorId detector id
+     * @param newState new task state
+     * @param newInitProgress new init progress
+     * @param newError new error
+     * @return true if realtime task changed comparing with realtime task cache.
+     */
+    public boolean realtimeTaskChanged(String detectorId, String newState, Float newInitProgress, String newError) {
```
ohltyler, line 787: Seems to me that the name of this method would imply that it is just returning a boolean on whether the RT task has changed or not, and not also potentially updating the existing cache. Thoughts on changing the name to something more clear?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1538,6 +1544,59 @@ public void updateLatestRealtimeTask(
...
+        }
+
+        error = Optional.ofNullable(error).orElse("");
+        if (!adTaskCacheManager.realtimeTaskChanged(detectorId, state, initProgress, error)) {
+            return;
+        }
+        if (ADTaskState.STOPPED.name().equals(state)) {
+            adTaskCacheManager.removeRealtimeTaskCache(detectorId);
```
ohltyler, line 1584: I only see a detector's removed from the cache if it is stopped or deleted. What about the failed state?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1538,6 +1544,59 @@ public void updateLatestRealtimeTask(
...
+        if (initProgress != null) {
+            updatedFields.put(INIT_PROGRESS_FIELD, initProgress);
+            updatedFields.put(ESTIMATED_MINUTES_LEFT_FIELD, Math.max(0, NUM_MIN_SAMPLES - rcfTotalUpdates) * detectorIntervalInMinutes);
+        }
+        if (state != null) {
+            updatedFields.put(STATE_FIELD, state);
+        }
+        if (error != null) {
+            updatedFields.put(ERROR_FIELD, error);
+        }
```
ohltyler, line 1596: What about the case from going to a non-null `error` field to a null `error` field (e.g., user updates a buggy detector and restarts)? Wouldn't we always want to clear the persisted `error` field here?

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Sure, how about `checkIfRealtimeTaskChangedAndUpdateCache`?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: If one interval failed, the realtime job will not stop, so we will keep using current task

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: Line 1579 will reset error as empty if no error happens in current interval

`error = Optional.ofNullable(error).orElse("");`

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ohltyler: I'm ok with that, sure

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ohltyler: Got it. So any prolonged failure that leads to an EndRunException will lead to stopped state, which will remove the task from the cache. Makes sense

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ohltyler: I see, makes sense.

ohltyler (approved):

PR #131, opened 2021-07-15
add delete anomaly results API

ylwu-amzn:
### Description
With historical analysis, user can generate much more anomaly result in a short time, especially for HC detector. AD plugin will roll over anomaly result index if its doc count reaches some limit, and will delete roll-over index after some period (30 days by default). So if user run historical analysis, it may generate too many AD results and speed up the rollover and AD result index cleanup. 

This PR adds delete anomaly result API. User can use this API to delete anomaly results with specific query like delete all anomaly results of one detector/task.

Sample request

```
DELETE _plugins/_anomaly_detection/detectors/results
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "detector_id": {
              "value": "6tZminoBaXxjTubPk0k6"
            }
          }
        }
      ]
    }
  }
}
```
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/DeleteAnomalyResultsTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/DeleteAnomalyResultsTransportAction.java
@@ -1,0 +1,86 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.transport;
+
+import static org.opensearch.ad.settings.AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES;
+import static org.opensearch.ad.util.ParseUtils.addUserBackendRolesFilter;
+import static org.opensearch.ad.util.ParseUtils.getUserContext;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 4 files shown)


zhanghg08 (commented):
A high level comment, do we need to consider FGAC with delete api?

ylwu-amzn:
> A high level comment, do we need to consider FGAC with delete api?

Yes, check DeleteAnomalyResultsTransportAction.java line 79 
```
addUserBackendRolesFilter(user, request.getSearchRequest().source());
```

zhanghg08 (approved):
LGTM!

src/main/java/org/opensearch/ad/transport/DeleteAnomalyResultsTransportAction.java
```diff
@@ -0,0 +1,86 @@
...
+        if (user == null || !filterEnabled) {
+            // Case 1: user == null when 1. Security is disabled. 2. When user is super-admin
+            // Case 2: If Security is enabled and filter is disabled, proceed with search as
+            // user is already authenticated to hit this API.
+            client.execute(DeleteByQueryAction.INSTANCE, request, listener);
+        } else {
+            // Security is enabled and backend role filter is enabled
+            try {
+                addUserBackendRolesFilter(user, request.getSearchRequest().source());
+                client.execute(DeleteByQueryAction.INSTANCE, request, listener);
```
spbjss, line 80: client.execute() will be called anyway?

ylwu-amzn: Yes, split the code branch clearly to make it more readable.

spbjss (approved):
looks good to me.

Issue #97, opened 2021-06-09
[Meta] Anomaly Detection: upgrade to use RCF 2.0


sean-zheng-amazon:
**Is your feature request related to a problem? Please describe.**
RCF 2.0 provides significant system performance improvement over RCF 1.0. We want to upgrade AD to use RCF 2.0 to benefit from the performance gain.

**Describe the solution you'd like**
The biggest challenge here is the backward compatibility. API wise we can hide the change, but we have model snapshots persisted, and the serialization format has been changed. So we need to ensure the plugin can read both old and the new formats. Once read in, the new writes will all in new formats. This way, we can provide seamless upgrade without service down time.

**Describe alternatives you've considered**
An alternative is to provide a tool to convert old snapshots to new format as a batch operation. But UX wise, this is not ideal, since users need to explicitly run the tool to upgrade.

PR #137, opened 2021-07-26
Change to use protostuff for serializing/deserializing RCF models

kaituo:

### Description

1)Change to use protostuff instead of gson for deserialization/serialization as protostuff is more efficient.
2)Use object pool for protostuff deserialization. Otherwise, we have to allocate, and GC has to destroy a buffer for every RCF deserialization.
3)Deal with backward compatibility when a new node using compact rcf reading old gson checkpoints.
4)Move the logic of deserializing checkpoints from ModelManager to CheckpointDao as I can hide the complexity of backward compatibility in CheckpointDao without exposing it in two places.

Testing done:
1. Did end-to-end testing to make sure backward compatibility works.
2. Fixed/added related unit tests.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
@@ -27,6 +27,7 @@
 package org.opensearch.ad.ml;
 
 import java.io.IOException;
+import java.io.StringReader;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.time.Instant;
@@ -35,12 +36,14 @@ import java.time.ZonedDateTime;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 
+import org.apache.commons.pool2.impl.GenericObjectPool;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.ParameterizedMessage;
@@ -61,12 +64,16 @@ import org.opensearch.action.get.MultiGetResponse;
 import org.opensearch.action.index.IndexRequest;
 import org.opensearch.action.index.IndexResponse;
 import org.opensearch.action.support.IndicesOptions;
+import org.opensearch.action.update.UpdateRequest;
+import org.opensearch.action.update.UpdateResponse;
+import org.opensearch.ad.common.exception.ResourceNotFoundException;
 import org.opensearch.ad.constant.CommonName;
 import org.opensearch.ad.indices.ADIndex;
 import org.opensearch.ad.indices.AnomalyDetectionIndices;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.util.ClientUtil;
 import org.opensearch.client.Client;
+import org.opensearch.common.Strings;
 import org.opensearch.index.IndexNotFoundException;
 import org.opensearch.index.query.MatchQueryBuilder;
 import org.opensearch.index.reindex.BulkByScrollResponse;
@@ -75,10 +82,19 @@ import org.opensearch.index.reindex.DeleteByQueryRequest;
 import org.opensearch.index.reindex.ScrollableHitSource;
 
 import com.amazon.randomcutforest.RandomCutForest;
-import com.amazon.randomcutforest.serialize.RandomCutForestSerDe;
+import com.amazon.randomcutforest.config.Precision;
+import com.amazon.randomcutforest.serialize.json.v1.V1JsonToV2StateConverter;
+import com.amazon.randomcutforest.state.RandomCutForestMapper;
+import com.amazon.randomcutforest.state.RandomCutForestState;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+
+import io.protostuff.LinkedBuffer;
+import io.protostuff.ProtostuffIOUtil;
+import io.protostuff.Schema;
 
 /**
  * DAO for model checkpoints.
@@ -100,6 +116,7 @@ public class CheckpointDao {
     public static final String ENTITY_RCF = "rcf";
     public static final String ENTITY_THRESHOLD = "th";
     public static final String FIELD_MODEL = "model";
+    public static final String FIELD_MODELV2 = "modelV2";
     public static final String TIMESTAMP = "timestamp";
     public static final String DETECTOR_ID = "detectorId";
 
@@ -111,7 +128,9 @@ public class CheckpointDao {
     private final String indexName;
 
     private Gson gson;
-    private RandomCutForestSerDe rcfSerde;
+    private RandomCutForestMapper mapper;
+    private Schema<RandomCutForestState> schema;
+    private V1JsonToV2StateConverter converter;
 
     private final Class<? extends ThresholdingModel> thresholdingModelClass;
 
@@ -120,6 +139,10 @@ public class CheckpointDao {
     // we won't read/write a checkpoint larger than a threshold
     private final int maxCheckpointBytes;
 
+    private final TypeAdapter<JsonObject> strictGsonObjectAdapter;
+    private final GenericObjectPool<LinkedBuffer> deserializeRCFBufferPool;
+    private final int deserializeRCFBufferSize;
+
     /**
      * Constructor with dependencies and configuration.
      *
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -100,6 +116,7 @@
     public static final String ENTITY_RCF = "rcf";
     public static final String ENTITY_THRESHOLD = "th";
     public static final String FIELD_MODEL = "model";
+    public static final String FIELD_MODELV2 = "modelV2";
```
sudiptoguha, line 119: Do you want to make it a combination of the version string in RCF 2.0? So that if it becomes 2.1-rcX instead of 2.0-rc2 then you would not have to rename anything. That being said, the renaming can also be done then.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -298,16 +332,39 @@ public String toCheckpoint(EntityModel model) {
...
+            logger.error("Failed to borrow an buffer from object pool", e);
+            return rcfModelToCheckpoint(model, null);
+        }
+        return null;
+    }
+
+    String rcfModelToCheckpoint(RandomCutForest model, LinkedBuffer buffer) {
+        final LinkedBuffer deserializationBuffer = buffer == null ? LinkedBuffer.allocate(deserializeRCFBufferSize) : buffer;
+        try {
+            RandomCutForestState state = mapper.toState(model);
```
sudiptoguha, line 361: Did you want to set mapper.setSaveExecutorContextEnabled(true);

What about mapper.setSaveTreeStateEnabled(true)? Saving the trees would speed up run time dramatically.

sudiptoguha: Maybe the initialization controls these?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -425,6 +492,52 @@ private void logFailure(BulkByScrollResponse response, String detectorID) {
...
 
+    private RandomCutForest restoreRCFModel(String rcfCheckpoint) {
+        if (Strings.isEmpty(rcfCheckpoint)) {
+            return null;
+        }
+
+        String checkpoint = rcfCheckpoint.trim();
+        RandomCutForest forest = null;
+        try {
+            if (checkpoint.charAt(0) == '{' && checkpoint.charAt(checkpoint.length() - 1) == '}' && isJson(checkpoint)) {
```
sudiptoguha, line 503: Make this a "isVersionEqualsOne?" function? Again not urgent - but we may have more use for that function later.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -455,26 +568,70 @@ public void restoreModelCheckpoint(String modelId, ActionListener<Optional<Entry
...
+        }, listener::onFailure));
+    }
+
+    /**
+     * Returns to listener the checkpoint for the threshold model.
+     *
+     * @param modelId id of the model
+     * @param listener onResponse is called with the model checkpoint, or empty for no such model
+     */
+    public void getThresholdModel(String modelId, ActionListener<Optional<ThresholdingModel>> listener) {
```
sudiptoguha, line 600: Question: If the threshold model was also updated to V2 (say?) would this suffice? Any code we want to leave here such that V2 of Threshold model adheres to that?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: This is for checkpoint format.  As long as rcf checkpoint format does not change, we don't have to touch this part, even if we move to rcX.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: Yes, mapper is a singleton and I initiated here: https://github.com/kaituo/anomaly-detection-1/blob/compactRCF2/src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java#L361

sudiptoguha: Ok. Do you want to save the trees? Saving the trees would be 10x faster (but some more space). Maybe have a config option so that we can decide on the future/case-by-case?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: yes, changed

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: Great question.  Currently, for single stream detectors, theshold model uses model (old) field in checkpoint; for HCAD detectors, threshold models use modelV2 (new)  (as for HCAD rcf, threshold, samples are in the same checkpoint.  Since rcf changed to modelV2, threshold uses modelV2 as well.).  When there is a new threshold model, there can be a few ways to migrate: for example, if the new threshold model only stores 5 numbers, we can store them in 5 different fields and fallback to model/modelV2 field if these 5 fields are not found.  Or we introduce a modelV3 field for the new model format.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: Yes, let me add "mapper.setSaveTreeStateEnabled(true)".  That's the only change I need to make, right?

sudiptoguha: yes.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -298,16 +332,39 @@ public String toCheckpoint(EntityModel model) {
...
+                    deserializeRCFBufferPool.invalidateObject(borrowedBuffer);
+                    borrowedBuffer = null;
+                }
+            } finally {
+                if (borrowedBuffer != null) {
+                    deserializeRCFBufferPool.returnObject(borrowedBuffer);
+                }
+            }
+        } catch (Exception e) {
+            logger.error("Failed to borrow an buffer from object pool", e);
```
jmazanec15, line 352: Is this an error if we continue to execute? Are there any valid cases where borrowedBuffer should be expected to be null? In line 359 we allocate a new LinkedBuffer.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -298,16 +332,39 @@ public String toCheckpoint(EntityModel model) {
...
+            return rcfModelToCheckpoint(model, null);
+        }
+        return null;
+    }
+
+    String rcfModelToCheckpoint(RandomCutForest model, LinkedBuffer buffer) {
+        final LinkedBuffer deserializationBuffer = buffer == null ? LinkedBuffer.allocate(deserializeRCFBufferSize) : buffer;
```
jmazanec15, line 359: Why are we calling it deserializationBuffer if we are serializing it in this function?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -425,6 +492,52 @@ private void logFailure(BulkByScrollResponse response, String detectorID) {
         }
     }
 
+    private RandomCutForest restoreRCFModel(String rcfCheckpoint) {
```
jmazanec15, line 495: Why is called restoreRCF instead of deserialize?

src/main/resources/mappings/checkpoint.json
```diff
@@ -32,6 +32,9 @@
           "type": "keyword"
         }
       }
+    },
+    "modelV2": {
+      "type": "text"
```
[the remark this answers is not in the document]
kaituo: Previous code works on text :) If changing to binary, need to change more things.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -298,16 +332,39 @@ public String toCheckpoint(EntityModel model) {
...
+            try {
+                return rcfModelToCheckpoint(model, borrowedBuffer);
+            } catch (Exception e) {
+                if (borrowedBuffer != null) {
+                    deserializeRCFBufferPool.invalidateObject(borrowedBuffer);
+                    borrowedBuffer = null;
+                }
+            } finally {
+                if (borrowedBuffer != null) {
+                    deserializeRCFBufferPool.returnObject(borrowedBuffer);
```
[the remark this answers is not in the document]
kaituo: after invalidate, I will set borrowedBuffer to null.  So the return logic won't be called.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: What I wrote is if we cannot borrow any buffer from object pool then go to allocate a new LinkedBuffer in line 359.

We return null when we successfully borrow an object but fail serialization.  I moved return null to the exception block to be more explicit and added more comments.

```
private String rcfModelToCheckpoint(RandomCutForest model) {
        LinkedBuffer borrowedBuffer = null;
        try {
            borrowedBuffer = deserializeRCFBufferPool.borrowObject();
            try {
                return rcfModelToCheckpoint(model, borrowedBuffer);
            } catch (Exception e) {
                if (borrowedBuffer != null) {
                    deserializeRCFBufferPool.invalidateObject(borrowedBuffer);
                    borrowedBuffer = null;
                }
                // return null when we successfully borrow an object but fail serialization
                return null;
            } finally {
                if (borrowedBuffer != null) {
                    deserializeRCFBufferPool.returnObject(borrowedBuffer);
                }
            }
        } catch (Exception e) {
            logger.error("Failed to borrow an buffer from object pool", e);
            return rcfModelToCheckpoint(model, null);
        }
    }
```

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: good catch. Will change to serializationBuffer

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: changed to deserializeRCFModel.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
sudiptoguha: Also consider adding "mapper.setPartialTreeStateEnabled(true);" this will save space by 20% but serialization/de time will increase by 20%. I think we are more space constrained for HCAD at the moment (If not true, then do not add this).

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -470,51 +437,34 @@ public void stopModel(String detectorId, String modelId) {
...
+                        ActionListener.wrap(r -> listener.onResponse(null), listener::onFailure)
+                    );
+            } else if (model instanceof ThresholdingModel) {
+                checkpointDao
+                    .putThresholdCheckpoint(
+                        modelId,
+                        (ThresholdingModel) model,
+                        ActionListener.wrap(r -> listener.onResponse(null), listener::onFailure)
+                    );
+            }
```
jmazanec15, line 464: If we use "else if" in line 457, should we throw IllegalStateException after line 464? Alternatively, 457 could just be else.

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -822,15 +649,23 @@ public void maintenance(ActionListener<Void> listener) {
...
-                }));
+                    maintenanceForIterator(models, iter, listener);
+                });
+                T model = modelState.getModel();
+                if (model instanceof RandomCutForest) {
+                    checkpointDao.putRCFCheckpoint(modelId, (RandomCutForest) model, checkpointListener);
+                } else if (model instanceof ThresholdingModel) {
+                    checkpointDao.putThresholdCheckpoint(modelId, (ThresholdingModel) model, checkpointListener);
+                } else {
+                    maintenanceForIterator(models, iter, listener);
```
jmazanec15, line 665: This should never be reached correct?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -120,79 +139,96 @@
...
         this.indexName = indexName;
         this.gson = gson;
-        this.rcfSerde = rcfSerde;
+        this.mapper = mapper;
+        this.schema = schema;
+        this.converter = converter;
         this.thresholdingModelClass = thresholdingModelClass;
         this.indexUtil = indexUtil;
         this.maxCheckpointBytes = maxCheckpointBytes;
+        this.converter = converter;
```
jmazanec15, line 186: Duplicate in line 182

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -120,79 +139,96 @@
...
+     * Puts a thresholding model checkpoint in the storage.
      *
      * @param modelId id of the model
-     * @param modelCheckpoint checkpoint of the model
+     * @param threshold the thresholding model
      * @param listener onResponse is called with null when the operation is completed
      */
-    public void putModelCheckpoint(String modelId, String modelCheckpoint, ActionListener<Void> listener) {
+    public void putThresholdCheckpoint(String modelId, ThresholdingModel threshold, ActionListener<Void> listener) {
+        String modelCheckpoint = AccessController.doPrivileged((PrivilegedAction<String>) () -> gson.toJson(threshold));
```
jmazanec15, line 227: Why does threshold mode put to FIELD_MODEL instead of FIELD_MODELV2?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -425,6 +492,52 @@ private void logFailure(BulkByScrollResponse response, String detectorID) {
...
+        return forest;
+    }
+
+    /**
+     * @param testString test string
+     * @return if the given string is json or not
+     */
+    private boolean isJson(String testString) {
+        try {
+            try (JsonReader reader = new JsonReader(new StringReader(testString))) {
```
jmazanec15, line 531: Why do we need nested try here?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: will change.  Disk will be become a bottleneck for 1 million entity.

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: I'll throw IllegalStateException after line 464

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: yeah, sent IllegalArgumentException instead.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: good catch.  Removed.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: Because I don't want to deal with backward compatibility when it is not needed.  Currently, threshold model is in FIELD_MODEL.

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: good catch.  removed the outmost try.

jmazanec15 (approved):

sudiptoguha (approved):

jmazanec15 (approved):

PR #138, opened 2021-07-26
Add new formula for estimating rcf model size

kaituo:

### Description
This PR created the formula for estimating compact rcf model size.

Testing done:
1. Compared the actual size and the estimated size. The numbers mostly match.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/ModelPartitioner.java
+++ b/src/main/java/org/opensearch/ad/ml/ModelPartitioner.java
@@ -36,9 +36,11 @@ import org.opensearch.ad.MemoryTracker;
 import org.opensearch.ad.common.exception.LimitExceededException;
 import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.util.DiscoveryNodeFilterer;
 
 import com.amazon.randomcutforest.RandomCutForest;
+import com.amazon.randomcutforest.config.Precision;
 
 /**
  * This class breaks the circular dependency between NodeStateManager and ModelManager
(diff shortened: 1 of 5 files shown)


sudiptoguha (approved):
Looks good to me. We should probably have an issue that the formula incorporates samplesize as a variable (now 256). It's ok at the moment - we cannot/should not make that change now. But larger samplesize forests be useful down the road and we would need to change this. But the mechanism of how to change is already here.

kaituo:
> Looks good to me. We should probably have an issue that the formula incorporates samplesize as a variable (now 256). It's ok at the moment - we cannot/should not make that change now. But larger samplesize forests be useful down the road and we would need to change this. But the mechanism of how to change is already here.

Yes, created an issue: https://github.com/opensearch-project/anomaly-detection/issues/146

src/main/java/org/opensearch/ad/ml/ModelPartitioner.java
```diff
@@ -95,6 +97,11 @@ public ModelPartitioner(int rcfNumSamplesInTree, int rcfNumTrees, DiscoveryNodeF
                 .numberOfTrees(rcfNumTrees)
                 .outputAfter(rcfNumSamplesInTree)
                 .parallelExecutionEnabled(false)
+                .compact(true)
+                .precision(Precision.FLOAT_32)
+                .boundingBoxCacheFraction(AnomalyDetectorSettings.REAL_TIME_BOUNDING_BOX_CACHE_RATIO)
+                // same with dimension for opportunistic memory saving
+                .shingleSize(rcfNumFeatures)
```
ylwu-amzn, line 104: Same issue with https://github.com/opensearch-project/anomaly-detection/pull/143#discussion_r677812252

kaituo: changed

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -50,19 +53,25 @@
...
+import org.opensearch.ad.model.Entity;
+import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.set.Sets;
 
 import com.amazon.randomcutforest.RandomCutForest;
+import com.google.common.collect.ImmutableList;
 
 public class ADTaskCacheManager {
```
kaituo, line 67: @sudiptoguha Ignore most of the diff in this file as they are from rebase.  It's Yaliang's code.

sudiptoguha (approved):

ylwu-amzn (approved):

PR #139, opened 2021-07-26
Deal with incompatible checkpoint exception

kaituo:

### Description

1)During a rolling upgrade or blue/green deployment, ResourceNotFoundException might be caused by an old node using RCF 1.0 that cannot recognize a new checkpoint produced by the coordinating node using compact RCF. We mute a node for a detector for 15 minutes if that node keeps returning ResourceNotFoundException (5 times in a row).
2)Previously, the node was muted on a global level. However, due to changes in 1), we have to change it to act on the detector level as not every detector has issues on the same node.
3)Fixed a bug in AnomalyResultTransportAction. In single-stream AD, we didn't filter out muted nodes when sending requests. This PR fixed that.

Testing done:
1. Manually verified the bug is fixed.
2. Fixed/added related unit tests.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/NodeStateManager.java
+++ b/src/main/java/org/opensearch/ad/NodeStateManager.java
@@ -26,10 +26,14 @@
 
 package org.opensearch.ad;
 
+import static org.opensearch.ad.settings.AnomalyDetectorSettings.BACKOFF_MINUTES;
+import static org.opensearch.ad.settings.AnomalyDetectorSettings.MAX_RETRY_FOR_UNRESPONSIVE_NODE;
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
 import java.time.Clock;
 import java.time.Duration;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -51,8 +55,10 @@ import org.opensearch.ad.transport.BackPressureRouting;
 import org.opensearch.ad.util.ClientUtil;
 import org.opensearch.ad.util.ExceptionUtil;
 import org.opensearch.client.Client;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.lease.Releasable;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
 import org.opensearch.common.xcontent.XContentParser;
@@ -65,18 +71,18 @@ import org.opensearch.common.xcontent.XContentType;
  */
 public class NodeStateManager implements MaintenanceState, CleanState {
     private static final Logger LOG = LogManager.getLogger(NodeStateManager.class);
+    public static final String NO_ERROR = "no_error";
     private ConcurrentHashMap<String, NodeState> states;
     private Client client;
     private ModelPartitioner modelPartitioner;
     private NamedXContentRegistry xContentRegistry;
     private ClientUtil clientUtil;
-    // map from ES node id to the node's backpressureMuter
-    private Map<String, BackPressureRouting> backpressureMuter;
+    // map from detector id to the map of ES node id to the node's backpressureMuter
+    private Map<String, Map<String, BackPressureRouting>> backpressureMuter;
     private final Clock clock;
-    private final Settings settings;
     private final Duration stateTtl;
-
-    public static final String NO_ERROR = "no_error";
+    private int maxRetryForUnresponsiveNode;
+    private TimeValue mutePeriod;
 
     /**
      * Constructor
@@ -88,7 +94,7 @@ public class NodeStateManager implements MaintenanceState, CleanState {
      * @param clock A UTC clock
      * @param stateTtl Max time to keep state in memory
      * @param modelPartitioner Used to partiton a RCF forest
-    
+     * @param clusterService Cluster service accessor
      */
     public NodeStateManager(
         Client client,
@@ -97,7 +103,8 @@ public class NodeStateManager implements MaintenanceState, CleanState {
         ClientUtil clientUtil,
         Clock clock,
         Duration stateTtl,
-        ModelPartitioner modelPartitioner
+        ModelPartitioner modelPartitioner,
+        ClusterService clusterService
     ) {
         this.states = new ConcurrentHashMap<>();
         this.client = client;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/ad/NodeStateManager.java
```diff
@@ -235,24 +259,38 @@ public void maintenance() {
...
+     * @param detectorId Detector ID
      */
-    public void resetBackpressureCounter(String nodeId) {
-        backpressureMuter.remove(nodeId);
+    public void resetBackpressureCounter(String nodeId, String detectorId) {
+        Map<String, BackPressureRouting> routingMap = backpressureMuter.get(detectorId);
+        if (routingMap == null || routingMap.isEmpty()) {
+            backpressureMuter.remove(detectorId);
+            return;
+        }
```
ohltyler, line 292: why only remove if `routingMap` is empty?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -348,19 +349,17 @@ public void onResponse(CompositeRetriever.Page entityFeatures) {
                                 continue;
                             }
                             String modelNodeId = modelNode.getId();
-                            if (stateManager.isMuted(modelNodeId)) {
-                                LOG.info(String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s", modelNodeId));
+                            if (stateManager.isMuted(modelNodeId, detectorId)) {
+                                LOG.info(String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s for %s", modelNodeId, detectorId));
```
ohltyler, line 353: minor - maybe add some wording to distinguish the node ID vs. detector ID?

Something like:
```
LOG.info(String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s for detector %s", modelNodeId, detectorId));
```

src/main/java/org/opensearch/ad/NodeStateManager.java
kaituo: if routingMap is not empty, there can be other nodes' mapping from node id to BackPressureRouting.  For example, current mapping is {node1->BackPressureRouting1, node2->BackPressureRouting2}.  The method arguments are (node1, detectorId).  We don't want to remove everything when the client just wants to remove node1. In this example, 

backpressureMuter is a map of map: {detectorId -> {node1->BackPressureRouting1, node2->BackPressureRouting2}}, while routingMap is a map  {node1->BackPressureRouting1, node2->BackPressureRouting2}.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: sure, changed

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -577,42 +577,56 @@ private void handleException(Exception e) {
...
                     continue;
                 }
-                String rcfNodeId = rcfNode.get().getId();
-                if (stateManager.isMuted(rcfNodeId)) {
-                    LOG.info(String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s", rcfNodeId));
+
+                DiscoveryNode node = rcfNode.get();
+                String rcfNodeId = node.getId();
+                if (stateManager.isMuted(rcfNodeId, adID)) {
+                    LOG.info(String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s for %s", rcfNodeId, rcfModelID));
```
ohltyler, line 593: similar string format change here

kaituo: yes, changed

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -1056,8 +1073,14 @@ private boolean shouldStart(
...
         }
 
-        if (stateManager.isMuted(thresholdNodeId)) {
-            listener.onFailure(new InternalFailure(adID, String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s", thresholdModelID)));
+        if (stateManager.isMuted(thresholdNodeId, adID)) {
+            listener
+                .onFailure(
+                    new InternalFailure(
+                        adID,
+                        String.format(Locale.ROOT, NODE_UNRESPONSIVE_ERR_MSG + " %s for %s", thresholdNodeId, thresholdModelID)
```
ohltyler, line 1081: here also, maybe specify threshold model ID

kaituo: changed as well

src/main/java/org/opensearch/ad/NodeStateManager.java
ohltyler: Got it. What about the case of a `routingMap` containing a single mapping that will be deleted based on the passed-in `nodeId`. Shall we remove it from `backpressureMuter` as well?

Example: call the method with `node1`, and `backpressureMuter` is {detectorId -> {node1->BackPressureRouting1}}.

Can we move the `routingMap.remove(nodeId)` before doing the empty/null check, so if there is one entry it will clear everything at once?

ohltyler:
> 1)ResourceNotFoundException might be caused by an old node using RCF 1.0 that cannot recognize a new checkpoint produced by the coordinating node using compact RCF. We mute a node for a detector for 15 minutes if that node keeps returning ResourceNotFoundException (5 times in a row).

Can you describe when this situation would come up?

src/main/java/org/opensearch/ad/NodeStateManager.java
kaituo: Have to do an empty/null check before "routingMap.remove(nodeId);" as if empty/null check is true, we don't need to do "routingMap.remove(nodeId);".

In your example, after executing "routingMap.remove(nodeId);", routingMap will become empty.  Since we call resetBackpressureCounter in every detector interval, the whole routingMap will be removed in the next interval.

I can do one more empty/null check  after "routingMap.remove(nodeId)", but this seems wordy.  Since we will do it in the next interval, it seems fine.  Let me know what you think.

kaituo:
> > 1)ResourceNotFoundException might be caused by an old node using RCF 1.0 that cannot recognize a new checkpoint produced by the coordinating node using compact RCF. We mute a node for a detector for 15 minutes if that node keeps returning ResourceNotFoundException (5 times in a row).
> 
> Can you describe when this situation would come up?

This can happen during rolling upgrade or a blue/green deployment.  Will modify description.

src/main/java/org/opensearch/ad/NodeStateManager.java
ohltyler: Understood. Seems unnecessary to rewrite it in a way to clean up everything in one go if it is getting cleaned on the next interval.

ohltyler (approved):

src/main/java/org/opensearch/ad/transport/BackPressureRouting.java
```diff
@@ -88,4 +84,20 @@ public boolean isMuted() {
     private void mute() {
         lastMuteTime = clock.millis();
     }
+
+    public int getMaxRetryForUnresponsiveNode() {
+        return maxRetryForUnresponsiveNode;
+    }
+
+    public void setMaxRetryForUnresponsiveNode(int maxRetryForUnresponsiveNode) {
+        this.maxRetryForUnresponsiveNode = maxRetryForUnresponsiveNode;
```
jmazanec15, line 93: Do the setters need to be thread safe?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -812,7 +837,7 @@ private boolean invalidQuery(SearchPhaseExecutionException ex) {
         @Override
         public void onResponse(RCFResultResponse response) {
             try {
-                stateManager.resetBackpressureCounter(rcfNodeID);
+                stateManager.resetBackpressureCounter(rcfNodeID, adID);
```
jmazanec15, line 840: Why do we reset here?

src/main/java/org/opensearch/ad/transport/BackPressureRouting.java
kaituo: the method is called by addSettingsUpdateConsumer in NodeStateManager's constructor and addSettingsUpdateConsumer is a synchronized method.  So it is thread safe in our implementation.  I'll add a comment saying it is up to the client to make the method thread safe.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: A successful responses means the node is responsive and working.  Will add a comment.

jmazanec15 (approved):

jmazanec15 (approved):

ohltyler (approved):

PR #140, opened 2021-07-26
Change to use async methods instead of deprecated sync methods.

kaituo:

### Description

Testing done:
1. Manually verified the change does not impact functionality.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
@@ -120,7 +120,7 @@ public class DeleteModelTransportAction extends
                 ActionListener
                     .wrap(
                         r -> LOG.info("Deleted model for [{}] with response [{}] ", adID, r),
-                        e -> LOG.error("fail to delete model for " + adID, e)
+                        e -> LOG.error("Fail to delete model for " + adID, e)
                     )
             );
 
--- a/src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
@@ -31,6 +31,7 @@ import java.util.List;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.action.ActionListener;
 import org.opensearch.action.FailedNodeException;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.nodes.TransportNodesAction;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
```diff
@@ -113,7 +114,15 @@ protected DeleteModelNodeResponse nodeOperation(DeleteModelNodeRequest request)
...
         LOG.info("Delete model for {}", adID);
         // delete in-memory models and model checkpoint
-        modelManager.clear(adID);
+        modelManager
+            .clear(
+                adID,
+                ActionListener
+                    .wrap(
+                        r -> LOG.info("Deleted model for [{}] with response [{}] ", adID, r),
+                        e -> LOG.error("fail to delete model for " + adID, e)
```
ohltyler, line 123: super minor: capitalize to `Fail` for consistency

ohltyler (dismissed):

src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
kaituo: yeah, changed

src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
```diff
@@ -113,7 +114,15 @@ protected DeleteModelNodeResponse nodeOperation(DeleteModelNodeRequest request)
         String adID = request.getAdID();
         LOG.info("Delete model for {}", adID);
         // delete in-memory models and model checkpoint
-        modelManager.clear(adID);
+        modelManager
+            .clear(
+                adID,
+                ActionListener
+                    .wrap(
+                        r -> LOG.info("Deleted model for [{}] with response [{}] ", adID, r),
```
ylwu-amzn, line 122: When will we call this delete model action? Will it cause too many log lines for HC detector?

kaituo: We call it when people stop or delete a detector.  So it should not cause too many log lines, right?

ohltyler (approved):

src/main/java/org/opensearch/ad/transport/DeleteModelTransportAction.java
ylwu-amzn: Yes, that's ok.

ylwu-amzn (approved):

Issue #141, opened 2021-07-26
[BUG] Backward compatiblity for multi-category field change


kaituo:
**Describe the bug**
backward compatibility for changed transport input/output. Without the change, when the communication involves an old node during rolling upgrade or blue/green deployment, the transport layer sends EOFException. They may stop the detector on AD's side as AD doesn't know how to handle the exception. On OpenSearch's side, the impact is not evaluated. But according to https://tinyurl.com/ysrjwpp6, out-of-bounds read/write is one of the most critical and current software security weaknesses.

PR #142, opened 2021-07-26
Deal with backward compatibility for changed transport input/output

kaituo:

### Description

1)backward compatibility for changed transport input/output. Without the change, when the communication involves an old node during rolling upgrade or blue/green deployment, the transport layer sends EOFException. They may stop the detector on AD's side as AD doesn't know how to handle the exception. On OpenSearch's side, the impact is not evaluated. But according to https://tinyurl.com/ysrjwpp6, out-of-bounds read/write is one of the most critical and current software security weaknesses.
2)Limit the amount of redirection to match a hash ring to one. This can avoid any potential infinite loop due to any bug.

Testing done:
1. Manually verified the change in rolling upgrade tests, where we have both old (1.0) and new (1.1) versions of AD.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/141
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/EntityProfileRequest.java
+++ b/src/main/java/org/opensearch/ad/transport/EntityProfileRequest.java
@@ -32,12 +32,15 @@ import java.io.IOException;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.constant.CommonName;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.model.EntityProfileName;
+import org.opensearch.ad.util.Bwc;
 import org.opensearch.common.Strings;
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
@@ -45,16 +48,28 @@ import org.opensearch.common.xcontent.ToXContentObject;
 import org.opensearch.common.xcontent.XContentBuilder;
 
 public class EntityProfileRequest extends ActionRequest implements ToXContentObject {
+    private static final Logger LOG = LogManager.getLogger(EntityProfileRequest.class);
     public static final String ENTITY = "entity";
     public static final String PROFILES = "profiles";
     private String adID;
+    // changed from String to Entity since 1.1
     private Entity entityValue;
     private Set<EntityProfileName> profilesToCollect;
 
     public EntityProfileRequest(StreamInput in) throws IOException {
         super(in);
         adID = in.readString();
-        entityValue = new Entity(in);
+        if (Bwc.supportMultiCategoryFields(in.getVersion())) {
+            entityValue = new Entity(in);
+        } else {
+            // entity profile involving an old node won't work. Read
+            // EntityProfileTransportAction.doExecute for details. Read
+            // a string to not cause EOF exception.
+            // Cannot assign null to entityValue as old node has no logic to
+            // deal with a null entity.
+            String oldFormatEntityString = in.readString();
+            entityValue = Entity.createSingleAttributeEntity(adID, CommonName.EMPTY_FIELD, oldFormatEntityString);
+        }
         int size = in.readVInt();
         profilesToCollect = new HashSet<EntityProfileName>();
         if (size != 0) {
@@ -87,7 +102,14 @@ public class EntityProfileRequest extends ActionRequest implements ToXContentObj
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
         out.writeString(adID);
-        entityValue.writeTo(out);
+        if (Bwc.supportMultiCategoryFields(out.getVersion())) {
+            entityValue.writeTo(out);
+        } else {
+            // entity profile involving an old node won't work. Read
+            // EntityProfileTransportAction.doExecute for details. Write
+            // a string to not cause EOF exception.
+            out.writeString(entityValue.toString());
+        }
         out.writeVInt(profilesToCollect.size());
         for (EntityProfileName profile : profilesToCollect) {
             out.writeEnum(profile);
--- a/src/main/java/org/opensearch/ad/transport/EntityProfileRequest.java
+++ b/src/main/java/org/opensearch/ad/transport/EntityProfileRequest.java
@@ -32,8 +32,6 @@ import java.io.IOException;
 import java.util.HashSet;
 import java.util.Set;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.ad.constant.CommonErrorMessages;
@@ -48,7 +46,6 @@ import org.opensearch.common.xcontent.ToXContentObject;
 import org.opensearch.common.xcontent.XContentBuilder;
 
 public class EntityProfileRequest extends ActionRequest implements ToXContentObject {
-    private static final Logger LOG = LogManager.getLogger(EntityProfileRequest.class);
     public static final String ENTITY = "entity";
     public static final String PROFILES = "profiles";
     private String adID;
--- a/src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
+++ b/src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
@@ -119,10 +119,12 @@ public class EntityResultRequest extends ActionRequest implements ToXContentObje
                 Map<String, String> attributes = entry.getKey().getAttributes();
                 if (attributes.size() != 1) {
                     // cannot send a multi-category field entity to old node since it will
-                    // cause EOF exception and stop the detector. Cannot log the entity
-                    // either as there can be a huge number of entities. Since the issue
-                    // is temporary and will be gone after upgrade completes, ignore them.
-                    continue;
+                    // cause EOF exception and stop the detector. The issue
+                    // is temporary and will be gone after upgrade completes.
+                    // Since one EntityResultRequest is sent to one node, we can safely
+                    // ignore the rest of the requests.
+                    LOG.info("Skip sending multi-category entities to an incompatible node. Attributes: ", attributes);
+                    break;
                 }
                 oldFormatEntities.put(entry.getKey().getAttributes().entrySet().iterator().next().getValue(), entry.getValue());
             }
--- a/src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
+++ b/src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
@@ -29,6 +29,7 @@ package org.opensearch.ad.transport;
 import static org.opensearch.action.ValidateActions.addValidationError;
 
 import java.io.IOException;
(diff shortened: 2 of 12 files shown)


src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
```diff
@@ -89,7 +109,26 @@ public long getEnd() {
...
+            Map<String, double[]> oldFormatEntities = new HashMap<>();
+            for (Map.Entry<Entity, double[]> entry : entities.entrySet()) {
+                Map<String, String> attributes = entry.getKey().getAttributes();
+                if (attributes.size() != 1) {
+                    // cannot send a multi-category field entity to old node since it will
+                    // cause EOF exception and stop the detector. Cannot log the entity
+                    // either as there can be a huge number of entities. Since the issue
+                    // is temporary and will be gone after upgrade completes, ignore them.
+                    continue;
+                }
```
ohltyler, line 126: Can this be logged? Seems this could provide some user confusion if trying to get multi-category results in a mixed cluster.

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -189,6 +190,25 @@ protected void doExecute(Task task, EntityResultRequest request, ActionListener<
             for (Entry<Entity, double[]> entityEntry : request.getEntities().entrySet()) {
                 Entity categoricalValues = entityEntry.getKey();
 
+                Map<String, String> attrValues = categoricalValues.getAttributes();
+                if (attrValues.containsKey(CommonName.EMPTY_FIELD)
```
ohltyler, line 194: Similar logic on lines 206,207. Can this be extracted out to a helper fn?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -189,6 +190,25 @@ protected void doExecute(Task task, EntityResultRequest request, ActionListener<
...
+                            detector.getCategoryField().get(0),
+                            attrValues.get(CommonName.EMPTY_FIELD)
+                        );
+                }
+
+                attrValues = categoricalValues.getAttributes();
+                if (attrValues != null && attrValues.containsKey(CommonName.EMPTY_FIELD)) {
+                    // above remediation is not useful, skip the entity
+                    continue;
+                }
```
ohltyler, line 210: Can you explain this if check? Based on lines 198-203 won't this always be true?

src/main/java/org/opensearch/ad/transport/RCFPollingTransportAction.java
```diff
@@ -154,6 +157,9 @@ public String executor() {
                 listener.onFailure(new AnomalyDetectionException(adID, FAIL_TO_GET_RCF_UPDATE_MSG, e));
             }
 
+        } else {
+            LOG.error("Fail to poll rcf for model {} due to an unexpetected bug.", rcfModelID);
```
ohltyler, line 161: minor - typo in `unexpected`

src/main/java/org/opensearch/ad/util/Bwc.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * We are gonna start supporting multi-category field since version 1.1.0.
+     * Since Opendistro and Elasticsearch will use a version after 1.0.0, we
+     * have to check Lucene version as well. Elasticsearch 7.10 uses Lucene 8.7.0
+     *
+     * @param version test version
+     * @return whether the version support multiple category fields
+     */
+    public static boolean supportMultiCategoryFields(Version version) {
+        return version.after(Version.V_1_0_0) && version.luceneVersion.onOrAfter(org.apache.lucene.util.Version.LUCENE_8_8_2);
```
ohltyler, line 30: Can you specify in the comments that OpenSearch 1.1.0 will be based on Lucene 8.8.2?

src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
kaituo: The comment explained why I don't want to log: "Cannot log the entity either as there can be a huge number of entities. Since the issue is temporary and will be gone after upgrade completes, ignore them." Any suggestion on how to log them?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: lines 198-203 may change categoricalValues and thus the evaluation result might be different.

src/main/java/org/opensearch/ad/transport/RCFPollingTransportAction.java
kaituo: fixed

src/main/java/org/opensearch/ad/util/Bwc.java
kaituo: yeah, added.

src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
ohltyler: I was more suggesting a broader way of logging the problem once per detector - something like "Skipping sending multi-category entities to old node".

Can this be determined outside of this inner `for` loop and logged once? Wouldn't each entity in `entities` have the same number of attributes, so it only has to be determined once?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
ohltyler: Got it - but I'm still confused on what this check means and when it would be skipped. Can you help explain? I will also wait for your commit to update the check into a helper fn since that may help clarify for me too.

src/main/java/org/opensearch/ad/transport/EntityResultRequest.java
kaituo: yes, added the log like 

```
Map<String, double[]> oldFormatEntities = new HashMap<>();
            for (Map.Entry<Entity, double[]> entry : entities.entrySet()) {
                Map<String, String> attributes = entry.getKey().getAttributes();
                if (attributes.size() != 1) {
                    // cannot send a multi-category field entity to old node since it will
                    // cause EOF exception and stop the detector. The issue
                    // is temporary and will be gone after upgrade completes.
                    // Since one EntityResultRequest is sent to one node, we can safely
                    // ignore the rest of the requests.
                    LOG.info("Skip sending multi-category entities to an incompatible node. Attributes: ", attributes);
                    break;
                }
                oldFormatEntities.put(entry.getKey().getAttributes().entrySet().iterator().next().getValue(), entry.getValue());
            }
            out.writeMap(oldFormatEntities, StreamOutput::writeString, StreamOutput::writeDoubleArray);
```

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: will send new commits after testing.  The new code is like

```
/**
     * Whether the received entity comes from an node that doesn't support multi-category fields.
     * This can happen during rolling-upgrade or blue/green deployment.
     *
     * Specifically, when receiving an EntityResultRequest from an incompatible node,
     * EntityResultRequest(StreamInput in) gets an String that represents an entity.
     * But Entity class requires both an category field name and value. Since we
     * don't have access to detector config in EntityResultRequest(StreamInput in),
     * we put CommonName.EMPTY_FIELD as the placeholder.  In this method,
     * we use the same CommonName.EMPTY_FIELD to check if the deserialized entity
     * comes from an incompatible node.  If it is, we will add the field name back
     * as EntityResultTranportAction has access to the detector config object.
     *
     * @param categoricalValues deserialized Entity from inbound message.
     * @return Whether the received entity comes from an node that doesn't support multi-category fields.
     */
    private boolean isEntityeFromOldNodeMsg(Entity categoricalValues) {
        Map<String, String> attrValues = categoricalValues.getAttributes();
        return (attrValues != null && attrValues.containsKey(CommonName.EMPTY_FIELD));
    }
```

Let me know if the explanation is not clear.

ohltyler: Cool, makes sense. Only thing I don't understand still is when the if check on line 207 will be true. Seems that it would get caught in the if check on line 194 and transformed first.

src/main/java/org/opensearch/ad/util/Bwc.java
ylwu-amzn: This looks like common problem when we do version upgrade. I think OpenSearch core team should consider the case that "Opendistro and Elasticsearch will use a version after 1.0.0" in `version.after(Version v)` method, so we don't need to worry about this and check lucene version in all places/plugins.

From https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/Version.java#L222, the version comparison in `after` method is based on `id`, not sure if OpenSearch core team can make sure the `id` of OpenSearch version always greater than ES versions. Can you confirm with @nknize or @adnapibar as I can see they edited the Version class.

src/main/java/org/opensearch/ad/transport/EntityProfileRequest.java
```diff
@@ -32,29 +32,44 @@
...
 import org.opensearch.ad.model.EntityProfileName;
+import org.opensearch.ad.util.Bwc;
 import org.opensearch.common.Strings;
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
 import org.opensearch.common.xcontent.ToXContentObject;
 import org.opensearch.common.xcontent.XContentBuilder;
 
 public class EntityProfileRequest extends ActionRequest implements ToXContentObject {
+    private static final Logger LOG = LogManager.getLogger(EntityProfileRequest.class);
```
ylwu-amzn, line 51: Seems this `LOG` not being used

src/main/java/org/opensearch/ad/util/Bwc.java
kaituo: I checked the code, the id of OpenSearch and Elasticsearch has no such after guarantee you want. 

For example, OpenSearcdh V_1_1_0's id is 1010099, while ElasticSearch's V_7_10_2's id is 7100299.  If we just compare id, ElasticSearch is after V_1_1_0.  So it is wrong. I can confirm with the people you mentioned.

src/main/java/org/opensearch/ad/transport/EntityProfileRequest.java
kaituo: will remove.

src/main/java/org/opensearch/ad/util/Bwc.java
adnapibar: The version comparison works (i.e OpenSearch 1.x is after Elasticsearch v7.10.2) due to this logic https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/Version.java#L210-L211

kaituo: Thanks for confirming, Rabi.  I will remove the check for lucene version.

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: you are right.  I cannot think of a case when the check on line 207 will be true.  will remove the branch around line 207.

ylwu-amzn (approved):

ohltyler (approved):

PR #143, opened 2021-07-26
Change to use compact rcf model config

kaituo:

### Description

1)Change to use compact rcf, and float 32 precision.
2)Use zero bounding box cache for real-time detectors. The change means every RCF update/score has to recreate all of the bounding boxes. However, the change is fine for real-time detectors as we won't access the same forest a second time, at least for another minute, as the lowest resolution of a detector interval is one minute.
3)Use bounding box cache ratio 1 (i.e., all bounding box cache access is a hit) for historical detectors. Having a full bounding box intact is the original behavior of RCF 1.0. In batch mode, recreating all of the bounding boxes all the time impacts performance.  Thus, we keep the original behavior and make the bounding box cache ratio to be 1.
4)Enable shingleSize optimization. This optimization will try to save space opportunistically. Usually, we use it as shingleSize(dimension). When a new point comes in, we will look at the point store if there is any overlapping. Say the previously-stored vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize overlapping x3, x4, and only store x5, x6.

Testing done:
1. tested the change does not impact existing functionality.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -341,7 +341,11 @@ public class EntityColdStarter implements MaintenanceState {
             .precision(Precision.FLOAT_32)
             .boundingBoxCacheFraction(AnomalyDetectorSettings.REAL_TIME_BOUNDING_BOX_CACHE_RATIO)
             // same with dimension for opportunistic memory saving
-            .shingleSize(rcfNumFeatures)
+            // Usually, we use it as shingleSize(dimension). When a new point comes in, we will
+            // look at the point store if there is any overlapping. Say the previously-stored
+            // vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize
+            // overlapping x3, x4, and only store x5, x6.
+            .shingleSize(shingleSize)
             .build();
         List<double[]> allScores = new ArrayList<>();
         int totalLength = 0;
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -71,6 +71,7 @@ import org.opensearch.common.settings.Settings;
 import org.opensearch.threadpool.ThreadPool;
 
 import com.amazon.randomcutforest.RandomCutForest;
+import com.amazon.randomcutforest.config.Precision;
 
 /**
  * Training models for HCAD detectors
@@ -237,8 +238,9 @@ public class EntityColdStarter implements MaintenanceState {
                         } else {
                             logger.info("Cannot get training data for {}", modelId);
                         }
-                    } finally {
                         listener.onResponse(null);
+                    } catch (Exception e) {
+                        listener.onFailure(e);
                     }
 
                 }, exception -> {
@@ -254,8 +256,9 @@ public class EntityColdStarter implements MaintenanceState {
                         } else {
                             nodeStateManager.setException(detectorId, new AnomalyDetectionException(detectorId, cause));
                         }
-                    } finally {
                         listener.onFailure(exception);
+                    } catch (Exception e) {
+                        listener.onFailure(e);
                     }
                 });
 
@@ -331,9 +334,14 @@ public class EntityColdStarter implements MaintenanceState {
             .dimensions(rcfNumFeatures)
             .sampleSize(rcfSampleSize)
             .numberOfTrees(numberOfTrees)
-            .lambda(rcfTimeDecay)
+            .timeDecay(rcfTimeDecay)
             .outputAfter(numMinSamples)
             .parallelExecutionEnabled(false)
+            .compact(true)
+            .precision(Precision.FLOAT_32)
+            .boundingBoxCacheFraction(AnomalyDetectorSettings.REAL_TIME_BOUNDING_BOX_CACHE_RATIO)
+            // same with dimension for opportunistic memory saving
+            .shingleSize(rcfNumFeatures)
             .build();
         List<double[]> allScores = new ArrayList<>();
         int totalLength = 0;
@@ -570,8 +578,9 @@ public class EntityColdStarter implements MaintenanceState {
             try {
                 double[][] trainData = featureManager.batchShingle(samples.toArray(new double[0][0]), this.shingleSize);
                 trainModelFromDataSegments(Collections.singletonList(trainData), entity, modelState);
-            } finally {
                 listener.onResponse(null);
+            } catch (Exception e) {
+                listener.onFailure(e);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
```diff
@@ -40,18 +40,22 @@
...
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 
 import com.amazon.randomcutforest.RandomCutForest;
+import com.amazon.randomcutforest.config.Precision;
 
 /**
  * AD batch task cache which will hold RCF, threshold model, shingle and training data.
  */
 public class ADBatchTaskCache {
+    private static final Logger LOG = LogManager.getLogger(ADBatchTaskCache.class);
```
ylwu-amzn, line 58: Seems this `LOG` not being used

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
```diff
@@ -70,14 +74,20 @@ protected ADBatchTaskCache(ADTask adTask) {
...
-            .lambda(TIME_DECAY)
+            .timeDecay(TIME_DECAY)
             .sampleSize(NUM_SAMPLES_PER_TREE)
             .outputAfter(NUM_MIN_SAMPLES)
             .parallelExecutionEnabled(false)
+            .compact(true)
+            .precision(Precision.FLOAT_32)
+            .boundingBoxCacheFraction(AnomalyDetectorSettings.BATCH_BOUNDING_BOX_CACHE_RATIO)
+            // same with dimension for opportunistic memory saving
+            .shingleSize(dimensions)
```
ylwu-amzn, line 90: Why not use `detector.getShingleSize()` as RCF's shingle size? If enabled feature size is 5, and detector single size is 8, then the RCF's shingle size will be 40?

ylwu-amzn: Seems `shingleSize` is new parameter of RCF, what's the difference with `dimension`? I don't see `EntityColdStarter.java` set `shingleSize` when create RCF.

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
kaituo: will remove

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
kaituo: rcf's shingle size is different from our detector's shingle size.  " If enabled feature size is 5, and detector single size is 8, then the RCF's shingle size will be 40" that's correct.

kaituo: We do use shingleSize at line 344: https://github.com/opensearch-project/anomaly-detection/pull/143/files#diff-235791084d307db4d86529d4f484ef3e6f213af48907cc003298b8e0e2211192R344

kaituo: I explained the shingleSize parameter in the overview: "Enable shingleSize optimization. This optimization will try to save space opportunistically. Usually, we use it as shingleSize(dimension). When a new point comes in, we will look at the point store if there is any overlapping. Say the previously-stored vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize overlapping x3, x4, and only store x5, x6."

kaituo: will add this as a comment as well

sudiptoguha: Wait! For RCF 2.0 we would need to set dimensions = 40 and shinglesize =8. We can set other values of shingle size -- the code will match opportunistically but we may not see any savings. For shingle size 8, we will get a match and only add 5 new values per item

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -331,9 +334,14 @@ private void trainModelFromDataSegments(List<double[][]> dataPoints, Entity enti
...
             .numberOfTrees(numberOfTrees)
-            .lambda(rcfTimeDecay)
+            .timeDecay(rcfTimeDecay)
             .outputAfter(numMinSamples)
             .parallelExecutionEnabled(false)
+            .compact(true)
+            .precision(Precision.FLOAT_32)
+            .boundingBoxCacheFraction(AnomalyDetectorSettings.REAL_TIME_BOUNDING_BOX_CACHE_RATIO)
+            // same with dimension for opportunistic memory saving
+            .shingleSize(rcfNumFeatures)
```
sudiptoguha, line 344: I think it should be ShingleSize?

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
sudiptoguha: so just ".shingleSize(detector.getShingleSize())"

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
sudiptoguha: that is ".shingleSize(shingleSize)"

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
kaituo: right, will change

@ylwu-amzn  shingleSize in rcf and AD are the same.  Sorry for the confusion.  My explanation about the shingleSize parameter still holds

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: changed.

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
ylwu-amzn: Cool. That part looks clear now.

For this 
> Enable shingleSize optimization. This optimization will try to save space opportunistically. Usually, we use it as shingleSize(dimension). When a new point comes in, we will look at the point store if there is any overlapping. Say the previously-stored vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize overlapping x3, x4, and only store x5, x6.

Can you add it to specific line of code? Still not quite understand which part in this PR does this optimization.

kaituo: I did.  The changed code looks like

```
rcfModel = RandomCutForest
            .builder()
            .dimensions(dimensions)
            .numberOfTrees(numberOfTrees)
            .timeDecay(TIME_DECAY)
            .sampleSize(NUM_SAMPLES_PER_TREE)
            .outputAfter(NUM_MIN_SAMPLES)
            .parallelExecutionEnabled(false)
            .compact(true)
            // same with dimension for opportunistic memory saving
            // Usually, we use it as shingleSize(dimension). When a new point comes in, we will
            // look at the point store if there is any overlapping. Say the previously-stored
            // vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize
            // overlapping x3, x4, and only store x5, x6.
            .shingleSize(shingleSize)
            .build();
```

sudiptoguha: Looks good.

ylwu-amzn (approved):

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
sudiptoguha: Not seeing change though ... should I refresh/do something else?

sudiptoguha (approved):

PR #144, opened 2021-07-26
Change max shingle size and the number of trees

kaituo:

### Description

1)Change to limit max shingle size to 1024. Previously, it was unbounded. The reasons are twofold. First, it may not work on the RCF side. It is unknown that a huge shingle size could give better accuracy. Second, the larger the shingle, the more resources (e.g., CPU for tree construction/traversal, memory for models, and disk for checkpoint) RCF will use. For example, a compact RCF model with 100 trees and 932 dimensions uses around 241 MB of memory. The same model in RCF 1.0 uses 958 MB.
2)Change the number of trees used in HCAD RCF from 10 to 30 trees. Ten trees are too aggressive and may hurt precision.

Testing done:
1. tested the change does not impact existing functionality.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -45,6 +45,7 @@ import org.opensearch.ad.annotation.Generated;
 import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.constant.CommonName;
 import org.opensearch.ad.constant.CommonValue;
+import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.settings.NumericSetting;
 import org.opensearch.ad.util.ParseUtils;
 import org.opensearch.common.ParseField;
@@ -211,8 +212,13 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         if (detectionInterval == null) {
             throw new IllegalArgumentException("Detection interval should be set");
         }
-        if (shingleSize != null && shingleSize < 1) {
-            throw new IllegalArgumentException("Shingle size must be a positive integer");
+        if (shingleSize != null && (shingleSize < 1 || shingleSize > AnomalyDetectorSettings.MAX_SHINGLE_SIZE)) {
+            throw new IllegalArgumentException(
+                "Shingle size must be a positive integer no larger than "
+                    + AnomalyDetectorSettings.MAX_SHINGLE_SIZE
+                    + ". Got "
+                    + shingleSize
+            );
         }
         int maxCategoryFields = NumericSetting.maxCategoricalFields();
         if (categoryFields != null && categoryFields.size() > maxCategoryFields) {
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -99,7 +99,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     public static final String CATEGORY_FIELD = "category_field";
     public static final String USER_FIELD = "user";
     public static final String DETECTOR_TYPE_FIELD = "detector_type";
-    public static final String DETECTION_DATE_RANGE_FIELD = "detection_date_range";
 
     private final String detectorId;
     private final Long version;
@@ -118,7 +117,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     private final List<String> categoryFields;
     private User user;
     private String detectorType;
-    private DetectionDateRange detectionDateRange;
 
     /**
      * Constructor function.
@@ -175,7 +173,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
             lastUpdateTime,
             categoryFields,
             user,
-            null,
             null
         );
     }
@@ -197,8 +194,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         Instant lastUpdateTime,
         List<String> categoryFields,
         User user,
-        String detectorType,
-        DetectionDateRange detectionDateRange
+        String detectorType
     ) {
         if (Strings.isBlank(name)) {
             throw new IllegalArgumentException("Detector name should be set");
@@ -212,7 +208,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         if (detectionInterval == null) {
             throw new IllegalArgumentException("Detection interval should be set");
         }
-        if (shingleSize != null && (shingleSize < 1 || shingleSize > AnomalyDetectorSettings.MAX_SHINGLE_SIZE)) {
+        if (invalidShingleSizeRange(shingleSize)) {
             throw new IllegalArgumentException(
                 "Shingle size must be a positive integer no larger than "
                     + AnomalyDetectorSettings.MAX_SHINGLE_SIZE
@@ -244,12 +240,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         this.categoryFields = categoryFields;
         this.user = user;
         this.detectorType = detectorType;
-        this.detectionDateRange = detectionDateRange;
-
-        // TODO: remove this check when we support HC historical detector
-        if (!isRealTimeDetector(detectionDateRange) && categoryFields != null && categoryFields.size() > 0) {
-            throw new IllegalArgumentException("Don't support high cardinality historical detector now");
-        }
     }
 
     public AnomalyDetector(StreamInput input) throws IOException {
@@ -271,11 +261,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
             this.user = new User(input);
         } else {
             user = null;
-        }
-        if (input.readBoolean()) {
-            detectionDateRange = new DetectionDateRange(input);
-        } else {
-            detectionDateRange = null;
         }
         detectorType = input.readOptionalString();
         if (input.readBoolean()) {
@@ -310,12 +295,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
             user.writeTo(output);
         } else {
             output.writeBoolean(false); // user does not exist
-        }
-        if (detectionDateRange != null) {
-            output.writeBoolean(true); // detectionDateRange exists
-            detectionDateRange.writeTo(output);
-        } else {
-            output.writeBoolean(false); // detectionDateRange does not exist
         }
         output.writeOptionalString(detectorType);
         if (uiMetadata != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -644,6 +651,25 @@ private AnomalyDetectorSettings() {}
...
+    // ======================================
+    // Checkpoint setting
+    // ======================================
+    // we won't accept a checkpoint larger than 30MB. Or we risk OOM.
+    // For reference, in RCF 1.0, the checkpoint of a RCF with 50 trees, 10 dimensions,
+    // 256 samples is of 3.2MB.
+    // In compact rcf, the same RCF is of 163KB.
+    // Since we allow at most 5 features, and the default shingle size is 8 and default
+    // tree number size is 100, we can have at most 25.6 MB. It is possible that cx increases
+    // the max features or shingle size, but we don't want to risk OOM for the flexibility.
```
ohltyler, line 663: Any experiments done on increasing these defaults with compact RCF, like higher default tree size or shingle size?

kaituo: There are some experiments here done by Sudipto/Josh: https://quip-amazon.com/A8hTAbgyHFtF/Random-Cut-Forest-serialization-benchmarks

ohltyler: Ok, will take a look. As a reminder, should not have inaccessible links in public PRs 😃

ohltyler (dismissed):

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -211,8 +212,13 @@ public AnomalyDetector(
         if (detectionInterval == null) {
             throw new IllegalArgumentException("Detection interval should be set");
         }
-        if (shingleSize != null && shingleSize < 1) {
-            throw new IllegalArgumentException("Shingle size must be a positive integer");
+        if (shingleSize != null && (shingleSize < 1 || shingleSize > AnomalyDetectorSettings.MAX_SHINGLE_SIZE)) {
```
ylwu-amzn, line 215: Thanks for adding this check. How about build a util method so historical analysis can reuse it?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -222,6 +222,21 @@ private AnomalyDetectorSettings() {}
             Setting.Property.Dynamic
         );
 
+    // for a batch operation, we want all of the bounding box in-place for speed
+    public static final double BATCH_BOUNDING_BOX_CACHE_RATIO = 1;
+
+    // for a real-time operation, we trade off speed for memory as real time opearation
+    // only has to do one update/scoring per interval
+    public static final double REAL_TIME_BOUNDING_BOX_CACHE_RATIO = 0;
```
ylwu-amzn, line 230: Does this mean we need to rebuild bounding box for every realtime job run? How much more memory will cost if we open this bounding box cache per RCF model?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -222,6 +222,21 @@ private AnomalyDetectorSettings() {}
...
+
+    public static final int DEFAULT_SHINGLE_SIZE = 8;
+
+    // Multi-entity detector model setting:
+    // TODO (kaituo): change to 4
```
ylwu-amzn, line 235: This TODO means we need to change this hard coded HCAD shingle size from 1 to 4? Or we should set the HCAD shingle size upper limit as 4 (or some other value)?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
kaituo: sure, added

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: Different config have different numbers. For 30 trees, 2 dimensions, using full bounding box costs 935 KB, while not using bounding box  costs 322 KB.

kaituo: Yes, it means we need to rebuild for every real time job run.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: It means we should have default value 4 but allow people to change like single-stream detectors.

ylwu-amzn: Got it. Should we set different upper limit of shingle size for single-stream detector and HCAD?

kaituo: We can do that when there is a need.

ylwu-amzn: sure

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: right, next time will send you offline.

sudiptoguha (dismissed):
ylwu-amzn should eyeball changes too. Looks good to me. Question: is the historical analysis using 100 trees? Did anyone try the getApproximateScore()? That may speed up considerably for larger number of trees.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
ylwu-amzn: Not sure if I missed something, but seems the util method not added yet?

kaituo: not yet.  Will send the new commit after retesting.

kaituo:
> ylwu-amzn should eyeball changes too. Looks good to me. Question: is the historical analysis using 100 trees? Did anyone try the getApproximateScore()? That may speed up considerably for larger number of trees.

for HCAD, a historical detector uses 30 trees; for single stream  a historical detector uses 100 trees. As we discussed offline, using getApproximateScore for single stream complicates the logic we have multiple-way split and run distributed models.

ylwu-amzn:
> > ylwu-amzn should eyeball changes too. Looks good to me. Question: is the historical analysis using 100 trees? Did anyone try the getApproximateScore()? That may speed up considerably for larger number of trees.
> 
> for HCAD, a historical detector uses 30 trees; for single stream a historical detector uses 100 trees. As we discussed offline, using getApproximateScore for single stream complicates the logic as we have multiple-way split and run distributed models.

Not sure if understand it correctly. I think `getApproximateScore` will run faster but it may get approximate score. @sudiptoguha, can you confirm if that approximate score will be highly confident and won't impact the detection result? 
For historical analysis, we have add rate limiting, we run 1000 detection interval in one batch, then run the next batch in 5 seconds. So the speed is not top bottleneck for historical analysis considering the batch interval time(5 seconds) is dominating the speed. But that may help for HC historical analysis if we multiply by huge amount of entities.

sudiptoguha:
> > > ylwu-amzn should eyeball changes too. Looks good to me. Question: is the historical analysis using 100 trees? Did anyone try the getApproximateScore()? That may speed up considerably for larger number of trees.
> > 
> > 
> > for HCAD, a historical detector uses 30 trees; for single stream a historical detector uses 100 trees. As we discussed offline, using getApproximateScore for single stream complicates the logic as we have multiple-way split and run distributed models.
> 
> Not sure if understand it correctly. I think `getApproximateScore` will run faster but it may get approximate score. @sudiptoguha, can you confirm if that approximate score will be highly confident and won't impact the detection result?
> For historical analysis, we have add rate limiting, we run 1000 detection intervals in one batch, then run the next batch in 5 seconds. So the speed is not top bottleneck for historical analysis considering the batch interval time(5 seconds) is dominating the speed. But that may help for HC historical analysis if we multiply by huge amount of entities.

When the model has enough samples (not the cold start scenario) then the approximateScore() will be fine -- those are the tests I saw. Not sure about both cold start (low data) and approximation ... I think it may be a good call to enable approximation for the "second 1000 intervals onwards" -- what do you think? Because when we have 1000+ intervals we must have higher threshold to not have too many anomaly alerts, and in that case we can use approximation. Also I would not use approximation for 30 trees (its already small) -- for 100 trees, the benefit kicks in because we would use 30ish trees in the average case (to decide the easy "not an anomaly" and easy "anomaly" cases) -- we may use the full 100 to decide the ambiguous cases.

kaituo:
> > > > ylwu-amzn should eyeball changes too. Looks good to me. Question: is the historical analysis using 100 trees? Did anyone try the getApproximateScore()? That may speed up considerably for larger number of trees.
> > > 
> > > 
> > > for HCAD, a historical detector uses 30 trees; for single stream a historical detector uses 100 trees. As we discussed offline, using getApproximateScore for single stream complicates the logic as we have multiple-way split and run distributed models.
> > 
> > 
> > Not sure if understand it correctly. I think `getApproximateScore` will run faster but it may get approximate score. @sudiptoguha, can you confirm if that approximate score will be highly confident and won't impact the detection result?
> > For historical analysis, we have add rate limiting, we run 1000 detection intervals in one batch, then run the next batch in 5 seconds. So the speed is not top bottleneck for historical analysis considering the batch interval time(5 seconds) is dominating the speed. But that may help for HC historical analysis if we multiply by huge amount of entities.
> 
> When the model has enough samples (not the cold start scenario) then the approximateScore() will be fine -- those are the tests I saw. Not sure about both cold start (low data) and approximation ... I think it may be a good call to enable approximation for the "second 1000 intervals onwards" -- what do you think? Because when we have 1000+ intervals we must have higher threshold to not have too many anomaly alerts, and in that case we can use approximation. Also I would not use approximation for 30 trees (its already small) -- for 100 trees, the benefit kicks in because we would use 30ish trees in the average case (to decide the easy "not an anomaly" and easy "anomaly" cases) -- we may use the full 100 to decide the ambiguous cases.

@sudiptoguha  I changed single stream detectors' trees to 30 according to your suggestion.  So all RCF (including historical and single-stream) models use 30 trees inside AD.

sudiptoguha (approved):

ylwu-amzn (approved):

Issue #151, opened 2021-07-30
Disable backward compatibility for testing on 1.0.0


kaituo:
**Is your feature request related to a problem? Please describe.**
For multi-category field testing, previously we enforced the version to be after 1.0.0.  This brings inconvenience for testing as version 1.1.0 is not released yet.  To ease testing, we temporarily disable backward compatibility logic.  Need to change it back when releasing 1.1.

PR #152, opened 2021-07-30
Bug fixes and disable BWC

kaituo:
### Description
This PR did the following:
1)Fixed a bug in ModelPartitioner that we should use estimated rcf instead of total model size (including rcf and threshold models).
2)Disabled backward-compatibility check so that we can run multi-category detectors in OpenSearch 1.0.
3)remove unused fields in MemoryTracker
4)add null check when deserializing the sample vector

Testing done:
1)added unit tests
2)did e2e tests for the basic single-stream detector workflow.
 
### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/151
 
### Check List


dblock:
@kaituo This is confusing to me, but I don't know the codebase, can you explain a little bit more? Why would multi-category field testing cause bcw testing to fail? Is there a way to fix this without disabling a very important set of tests?

ylwu-amzn (dismissed):

kaituo:
> @kaituo This is confusing to me, but I don't know the codebase, can you explain a little bit more? Why would multi-category field testing cause bcw testing to fail? Is there a way to fix this without disabling a very important set of tests?

@dblock Multi-category field testing doesn't cause bcw testing to fail. Multi-category field testing depends on bwc testing.  Bwc testing requires OpenSearch 1.1 (https://github.com/opensearch-project/anomaly-detection/pull/142) and we don't have OpenSearch 1.1 yet.  The dependencies on 1.1 makes it inconvenient for others to test multi-category field as they have to use a home-made 1.1.  The plan is that I keep testing by enabling bwc locally and using a home-made Opensearch 1.1.  When an official 1.1 is ready for testing,  I will change it back and re-enable bwc.

src/main/java/org/opensearch/ad/util/Bwc.java
```diff
@@ -18,13 +18,16 @@
...
+
     /**
      * We are gonna start supporting multi-category fields since version 1.1.0.
      *
      * @param version test version
      * @return whether the version support multiple category fields
      */
     public static boolean supportMultiCategoryFields(Version version) {
-        return version.after(Version.V_1_0_0);
+        // TODO: remove DISABLE_BWC before release
```
jmazanec15, line 30: Which release? I would recommend adding a more descriptive comment here.

src/main/java/org/opensearch/ad/util/Bwc.java
```diff
@@ -18,13 +18,16 @@
  *
  */
 public class Bwc {
+    public static boolean DISABLE_BWC = true;
```
jmazanec15, line 21: Why does this need to be a variable?

jmazanec15 (commented):
@kaituo So does this happen every minor release? When an official next minor version is ready, you submit a PR to reenable and then after the release you disable again?

src/main/java/org/opensearch/ad/util/Bwc.java
kaituo: added

src/main/java/org/opensearch/ad/util/Bwc.java
kaituo: I am using it in testing https://github.com/opensearch-project/anomaly-detection/blob/df40978d82c0ea652b05d3c9869e71b296789a20/src/test/java/org/opensearch/BwcTests.java#L102

Also, this flag enables us to easily switch on and off this feature without worrying about introducing bugs while changing code.

kaituo:
> @kaituo So does this happen every minor release? When an official next minor version is ready, you submit a PR to reenable and then after the release you disable again?

@jmazanec15  if we don't change transport request/response that can cause bwc issue again, we only has to do the disable/enable once. But I think that will happen again in the future. So we may have to do it again and again.  Any suggestion to avoid this?

src/main/java/org/opensearch/ad/NodeStateManager.java
```diff
@@ -363,21 +362,22 @@ public void setException(String detectorId, Exception e) {
...
         }
 
-        AnomalyDetectionException adExep = null;
-        if (e instanceof AnomalyDetectionException) {
-            adExep = (AnomalyDetectionException) e;
-        } else {
-            adExep = new AnomalyDetectionException(detectorId, e);
-        }
-        state.setException(adExep);
+        // AnomalyDetectionException adExep = null;
```
jmazanec15, line 373: Either remove commented code or add descriptive comment about why it is commented.

jmazanec15 (commented):
@kaituo Im wondering if it would be better to add a gradle arg that disables bwc compatibility check as opposed to manually coding it in. So, by default it would be enabled, but if you want to test without it, you could disable it. Not sure how feasible that is though. 

Also, why did you merge https://github.com/opensearch-project/anomaly-detection/pull/152/commits/daf104a1eca785bdc76e08de5bbbf5bfcc15d24c? Isnt this a part of another review?

src/main/java/org/opensearch/ad/NodeStateManager.java
kaituo: my bad.  reverted https://github.com/opensearch-project/anomaly-detection/commit/daf104a1eca785bdc76e08de5bbbf5bfcc15d24c

kaituo: About the gradle change: We need to use System.getProperty in the Bwc class to check a flag: if the system property is true, then set DISABLE_BWC to true; otherwise, set DISABLE_BWC to false.  It is doable. But the scope is much larger: we need to make tasks like gradle run to accept the property.  Tasks like grade run are defined in Opensearch core and we inherited the build script.

Also, it is not just gradle.  People need it to build domain and create e2e testing.  We need to add the system property to the start up script of Opensearch and managed service.

These stuff is out of AD and hard to control/maintain.  Any other ways to get this to work?

kaituo: @dblock Please also comment if you have any suggestion.

jmazanec15 (approved):

ylwu-amzn (approved):
LGTM

PR #157, opened 2021-08-04
Handle more AD exceptions thrown over the wire/network

kaituo:
### Description
OpenSearch restricts the kind of exceptions that can be thrown over the wire (Read OpenSearchException.OpenSearchExceptionHandle). Since we cannot add our exception like ResourceNotFoundException without modifying OpenSearch's code, we must unwrap the NotSerializableExceptionWrapper and check its root cause message. This PR adds checks all AD exceptions in such cases. Previously, we only checked a couple of them.

Previously, we wrap all exceptions thrown in AD using one of AnomalyDetectionExceptions. The wrap brings a complication to NotSerializableExceptionWrapper thrown over the wire as NotSerializableExceptionWrapper won't keep the original AnomalyDetectionExceptions object and only keeps the cause message. Therefore, we won't be able to restore the original exception encapsulated inside AnomalyDetectionExceptions. This PR stops wrapping the original exception inside AnomalyDetectionExceptions.

This PR also adds a null check inside EntityModel in case of a potential null pointer exception.

Testing done:
1. Added unit tests to check if exceptions wrapped inside NotSerializableExceptionWrapper can be decoded.
2. Did e2e testing for basic single-stream and HCAD workflow.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/common/exception/NotSerializedADExceptionName.java
+++ b/src/main/java/org/opensearch/ad/common/exception/NotSerializedADExceptionName.java
@@ -11,12 +11,18 @@
 
 package org.opensearch.ad.common.exception;
 
+import java.util.Optional;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
 import org.opensearch.OpenSearchException;
+import org.opensearch.common.io.stream.NotSerializableExceptionWrapper;
 
 /**
  * OpenSearch restricts the kind of exceptions that can be thrown over the wire
- * (Read OpenSearchException.OpenSearchExceptionHandle). Since we cannot
- * add our own exception like ResourceNotFoundException without modifying
+ * (Read OpenSearchException.OpenSearchExceptionHandle https://tinyurl.com/wv6c6t7x).
+ * Since we cannot add our own exception like ResourceNotFoundException without modifying
  * OpenSearch's code, we have to unwrap the NotSerializableExceptionWrapper and
  * check its root cause message.
  *
@@ -32,6 +38,7 @@ public enum NotSerializedADExceptionName {
     CANCELLATION_EXCEPTION_NAME_UNDERSCORE(OpenSearchException.getExceptionName(new ADTaskCancelledException("", ""))),
     DUPLICATE_TASK_EXCEPTION_NAME_UNDERSCORE(OpenSearchException.getExceptionName(new DuplicateTaskException("")));
 
+    private static final Logger LOG = LogManager.getLogger(NotSerializedADExceptionName.class);
     private final String name;
 
     NotSerializedADExceptionName(String name) {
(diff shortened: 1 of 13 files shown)


src/main/java/org/opensearch/ad/common/exception/NotSerializedADExceptionName.java
```diff
@@ -0,0 +1,44 @@
...
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.common.exception;
+
+import org.opensearch.OpenSearchException;
+
+/**
+ * OpenSearch restricts the kind of exceptions that can be thrown over the wire
+ * (Read OpenSearchException.OpenSearchExceptionHandle). Since we cannot
```
jmazanec15, line 18: I might add a link to the comment: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/OpenSearchException.java#L756-L762

src/main/java/org/opensearch/ad/common/exception/NotSerializedADExceptionName.java
```diff
@@ -0,0 +1,44 @@
...
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.common.exception;
+
+import org.opensearch.OpenSearchException;
+
+/**
+ * OpenSearch restricts the kind of exceptions that can be thrown over the wire
```
jmazanec15, line 17: Also, does over the wire refer to return exception to client or return exception across transport layer within cluster?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -1042,7 +1058,7 @@ private void handleThresholdResult() {
         }
     }
 
-    private void handlePredictionFailure(Exception e, String adID, String nodeID, AtomicReference<AnomalyDetectionException> failure) {
+    private void handlePredictionFailure(Exception e, String adID, String nodeID, AtomicReference<Exception> failure) {
```
jmazanec15, line 1061: Why are there 2 exception arguments in this function?

src/main/java/org/opensearch/ad/util/ExceptionUtil.java
```diff
@@ -88,6 +95,56 @@ public static boolean isException(Throwable exception, Class<? extends Exception
...
+     * @return converted AnomalyDetectionException
+     */
+    public static Optional<AnomalyDetectionException> convertWrappedAnomalyDetectionException(
+        NotSerializableExceptionWrapper exception,
+        String adID
+    ) {
+        String exceptionMsg = exception.getMessage().trim();
+
+        AnomalyDetectionException convertedException = null;
+        for (NotSerializedADExceptionName adException : NotSerializedADExceptionName.values()) {
```
jmazanec15, line 112: Do you think that instead of looping over the values in NotSerializedADExceptionName, it would be better to make a static method in NotSerializedADExceptionName that takes an exception message and returns the corresponding ad exception? 

It seems like this switch block should not be located outside of NotSerializedADExceptionName.

src/main/java/org/opensearch/ad/common/exception/NotSerializedADExceptionName.java
kaituo: added

src/main/java/org/opensearch/ad/common/exception/NotSerializedADExceptionName.java
kaituo: over the wire means transferred using network.  So the exception comes from a remote node.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: this is indeed confusing.  I'll rewrite and add comments.

src/main/java/org/opensearch/ad/util/ExceptionUtil.java
kaituo: good idea.  will change

jmazanec15 (approved):

src/main/java/org/opensearch/ad/ml/EntityModel.java
```diff
@@ -61,6 +62,9 @@ public EntityModel(Entity entity, Queue<double[]> samples, RandomCutForest rcf,
     }
 
     public void addSample(double[] sample) {
+        if (this.samples == null) {
```
ylwu-amzn, line 65: Should we add this null checking in `clear(0` function as well?

kaituo: yes, fixed.

ylwu-amzn (approved):

ylwu-amzn (approved):

jmazanec15 (approved):

PR #159, opened 2021-08-06
support historical analysis for multi-category HC

ylwu-amzn:

### Description

1. Support historical analysis of multi-category HC
2. Delete anomaly result when delete detector and add to hourly cron job
3. Fix HC detector query in historical analysis

 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
@@ -221,6 +221,28 @@ public class SearchFeatureDao extends AbstractRetriever {
      * @param listener listener to return back the entities
      */
     public void getHighestCountEntities(AnomalyDetector detector, long startTime, long endTime, ActionListener<List<Entity>> listener) {
+        getHighestCountEntities(detector, startTime, endTime, maxEntitiesForPreview, minimumDocCountForPreview, pageSize, listener);
+    }
+
+    /**
+     * Get list of entities with high count in descending order within specified time range
+     * @param detector detector config
+     * @param startTime start time of time range
+     * @param endTime end time of time range
+     * @param maxEntitiesSize max top entities
+     * @param minimumDocCount minimum doc count for top entities
+     * @param pageSize page size when query multi-category HC detector's top entities
+     * @param listener listener to return back the entities
+     */
+    public void getHighestCountEntities(
+        AnomalyDetector detector,
+        long startTime,
+        long endTime,
+        int maxEntitiesSize,
+        int minimumDocCount,
+        int pageSize,
+        ActionListener<List<Entity>> listener
+    ) {
         if (!detector.isMultientityDetector()) {
             listener.onResponse(null);
             return;
@@ -237,7 +259,7 @@ public class SearchFeatureDao extends AbstractRetriever {
         AggregationBuilder bucketAggs = null;
 
         if (detector.getCategoryField().size() == 1) {
-            bucketAggs = AggregationBuilders.terms(AGG_NAME_TOP).size(maxEntitiesForPreview).field(detector.getCategoryField().get(0));
+            bucketAggs = AggregationBuilders.terms(AGG_NAME_TOP).size(maxEntitiesSize).field(detector.getCategoryField().get(0));
         } else {
             /*
              * We don't have an efficient solution for terms aggregation on multiple fields.
@@ -339,7 +361,7 @@ public class SearchFeatureDao extends AbstractRetriever {
                 .subAggregation(
                     PipelineAggregatorBuilders
                         .bucketSort("bucketSort", Arrays.asList(new FieldSortBuilder("_count").order(SortOrder.DESC)))
-                        .size(maxEntitiesForPreview)
+                        .size(maxEntitiesSize)
                 );
         }
 
@@ -352,7 +374,14 @@ public class SearchFeatureDao extends AbstractRetriever {
         client
             .search(
                 searchRequest,
-                new TopEntitiesListener(listener, detector, searchSourceBuilder, clock.millis() + previewTimeoutInMilliseconds)
+                new TopEntitiesListener(
+                    listener,
+                    detector,
+                    searchSourceBuilder,
+                    clock.millis() + previewTimeoutInMilliseconds,
+                    maxEntitiesSize,
+                    minimumDocCount
+                )
             );
     }
 
@@ -362,18 +391,24 @@ public class SearchFeatureDao extends AbstractRetriever {
         private List<Entity> topEntities;
         private SearchSourceBuilder searchSourceBuilder;
         private long expirationEpochMs;
+        private long minimumDocCount;
+        private int maxEntitiesSize;
 
         TopEntitiesListener(
             ActionListener<List<Entity>> listener,
             AnomalyDetector detector,
             SearchSourceBuilder searchSourceBuilder,
-            long expirationEpochMs
+            long expirationEpochMs,
+            int maxEntitiesSize,
+            int minimumDocCount
         ) {
             this.listener = listener;
             this.detector = detector;
             this.topEntities = new ArrayList<>();
             this.searchSourceBuilder = searchSourceBuilder;
             this.expirationEpochMs = expirationEpochMs;
+            this.maxEntitiesSize = maxEntitiesSize;
+            this.minimumDocCount = minimumDocCount;
         }
 
         @Override
@@ -402,10 +437,7 @@ public class SearchFeatureDao extends AbstractRetriever {
                         .map(bucket -> bucket.getKeyAsString())
                         .collect(Collectors.toList())
                         .stream()
-                        .map(
-                            entityValue -> Entity
-                                .createSingleAttributeEntity(detector.getDetectorId(), detector.getCategoryField().get(0), entityValue)
-                        )
+                        .map(entityValue -> Entity.createSingleAttributeEntity(detector.getCategoryField().get(0), entityValue))
                         .collect(Collectors.toList());
                     listener.onResponse(topEntities);
                 } else {
@@ -413,16 +445,16 @@ public class SearchFeatureDao extends AbstractRetriever {
                     List<Entity> pageResults = compositeAgg
                         .getBuckets()
                         .stream()
-                        .filter(bucket -> bucket.getDocCount() >= minimumDocCountForPreview)
-                        .map(bucket -> Entity.createEntityByReordering(detector.getDetectorId(), bucket.getKey()))
+                        .filter(bucket -> bucket.getDocCount() >= minimumDocCount)
+                        .map(bucket -> Entity.createEntityByReordering(bucket.getKey()))
                         .collect(Collectors.toList());
                     // we only need at most maxEntitiesForPreview
-                    int amountToWrite = maxEntitiesForPreview - topEntities.size();
+                    int amountToWrite = maxEntitiesSize - topEntities.size();
                     for (int i = 0; i < amountToWrite && i < pageResults.size(); i++) {
                         topEntities.add(pageResults.get(i));
                     }
                     Map<String, Object> afterKey = compositeAgg.afterKey();
-                    if (topEntities.size() >= maxEntitiesForPreview || afterKey == null) {
+                    if (topEntities.size() >= maxEntitiesSize || afterKey == null) {
                         listener.onResponse(topEntities);
                     } else if (expirationEpochMs < clock.millis()) {
                         if (topEntities.isEmpty()) {
@@ -541,11 +573,12 @@ public class SearchFeatureDao extends AbstractRetriever {
 
     public void getFeaturesForPeriodByBatch(
         AnomalyDetector detector,
+        Entity entity,
         long startTime,
         long endTime,
         ActionListener<Map<Long, Optional<double[]>>> listener
     ) throws IOException {
-        SearchSourceBuilder searchSourceBuilder = batchFeatureQuery(detector, startTime, endTime, xContent);
+        SearchSourceBuilder searchSourceBuilder = batchFeatureQuery(detector, entity, startTime, endTime, xContent);
         logger.debug("Batch query for detector {}: {} ", detector.getDetectorId(), searchSourceBuilder);
 
         SearchRequest searchRequest = new SearchRequest(detector.getIndices().toArray(new String[0])).source(searchSourceBuilder);
--- a/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
@@ -378,6 +378,7 @@ public class SearchFeatureDao extends AbstractRetriever {
                     listener,
                     detector,
                     searchSourceBuilder,
+                    // TODO: tune timeout for historical analysis based on performance test result
                     clock.millis() + previewTimeoutInMilliseconds,
                     maxEntitiesSize,
                     minimumDocCount
--- a/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
+++ b/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
@@ -387,7 +387,7 @@ public class ADBatchTaskRunner {
             internalHCListener.onFailure(e);
         });
 
-        int minimumDocCount = Math.min((int) (bucketInterval / adTask.getDetector().getDetectorIntervalInMilliseconds()), 1);
+        int minimumDocCount = Math.min((int) (bucketInterval / adTask.getDetector().getDetectorIntervalInMilliseconds()) / 2, 1);
         searchFeatureDao
             .getHighestCountEntities(
                 adTask.getDetector(),
--- a/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
+++ b/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
@@ -311,23 +311,23 @@ public class ADBatchTaskRunner {
                 adTask.getDetectionDateRange().getStartTime().toEpochMilli(),
                 MAX_TOP_ENTITIES_LIMIT_FOR_HISTORICAL_ANALYSIS
             );
-            long interval = adTask.getDetector().getDetectorIntervalInMilliseconds();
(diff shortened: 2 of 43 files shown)


src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1514,6 +1520,7 @@ public void deleteADTasks(String detectorId, AnomalyDetectorFunction function, A
         client.execute(DeleteByQueryAction.INSTANCE, request, ActionListener.wrap(r -> {
             if (r.getBulkFailures() == null || r.getBulkFailures().size() == 0) {
                 logger.info("AD tasks deleted for detector {}", detectorId);
+                deleteADResultOfDetector(detectorId);
```
kaituo, line 1523: I saw you call deleteADTasks inside DeleteAnomalyDetectorTransportAction. Do we wait until all results are deleted before returning delete success to users? There can be a huge amount of results.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -352,7 +374,14 @@ public void getHighestCountEntities(AnomalyDetector detector, long startTime, lo
         client
             .search(
                 searchRequest,
-                new TopEntitiesListener(listener, detector, searchSourceBuilder, clock.millis() + previewTimeoutInMilliseconds)
+                new TopEntitiesListener(
+                    listener,
+                    detector,
+                    searchSourceBuilder,
+                    clock.millis() + previewTimeoutInMilliseconds,
```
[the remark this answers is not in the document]
ylwu-amzn: Need to do more performance test to decide. Will add TODO here.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -349,8 +421,8 @@ private void searchTopEntities(
 
         String topEntitiesAgg = "topEntities";
         AggregationBuilder aggregation = new TermsAggregationBuilder(topEntitiesAgg)
-            .field(adTask.getDetector().getCategoryField().get(0)) // support multi category fields
-            .size(MAX_TOP_ENTITIES_LIMIT_FOR_HISTORICAL_ANALYSIS);
+            .field(adTask.getDetector().getCategoryField().get(0))
+            .size(MAX_PAGE_SIZE_FOR_HISTORICAL_ANALYSIS);
```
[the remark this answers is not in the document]
ylwu-amzn: Plan to add some dynamic setting to let user tune how many top entities can run for HC. User may choose a bigger number than 10, 000, that exceeds the default page size limit.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: No, `deleteADResultOfDetector` just send out delete request, don't wait until it's done. After delete request sent out, will continue to delete AD job and return response to user.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -315,19 +321,85 @@ public void getTopEntities(ADTask adTask, ActionListener<String> internalHCListe
...
+        });
+        searchFeatureDao
+            .getHighestCountEntities(
+                adTask.getDetector(),
                 dataStartTime,
-                dataStartTime + interval,
-                internalHCListener
+                dataEndTime,
+                MAX_TOP_ENTITIES_LIMIT_FOR_HISTORICAL_ANALYSIS,
+                1,
```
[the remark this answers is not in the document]
kaituo: you missed my comment.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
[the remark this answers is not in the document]
kaituo: Not sure I understand you.  This is terms aggregation.  You cannot do pagination if the number is more than 10,000.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
kaituo: Ah I see now. 

I am hesitated on this approach because we can query/delete a huge number of docs.  we can have 10 shards each shard 65 GB for each rolled over index.  And the data is stored for 30 days.  We don't know how many result indices there.  Have you measured the performance impacts? The performance will be impacted depends on the index number/size and cluster resources/workload. Is there a max time before cancelling? There are throttling like mentioned in https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html Do you consider using it?  

See the quote:
"However, this query is really slow. It scales linearly with document size. If you have enough documents that you need to be rotating them to prevent your Elasticsearch instance from bursting into flames, you probably can’t delete records this way, and will need to use time-based indices instead."

https://www.cloudsavvyit.com/7152/how-to-rotate-and-delete-old-elasticsearch-records-after-a-month/

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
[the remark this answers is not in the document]
ylwu-amzn: Oh, thanks for reminding. 
If each bucket is just one detector interval, then we should count in the entity even it has only 1 doc. Will priority tracker prioritize entities with bigger doc count?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: With historical analysis, user can generate huge amount of anomaly result by running a long period of time and even more by running HC detector. That may make AD result index roll over quickly, that will impact realtime AD result retention period. Will do some performance test later and tune this part. One option is we can add dynamic setting to let user control delete anomaly result or not as it's hard to say what's the user's most concern. If we have concern about performance based on the test result, we can set it as false by default.

ylwu-amzn: We changed a lot code recently, will redo performance test to tight up before release. Create one issue to track https://github.com/opensearch-project/anomaly-detection/issues/172

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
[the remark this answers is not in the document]
ylwu-amzn: That depends the performance test result. If the multi-category search performance is good for historical analysis. We can use the similar way for single-category HC. 
Anyway, I will roll back the constant to `MAX_TOP_ENTITIES_LIMIT_FOR_HISTORICAL_ANALYSIS`, so we don't have confusion.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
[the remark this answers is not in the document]
kaituo: priority tracker won't. The number is ideally a number between 1 and piece duration/detector interval.  Using 1 is easy, but may sacrifice the actual entities detected for historical range more than 1,000 intervals.  I'll leave you decide.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
kaituo: I remember you created an API for users to delete results by themselves, right?  This can be a mitigation if issue arise.

A common solution for this problem is mark the stuff to delete and do that when the system is not busy. 

How about you disable this by default before your perf test indicate otherwise?

Check these posts, delete by query is not suitable for the use case.

https://discuss.elastic.co/t/searching-performance-dramatically-decreases-15s-if-the-indexation-process-is-enabled/37833/11

https://discuss.elastic.co/t/delete-by-query-performance-optimization/244437

https://discuss.elastic.co/t/alternative-for-delete-by-query-or-solution-for-performance-improvement/272298

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -792,12 +791,6 @@ public void startADBatchTaskOnWorkerNode(
                 adTaskManager.cleanDetectorCache(adTask, transportService, () -> handleException(adTask, e));
             } else {
                 adTaskManager.entityTaskDone(adTask, e, transportService);
-                threadPool
```
kaituo: So if there is an exception, we will not start the remaining entity task and just stop?

kaituo: talked with you offline.  Summary for others: entityTaskDone will send a msg from the worker node to the coordinating node.  The coordinating node will put the failing task to the end of running queue and start a new task lane (may fail if others steal the lane in between the messaging) for the current head entity task on the queue.  Without removing the code, the worker node will restart the failing task (max retries: 3).  When the coordinating node comes to check, it finds the lane is already used and won't start new task.  After removing the code, the coordinating node will sends a new task to start.  The benefit of removing the code is that the failing task might be failing quickly again.  It is better to try other tasks before exhausting the 3 retries.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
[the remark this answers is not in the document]
ylwu-amzn: Have some misunderstanding about priority tracker. Based on our discussion, we can use 1/2 of total detector intervals in each bucket. Will tune this part.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: Thanks, as we discussed, for some smaller cluster which doesn't have big data volume, it's better to provide some auto way to help cleanup anomaly result. Agree that it's risky for big clusters. I will add dynamic setting for this and set default as close.

kaituo:
@ylwu-amzn Are your changes in?  Please don't force push as it is hard for reviewers to see the diff.  Also, the PR build failed now.

ylwu-amzn:
> @ylwu-amzn Are your changes in? Please don't force push as it is hard for reviewers to see the diff. Also, the PR build failed now.

Thanks for reminding. Pushed changes accidentally. Have fixed the failed build. BTW, you can stop review if the build failed.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -386,13 +386,15 @@ private void searchTopEntitiesForMultiCategoryHC(
             logger.error("Failed to get top entities for detector " + adTask.getDetectorId(), e);
             internalHCListener.onFailure(e);
         });
+
+        int minimumDocCount = Math.min((int) (bucketInterval / adTask.getDetector().getDetectorIntervalInMilliseconds()), 1);
```
kaituo, line 390: This would choose entities that appear in every detector interval.  Is this what you want?  It sounds too restricted.

kaituo:
> > @ylwu-amzn Are your changes in? Please don't force push as it is hard for reviewers to see the diff. Also, the PR build failed now.
> 
> Thanks for reminding. Pushed changes accidentally. Have fixed the failed build. BTW, you can stop review if the build failed.

Just try to push you so that I can approve the PR.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: Good catch, forgot to add * 0.5 , will tune this

kaituo (dismissed):

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -306,28 +311,96 @@ public void getTopEntities(ADTask adTask, ActionListener<String> internalHCListe
...
+                adTaskCacheManager.addPendingEntities(adTask.getDetectorId(), topNEntities);
+                adTaskCacheManager.setTopEntityCount(adTask.getDetectorId(), topNEntities.size());
+                internalHCListener.onResponse("Get top entities done");
+            }
+        }, e -> {
+            logger.error("Failed to get top entities for detector " + adTask.getDetectorId(), e);
+            internalHCListener.onFailure(e);
+        });
+
+        int minimumDocCount = Math.min((int) (bucketInterval / adTask.getDetector().getDetectorIntervalInMilliseconds()) / 2, 1);
```
ohltyler, line 390: Why have a `Math.min` with a hardcoded `1` value? Wouldn't we always want the min count to be at least 1? This would allow to be 0 if the `bucketInterval` is small enough

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -413,81 +486,84 @@ public void forwardOrExecuteADTask(
...
+                            ADTask adEntityTask = new ADTask.Builder()
+                                .detectorId(adTask.getDetectorId())
+                                .detector(detector)
+                                .isLatest(true)
+                                .taskType(ADTaskType.HISTORICAL_HC_ENTITY.name())
+                                .executionStartTime(now)
+                                .taskProgress(0.0f)
+                                .initProgress(0.0f)
+                                .state(ADTaskState.INIT.name()) // TODO where to set INIT state
+                                .initProgress(0.0f) // TODO where to set INIT state
```
ohltyler, line 543: nit: unsure what these comments mean, can they be removed?

ylwu-amzn: sure, will remove these comments

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: Good catch, should be Math.max.

`bucketInterval` should not be smaller than `detectorInterval`, check line 338 of this class
> Math.max((dataEndTime - dataStartTime) / MAX_TOP_ENTITY_SEARCH_BUCKETS, detectorInterval),

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -1145,6 +1225,30 @@ private void runNextPiece(
...
+                adTaskCacheManager.updateDetectorTaskState(detectorId, task.get().getState());
+                if (adTaskCacheManager.isDetectorTaskStateChanged(detectorId, newState)) {
+                    adTaskManager
+                        .updateADTask(
+                            detectorTaskId,
+                            ImmutableMap.of(STATE_FIELD, newState),
+                            ActionListener
+                                .wrap(
+                                    r -> { adTaskCacheManager.updateDetectorTaskState(detectorId, newState); },
+                                    e -> { logger.debug("Failed to update detector level task " + detectorTaskId, e); }
```
ohltyler, line 1240: If we  error log the exception on line 1245 can we remove this debug line?

ohltyler, line 1240: Should this be `logger.error`?

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -902,4 +905,43 @@ public String pollDeletedDetectorTask() {
...
 
+    /**
+     * Add deleted detector's id to deleted detector queue.
+     * @param detectorId detector id
+     */
+    public void addDeletedDetector(String detectorId) {
+        if (deletedDetectors.size() < maxCachedDeletedTask) {
+            deletedDetectors.add(detectorId);
+        }
+    }
```
ohltyler, line 916: If the queue is full how do we handle an excess amount of detectors-to-be-deleted?

ylwu-amzn: `maxCachedDeletedTask` is 1000 by default (that means each node can have 1000 deleted detector in cache), should be enough to cache deleted detectors. If user delete too many detectors in short time, we will ignore excess deleted detectors, that means user may need to delete these detector's result manually with delete result API.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: sure, will change to logger.error.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: > Wouldn't we always want the min count to be at least 1?

If the bucket interval is very big, for example it includes 1000 detector intervals, then we should not include entities which just has one doc in this bucket interval. From Kaituo, the priority tracker will refresh priority based on current bucket's top entities occurrence or not, ignoring their doc count. So we should not include such "small" entities in our top entities result. And it's hard to find a one-size-fits-all threshold for entity doc count, just use 1/2 interval counts as some easy threshold.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -2212,4 +2254,91 @@ private void resetAllRealtimeTasksAsStopped() {
             );
     }
 
+    public String convertEntityToString(ADTask adTask) {
```
ohltyler, line 2257: Any reason for not creating some of these methods in the `Entity` class's set of methods? I remember Kaituo has some for other inputs to convert to Entity there, like [fromJsonArray](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/model/Entity.java#L388)

ylwu-amzn: This logic is only special to historical task, not general enough for other usage. If somewhere else need to reused, we can move this to Entity. Currently prefer to keep these in task manager to not confuse others.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ohltyler: Got it. `Math.max` makes sense here to raise the threshold, related to # intervals per bucket.

ohltyler (approved):

ohltyler (approved):

PR #162, opened 2021-08-09
Disable model splitting in single-stream detectors

kaituo:
### Description

We split and distribute models to different nodes to avoid large models on a single node. The splitting is unnecessary after introducing compact rcf as the model is smaller (at least 4x smaller). Splitting also undoes the shared point store optimization among trees. Also, splitting brings complications when computing expected values. Thus, this PR disables splitting by increasing the desired model size. We won't split a model whose size is less than the desired size.

This PR adjusts max features and shingle size accordingly to avoid huge models without explicit benefits.

This PR also adjusts the model size formula due to the change https://github.com/aws/random-cut-forest-by-aws/pull/265. I will update the rcf version once rcf 2.0 is released in maven.

Experiments on estimated vs actual model size: https://gist.github.com/kaituo/eeacd12c952363ef3a635ff12674197b

Testing done:
1. tested single-stream models won't be split after the change.
2. Updated unit tests.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/MemoryTracker.java
+++ b/src/main/java/org/opensearch/ad/MemoryTracker.java
@@ -226,7 +226,13 @@ public class MemoryTracker {
      *   + IndexManager
      *     - int array for free indexes: 256 * numberOfTrees * 4, where 4 is the size of an integer
      *   - two int array for locationList and refCount: 256 * numberOfTrees * 4 bytes * 2
-     *   - a float array for data store: 256 * trees * dimension * 4 bytes
+     *   - a float array for data store: 256 * trees * dimension * 4 bytes: due to various
+     *     optimization like shingleSize(dimensions), we don't use all of the array.  The actual
+     *     usage percentage is
+     *     {@code IF(dimensions>=32, 1/(LOG(dimensions+1, 2)+LOG(dimensions+1, 10)), 1/LOG(dimensions+1, 2))}
+     *     where LOG gets the logarithm of a number and the syntax of LOG is {@code LOG (number, [base])}.
+     *     We derive the formula by observing the point store usage ratio is a decreasing function of dimensions
+     *     and the relationship is logarithm. Adding 1 to dimension to ensure dimension 1 results in a ratio 1.
      * - ComponentList: an array of size numberOfTrees
      *   + SamplerPlusTree
      *    - CompactSampler: 2248
@@ -248,7 +254,7 @@ public class MemoryTracker {
      *  56 + # trees * (2248 + 152 + 6120 + 104 + (1040 + 255* (dimension * 4 * 2 + 64)) * adjusted bounding box cache ratio) +
      *  (256 * # trees  * 2 + 256 * # trees * dimension) * 4 bytes  * 0.5 + 1064 + 24 + 24 + 16
      *  = 56 + # trees * (8624 + (1040 + 255 * (dimension * 8 + 64)) * adjusted bounding box cache ratio) + 256 * # trees *
-     *   (2 + dimension) * 4 * 0.5 + 1128
+     *   (3 + dimension) * 4 * 0.5 + 1128
      *
      * @param dimension The number of feature dimensions in RCF
      * @param numberOfTrees The number of trees in RCF
@@ -256,11 +262,28 @@ public class MemoryTracker {
      * @return estimated RCF model size
      */
     public long estimateRCFModelSize(int dimension, int numberOfTrees, double boundingBoxCacheFraction) {
-        float averagePointStoreUsage = dimension == 1 ? 1 : 0.5f;
-        float actualBoundingBoxUsage = boundingBoxCacheFraction >= 0.3 ? 1 : (float) boundingBoxCacheFraction;
+        double averagePointStoreUsage = 0;
+        int logNumber = dimension + 1;
+        if (dimension >= 32) {
+            averagePointStoreUsage = 1.0d / (log2(logNumber) + Math.log10(logNumber));
+        } else {
+            averagePointStoreUsage = 1.0d / log2(logNumber);
+        }
+        double actualBoundingBoxUsage = boundingBoxCacheFraction >= 0.3 ? 1d : boundingBoxCacheFraction;
         long compactRcfSize = (long) (56 + numberOfTrees * (8624 + (1040 + 255 * (dimension * 8 + 64)) * actualBoundingBoxUsage) + 256
-            * numberOfTrees * (2 + dimension) * 4 * averagePointStoreUsage + 1128);
+            * numberOfTrees * (3 + dimension) * 4 * averagePointStoreUsage + 1128);
         return compactRcfSize;
+    }
+
+    /**
+     * Function to calculate the log base 2 of an integer
+     *
+     * @param N input number
+     * @return the base 2 logarithm of an integer value
+     */
+    public static double log2(int N) {
+        // calculate log2 N indirectly using log() method
+        return Math.log(N) / Math.log(2);
     }
 
     public long estimateTotalModelSize(int dimension, int numberOfTrees, double boundingBoxCacheFraction) {
--- a/src/main/java/org/opensearch/ad/MemoryTracker.java
+++ b/src/main/java/org/opensearch/ad/MemoryTracker.java
@@ -37,6 +37,7 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.ad.breaker.ADCircuitBreakerService;
 import org.opensearch.ad.common.exception.LimitExceededException;
 import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.ad.util.MathUtil;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.monitor.jvm.JvmService;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -235,7 +245,9 @@ private AnomalyDetectorSettings() {}
     // TODO (kaituo): change to 4
     public static final int DEFAULT_MULTI_ENTITY_SHINGLE = 1;
 
-    public static final int MAX_SHINGLE_SIZE = 1024;
+    // max shingle size we have seen from external users
+    // the larger shingle size, the harder to fill in a complete shingle
+    public static final int MAX_SHINGLE_SIZE = 60;
```
sudiptoguha, line 250: Makes sense. Does it make sense to bound the maximum dimensions directly to be 60x10 = 600? That is a bit less restrictive and even if people were using 100 features, they're ok as long as shingle size is not above 600/100.

src/test/java/org/opensearch/ad/MemoryTrackerTests.java
```diff
@@ -76,7 +76,7 @@ public void setUp() throws Exception {
         super.setUp();
         rcfNumFeatures = 1;
         rcfSampleSize = 256;
-        numberOfTrees = 10;
+        numberOfTrees = 30;
```
sudiptoguha, line 79: Much needed! An ensemble of 10 trees is quite noisy.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -210,7 +210,17 @@ private AnomalyDetectorSettings() {}
...
+    // The smallest machine in the Amazon managed service has 1GB heap.
+    // With the setting, the desired model size there is of 2 MB.
+    // By default, we can have at most 5 features. Since the default shingle size
+    // is 8, we have at most 40 dimensions in RCF. In our current RCF setting,
+    // 30 trees, and bounding box cache ratio 0, 40 dimensions use 449KB.
+    // Users can increase the number of features to 10 and shingle size to 60,
+    // 30 trees, bounding box cache ratio 0, 600 dimensions use 1.8 MB.
+    // Since these sizes are smaller than the threshold 2 MB, we won't split models
+    // even in the smallest machine.
+    public static final double DESIRED_MODEL_SIZE_PERCENTAGE = 0.002;
```
sudiptoguha, line 223: Much needed. One way to view this is were more cautious when the models were 10x in size. Given that the models are smaller and are unlikely to grow fast, this makes sense. This is only 0.2 %.

src/main/java/org/opensearch/ad/MemoryTracker.java
```diff
@@ -248,21 +254,38 @@ public long estimateTotalModelSize(AnomalyDetector detector, int numberOfTrees,
...
         return compactRcfSize;
     }
 
+    /**
+     * Function to calculate the log base 2 of an integer
+     *
+     * @param N input number
+     * @return the base 2 logarithm of an integer value
+     */
+    public static double log2(int N) {
```
wnbts, line 284: minor. this function log2 is not part of the MemoryTracker interface. it can be moved to some other util class or be private.

wnbts (commented):
does new model size estimation affect handling of existing partitioned models?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: I understand we are more flexible if going this way.  But the change brings two complications:

1. dimension is a rcf concept.  For AD users, it is easier to have two concepts: max features and shingles.
2. the performance impact is unknown.  For example, previously, we query 10 features in one query; now we need to query 600 features in one query.  This needs to be evaluated.

src/test/java/org/opensearch/ad/MemoryTrackerTests.java
kaituo: the change has already made in previous PRs.  This is just tests :)

kaituo:
> does new model size estimation affect handling of existing partitioned models?

for old coordinating node, it will still using old way of estimate model size.  When a new node using compact rcf is involved, old estimation would overestimate memory usage.  For example, for single-stream detectors, old rcf (100 trees, 1 feature, shingle size 8) uses 11_571_200 bytes, while new rcf (30 trees, 1 feature, shingle size 8) uses 1_218_923 bytes. This would reject model allocation we should not have. These inconsistency happens during rolling upgrade or blue green deployment when both old and new nodes exist in the cluster.  It won't impact old cluster or the cluster that has finished upgrading.

src/main/java/org/opensearch/ad/MemoryTracker.java
kaituo: moved to MathUtil.

wnbts:
> happens during rolling upgrade or blue green deployment when both old and new nodes exist in the cluster. It won't impact old cluster or the cluster that has finished upgrading.



> > does new model size estimation affect handling of existing partitioned models?
> 
> for old coordinating node, it will still using old way of estimate model size. When a new node using compact rcf is involved, old estimation would overestimate memory usage. For example, for single-stream detectors, old rcf (100 trees, 1 feature, shingle size 8) uses 11_571_200 bytes, while new rcf (30 trees, 1 feature, shingle size 8) uses 1_218_923 bytes. This would reject model allocation we should not have. These inconsistency happens during rolling upgrade or blue green deployment when both old and new nodes exist in the cluster. It won't impact old cluster or the cluster that has finished upgrading.

Will new nodes be able to find all old partitioned models? For example, a detector has 5 partitioned rcf now and new nodes will compute to have 1. Will the other four be found?

kaituo:
> > > does new model size estimation affect handling of existing partitioned models?
> > 
> > 
> > for old coordinating node, it will still using old way of estimate model size. When a new node using compact rcf is involved, old estimation would overestimate memory usage. For example, for single-stream detectors, old rcf (100 trees, 1 feature, shingle size 8) uses 11_571_200 bytes, while new rcf (30 trees, 1 feature, shingle size 8) uses 1_218_923 bytes. This would reject model allocation we should not have. These inconsistency happens during rolling upgrade or blue green deployment when both old and new nodes exist in the cluster. It won't impact old cluster or the cluster that has finished upgrading.
> 
> Will new nodes be able to find all old partitioned models? For example, a detector has 5 partitioned rcf now and new nodes will compute to have 1. Will the other four be found?

The other four won't be found.  The new node will load partition 0 from checkpoints, convert to rcf 2.0, and run from there.  Any easy way to fix this?

wnbts:
> The other four won't be found. The new node will load partition 0 from checkpoints, convert to rcf 2.0, and run from there. Any easy way to fix this?

There could be several ways to fix/mitigate, in no particular order.
1. During loading, if rcf version is 1.0, keep getting more partitions to merge.
2. Use the checkpoint but update its config during deserialization. For example, increase number of trees to the desired.

And it would be good to think about how to handle all possible cases. For example, when new computed partition number is larger than the current. And maybe a catch-all solution such as retrain a model if all fail/are too hard/expensive to recover.

wnbts (approved):

kaituo:
> > The other four won't be found. The new node will load partition 0 from checkpoints, convert to rcf 2.0, and run from there. Any easy way to fix this?
> 
> There could be several ways to fix/mitigate, in no particular order.
> 
>     1. During loading, if rcf version is 1.0, keep getting more partitions to merge.

This requires RCF support merging multiple multiple forests and combine them into one forest.

>     2. Use the checkpoint but update its config during deserialization. For example, increase number of trees to the desired.

This requires RCF dynamically changing the number of trees.

Both suggestions are great.  Does RCF support these operations?

> And it would be good to think about how to handle all possible cases. For example, when new computed partition number is larger than the current. And maybe a catch-all solution such as retrain a model if all fail/are too hard/expensive to recover.

Yes, will have a doc to list these backward compatibility cases.  The new computed partition number won't be larger as the new number is always 1.

wnbts:
> Both suggestions are great. Does RCF support these operations?

For merging, some work is needed to merge the samples in states and deserialize the merged state to model.

For setting new forest size, it should be just set the new number of trees on the state and deserialize modified state to model.

kaituo:
> > Both suggestions are great. Does RCF support these operations?
> 
> For merging, some work is needed to merge the samples in states and deserialize the merged state to model.
> 
> For setting new forest size, it should be just set the new number of trees on the state and deserialize modified state to model.

Got it.  The 2nd one requires changes on Josh's converter code.

sudiptoguha:
> > > Both suggestions are great. Does RCF support these operations?
> > 
> > 
> > For merging, some work is needed to merge the samples in states and deserialize the merged state to model.
> > For setting new forest size, it should be just set the new number of trees on the state and deserialize modified state to model.
> 
> Got it. The 2nd one requires changes on Josh's converter code.

Both require some work (unless we are decreasing the number of trees). Neither are very difficult -- let me think for a bit.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
sudiptoguha: I think for the first the explanation is easy. I think some shrive like Kinesis sets the product to be 900 (or something similar). Basically, if we want large models we can be OOM -- that should be easy to explain.

For 2, If the user creates a 600 features in one query then that may be a bigger issue independent of AD :)

This seems like a two way door - if use cases come up, the change can be done easily.

sudiptoguha (approved):

kaituo:
> > > > Both suggestions are great. Does RCF support these operations?
> > > 
> > > 
> > > For merging, some work is needed to merge the samples in states and deserialize the merged state to model.
> > > For setting new forest size, it should be just set the new number of trees on the state and deserialize modified state to model.
> > 
> > 
> > Got it. The 2nd one requires changes on Josh's converter code.
> 
> Both require some work (unless we are decreasing the number of trees). Neither are very difficult -- let me think for a bit.

I like the 2nd option better (changing converted code to provide an option to add the number of trees).  It is easier on both ends (AD and RCF).

kaituo:
@sudiptoguha Agree it is two way door.  Let's see if there is any customer needs.

wnbts:
the converted state has setter method for number of trees

kaituo:
> the converted state has setter method for number of trees

@sudiptoguha will prepare methods to merge multiple forests into one and adjust tree number (both increasing or decreasing).

PR #163, opened 2021-08-09
extended rcf with thresholding

wnbts:
### Description
This is the initial pr for adding a new anomaly detection model that extends the current rcf with thresholding, named Extended Random Cut Forest (ERCF). For the sake of time and unblocking further development, the initial unit tests only include those for the primary features/api to be used in this plugin. Implementation changes, testing, and documentation will ensue in following pr for this plugin or the rcf package.
 
### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ out/
 .vscode
 bin/
 src/test/resources/job-scheduler
+lombok.config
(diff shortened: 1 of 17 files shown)


kaituo:
As discussed offline, we will use the code from RCF directly.  The dependency doesn't have to be from maven.  It can be a binary file dependency.

PR #166, opened 2021-08-12
add ercf to entitymodel

wnbts:
### Description
This pr adds the new extended rcf model to entitymodel. Anomaly detection models, for single-entity and for high-cardinality, will be managed as entity models.
  
### Check List


kaituo (approved):

src/main/java/org/opensearch/ad/ml/EntityModel.java
```diff
@@ -47,6 +49,19 @@ public EntityModel(Entity entity, Queue<double[]> samples, RandomCutForest rcf,
...
+     *
+     * @param entity entity if any
+     * @param samples samples with the model
+     * @param ercf extended rcf model
+     */
+    public EntityModel(Entity entity, Queue<double[]> samples, ExtendedRandomCutForest ercf) {
+        this.entity = entity;
+        this.samples = samples;
+        this.ercf = ercf;
+    }
```
cswiercz, line 63: Are `this.rcf` and `this.threshold` initialized to `null`? Since ERCF = RCF + Thresholder it's important that this `EntityModel` doesn't have instances of all three types.

cswiercz (approved):

src/main/java/org/opensearch/ad/ml/EntityModel.java
wnbts: yes, default values are null.

PR #169, opened 2021-08-12
get thresholding result from ercf

wnbts:
### Description
This pr uses ercf to get anomaly result.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/ModelManager.java
+++ b/src/main/java/org/opensearch/ad/ml/ModelManager.java
@@ -62,6 +62,7 @@ import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 
+import com.amazon.randomcutforest.ERCF.AnomalyDescriptor;
 import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.config.Precision;
 import com.amazon.randomcutforest.returntypes.DiVector;
@@ -809,6 +810,7 @@ public class ModelManager implements DetectorModelSize {
         AnomalyDetector detector,
         Entity entity
     ) {
+        ThresholdingResult result = new ThresholdingResult(0, 0, 0);
         if (modelState != null) {
             EntityModel entityModel = modelState.getModel();
 
@@ -817,20 +819,22 @@ public class ModelManager implements DetectorModelSize {
                 modelState.setModel(entityModel);
             }
 
-            // trainModelFromExistingSamples may be able to make models not null
-            if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
-                entityColdStarter.trainModelFromExistingSamples(modelState);
-            }
-
-            if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
-                return score(datapoint, modelId, modelState);
+            if (entityModel.getErcf().isPresent()) {
+                result = toResult(entityModel.getErcf().get().process(datapoint));
             } else {
-                entityModel.addSample(datapoint);
-                return new ThresholdingResult(0, 0, 0);
+                // trainModelFromExistingSamples may be able to make models not null
+                if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
+                    entityColdStarter.trainModelFromExistingSamples(modelState);
+                }
+
+                if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
+                    return score(datapoint, modelId, modelState);
+                } else {
+                    entityModel.addSample(datapoint);
+                }
             }
-        } else {
-            return new ThresholdingResult(0, 0, 0);
         }
+        return result;
     }
 
     public ThresholdingResult score(double[] feature, String modelId, ModelState<EntityModel> modelState) {
@@ -918,5 +922,9 @@ public class ModelManager implements DetectorModelSize {
         while (samples.peek() != null) {
             toModel.addSample(samples.poll());
         }
+    }
+
+    private ThresholdingResult toResult(AnomalyDescriptor anomalyDescriptor) {
+        return new ThresholdingResult(anomalyDescriptor.getAnomalyGrade(), /*TODO: pending ercf*/1.0, anomalyDescriptor.getRcfScore());
     }
 }
--- a/src/test/java/org/opensearch/ad/ml/ModelManagerTests.java
+++ b/src/test/java/org/opensearch/ad/ml/ModelManagerTests.java
@@ -91,6 +91,8 @@ import org.opensearch.threadpool.ThreadPool;
 import org.powermock.modules.junit4.PowerMockRunner;
 import org.powermock.modules.junit4.PowerMockRunnerDelegate;
 
+import com.amazon.randomcutforest.ERCF.AnomalyDescriptor;
+import com.amazon.randomcutforest.ERCF.ExtendedRandomCutForest;
 import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.returntypes.DiVector;
 import com.google.common.collect.Sets;
@@ -125,6 +127,15 @@ public class ModelManagerTests {
 
     @Mock
     private EntityCache cache;
+
+    @Mock
+    private ModelState<EntityModel> modelState;
+
+    @Mock
+    private EntityModel entityModel;
+
+    @Mock
+    private ExtendedRandomCutForest ercf;
 
     private double modelDesiredSizePercentage;
     private double modelMaxSizePercentage;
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -817,20 +819,25 @@ public ThresholdingResult getAnomalyResultForEntity(
...
 
-            // trainModelFromExistingSamples may be able to make models not null
-            if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
-                entityColdStarter.trainModelFromExistingSamples(modelState);
-            }
-
-            if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
-                return score(datapoint, modelId, modelState);
+            if (entityModel.getErcf().isPresent()) {
+                result = toResult(entityModel.getErcf().get().process(datapoint));
```
cswiercz, line 823: Design nitpick: the original code (below) uses early returns. The new code here is saying that if an ERCF exists then we will prefer it's output as opposed to the "original model" output. Why not do an early return here? It's allowed since this is the only line that sets `result`. It will decrease the cognitive load, a bit.

wnbts: * yes, ercf will be the current/only model in the system.
* multiple returns are considered less clear than a single return so the new code switches to that. and old code for old models will be removed because with later changes, old models will be converted to ercf before use.

cswiercz: Sounds good.

cswiercz (approved):

src/test/java/org/opensearch/ad/ml/ModelManagerTests.java
```diff
@@ -266,6 +277,9 @@ public void setup() {
         detectorId = "detectorId";
         rcfModelId = "detectorId_model_rcf_1";
         thresholdModelId = "detectorId_model_threshold";
+
+        when(this.modelState.getModel()).thenReturn(this.entityModel);
+        when(this.entityModel.getErcf()).thenReturn(Optional.of(this.ercf));
```
kaituo, line 282: This would make it hard to test the else branch of entityModel.getErcf().isPresent() == false.  Can we move it to somewhere else?

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -817,20 +819,25 @@ public ThresholdingResult getAnomalyResultForEntity(
...
+                // trainModelFromExistingSamples may be able to make models not null
+                if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
+                    entityColdStarter.trainModelFromExistingSamples(modelState);
+                }
+
+                if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
+                    return score(datapoint, modelId, modelState);
+                } else {
+                    entityModel.addSample(datapoint);
+                    return new ThresholdingResult(0, 0, 0);
```
kaituo, line 834: Seems line 834 and 838 can be removed as those are the last line of the branches and they are doing the same thing of line 840.

wnbts: yes, they can be removed.

src/test/java/org/opensearch/ad/ml/ModelManagerTests.java
wnbts: it shouldn't not affect new test cases since new test cases can always override this behavior so this is just a default for convenience.

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -817,20 +819,22 @@ public ThresholdingResult getAnomalyResultForEntity(
...
+                // trainModelFromExistingSamples may be able to make models not null
+                if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
+                    entityColdStarter.trainModelFromExistingSamples(modelState);
+                }
+
+                if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
+                    return score(datapoint, modelId, modelState);
+                } else {
+                    entityModel.addSample(datapoint);
+                }
```
LiuJoyceC, line 834: [Clarification question] Just to confirm, does invoking `entityColdStarter.trainModelFromExistingSamples(modelState)` synchronously change the result of `entityModel.getRcf()` and/or `entityModel.getThreshold()`? Otherwise the second if statement is redundant.

[Minor suggestion, non-blocking] Since this set of states is being checked twice (just in opposite cases), it would be more elegant and easier to read to add a method on `entityModel` that checks this exact condition. Then you could have:
```
if (entityModel.hasRcfOrThreshold()) { ... }
if (!entityModel.hasRcfOrThreshold() { ... } else { ... }
```

src/test/java/org/opensearch/ad/ml/ModelManagerTests.java
```diff
@@ -973,4 +987,19 @@ public void getPreviewResults_returnAnomalies_forLastAnomaly() {
...
+        anomalyDescriptor.setAnomalyGrade(1);
+        when(this.ercf.process(this.point)).thenReturn(anomalyDescriptor);
+
+        ThresholdingResult result = modelManager
+            .getAnomalyResultForEntity(this.point, this.modelState, this.detectorId, this.anomalyDetector, null);
+        assertEquals(
+            new ThresholdingResult(anomalyDescriptor.getAnomalyGrade(), /*TODO: pending ercf*/1.0, anomalyDescriptor.getRcfScore()),
+            result
+        );
+    }
```
LiuJoyceC, line 1004: Instead of hardcoding the value 2 for the RCF score and 1 for the anomaly grade, this test function could accept arguments for the RCF score and anomaly grade. Then you can test a few different combinations of values, including edge cases where an unexpected behavior could happen.

LiuJoyceC: What is `/*TODO: pending ercf*/`? Is that waiting on another PR?

src/main/java/org/opensearch/ad/ml/ModelManager.java
wnbts: 1. Both should be set or unset. But it's fine and the code for old models will be removed soon as it is replaced by new ercf.
2. As above, this branch will be left as it is and soon be replaced by new logic based on ercf.

src/test/java/org/opensearch/ad/ml/ModelManagerTests.java
wnbts: 1. Parameterized testing is useful. In this case, the logic is quite simple as just copying data so the test can therefore be simple. If the logic becomes more complex, the corresponding tests can be parameterized to scale to the test cases. However, that is not always needed.
2. The comment is a TODO item for the confidence part of ercf is still in development and not available yet. It unblocks development and integration on the plugin side for the time being.

kaituo (approved):

src/main/java/org/opensearch/ad/ml/ModelManager.java
ylwu-amzn: 1. Can the `process` method support batch data? Asking this as we have historical analysis which run in batch. 
2. Do you have any suggestion about how many data points should be good enough for ERCF to generate results with high confidence.

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -817,20 +819,22 @@ public ThresholdingResult getAnomalyResultForEntity(
...
-
-            if (entityModel.getRcf() != null && entityModel.getThreshold() != null) {
-                return score(datapoint, modelId, modelState);
+            if (entityModel.getErcf().isPresent()) {
+                result = toResult(entityModel.getErcf().get().process(datapoint));
             } else {
-                entityModel.addSample(datapoint);
-                return new ThresholdingResult(0, 0, 0);
+                // trainModelFromExistingSamples may be able to make models not null
+                if (entityModel.getRcf() == null || entityModel.getThreshold() == null) {
```
ylwu-amzn, line 826: Should we remove such RCF and Threshold things as ERCF hide these behind the scene ?

src/main/java/org/opensearch/ad/ml/ModelManager.java
wnbts: 1. No, it only takes one point at a time. For a batch of data points, the client code calls on each of them.
2. Not new numbers yet. The current number should be sufficient.

src/main/java/org/opensearch/ad/ml/ModelManager.java
wnbts: Yes, eventually. This branch dealing with existing models will be replaced with new pr.

LiuJoyceC (approved):

PR #170, opened 2021-08-12
support backward compatibility of historical analysis and realtime task

ylwu-amzn:
### Description

This PR contains mainly 3 changes:

## Data migration

In unified flow, we added realtime task for AD job to use same way to track state and error  as historical analysis, and frontend can search realtime tasks to avoid calling profile API for multiple times on detector list page. We need to backfill realtime tasks for old AD job to support BWC. If there is any AD node running AD version greater than 1.0, we will choose one owning node with highest AD version to backfill data. 

It's possible that some AD job started after migration job starts, for this case the job runner will check if realtime task exists and create a new realtime task if not.

## AD version based routing

We plan to deprecate historical detector and move detection date range from historical detector to AD task. So user can create a general detector and use it to run both realtime and historical anomaly detection. This data model change need BWC support as well.

To support historical analysis for HC detector (both single-category and multi-category) in next release, we added new task actions for HC detector like "NEXT_ENTITY" and we changed the interface on transport layer like `ForwardADTaskRequest`. To make BWC easier, this PR added AD version hash ring and only dispatch tasks to nodes with the same local AD version. As AD 1.0 has no such routing logic, the old AD node may still send request to new AD node in mixed cluster. As the mixed cluster should be short period during rolling upgrade, the new AD node just simply rejects the requests from old AD node. 

We should make sure not running a new task if a detector is already running on old AD nodes. So tune the task profile transport interface to make sure it can work on different AD version. Mainly tune the request/response serialization/deserialization logic.

And in this way, it makes the BWC support&test easier that we can build bump minor version of AD and deploy it to some data nodes to test BWC rather than build a new version of OpenSearch.

## Auto tune task slots for HC detector historical analysis

HC detector historical analysis will run multiple entities in parallel on different worker nodes. We abstract the entity run with "task lane". Each entity task will run in one "task lane". When entity task done, it will send "entity task done" message to coordinating node to poll next entity. It's ok for one HC detector running in cluster. But if multiple HC detector running, other HC detectors may steal "task lanes" and in worst case some HC detector will lose all task lanes and fail.

To solve this, this PR adds a new node type "Lead Node". Now we will have three types of nodes: "Lead Node", "Coordinating Node" and "Worker Node". When a start historical analysis request arrives, will find "Lead Node" with AD version hash ring and constant model id. "Lead Node" will gather all assigned detector slots from all "Coordinating Node" and all running tasks on "Worker Node". Then "Lead Node" will calculate how many task slots the cluster can run totally `total_task_slots=eligible_data_nodes * max_tasks_per_node`, then we can get available task slots as `available_task_slots=total_task_slots - max(assgined_task_slots, running_tasks)`.   (**Note**: We need to use `max(assgined_task_slots, running_tasks)` to cover edge case that coordinating node left and we can't get assigned task slots on it). We can calculate how many task slots the detector needs based on `assigned_task_slots=min(max_running_entities_per_detector, available_task_slots, entity_count_in_detection_date_range)`. The assigned task slots will be stored on "Coordinating Node", "Lead Node" will not cache anything, it just calculate how many task slots available and how many tasks slots should be assigned to detector.

The cluster is dynamic, so we should tune task slots dynamically to fit the cluster. For example, when a node joins/left or user tune AD dynamic setting like how many entities can run for one HC detector. When each entity task done, we will check if we should scale task slots.

 
 
### Check List


kaituo:
This change is not what we discussed/agreed last time.  Suggest we have discussions first before the PR.

ylwu-amzn:
> This change is not what we discussed/agreed last time. Suggest we have discussions first before the PR.

Yeah, that's why I send it out as draft for discussion. Based on our short offline discussion today, this AD version hash ring and dispatch historical task only to nodes with same AD version looks reasonable. I will do more test and cleanup this PR, but feel free to comment on the draft version. So we can gather feedback earlier to save effort.

kaituo (commented):
publish comments earlier.  Haven't checked new commits you sent today so my comments might not be relevant any more.  I only finished 20/50 files.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -632,6 +634,34 @@ private void indexAnomalyResultException(
...
+                    .info(
+                        "Updated latest realtime task successfully for detector {}, taskState: {},"
+                            + " RCF total update: {}, detectorIntervalInMinutes: {}",
+                        detectorId,
+                        taskState,
+                        rcfTotalUpdates,
+                        detectorIntervalInMinutes
+                    );
+            }, e -> {
+                if ((e instanceof ResourceNotFoundException) && e.getMessage().contains("can't find latest task")) {
```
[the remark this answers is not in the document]
ylwu-amzn: Sure, fixed

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+
+            if (!detectionIndices.doesAnomalyDetectorJobIndexExist()) {
+                logger.info("AD job index doesn't exist, no need to migrate");
+                return;
+            }
+
+            if (detectionIndices.doesDetectorStateIndexExist()) {
+                migrateDetectorInternalStateToRealtimeTask();
+            } else {
+                // If detection index doesn't exist, create index and execute historical detector.
```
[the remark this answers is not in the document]
ylwu-amzn: Wrong comments, fixed

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -141,4 +190,265 @@ public boolean build() {
     public void recordMembershipChange() {
         membershipChangeRequied.set(true);
     }
+
+    /**
+     * Build AD version based circles with discovery node delta change. Listen to master event in
+     * {@link ADClusterEventListener#clusterChanged(ClusterChangedEvent)}.
+     * Will remove the removed nodes from cache and send request to newly added nodes to get their plugin information,and.
```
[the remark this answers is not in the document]
ylwu-amzn: fixed

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -141,4 +190,265 @@ public boolean build() {
...
+                adVersionHashRingInited.set(true);
+                actionListener.onResponse(true);
+            }, e -> {
+                adVersionCircleInProgress.release();
+                actionListener.onFailure(e);
+                LOG.error("Fail to get node info to build AD version hash ring", e);
+            }));
+        } catch (Exception e) {
+            LOG.error("Failed to build AD version circles", e);
+            adVersionCircleInProgress.release();
```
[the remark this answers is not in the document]
ylwu-amzn: Good catch, fixed

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -141,4 +190,265 @@ public boolean build() {
...
+                        DiscoveryNode curNode = nodeInfo.getNode();
+                        if (!nodeFilter.isEligibleDataNode(curNode)) {
+                            continue;
+                        }
+                        TreeMap<Integer, DiscoveryNode> circle = null;
+                        for (PluginInfo pluginInfo : plugins.getPluginInfos()) {
+                            if (AD_PLUGIN_NAME.equals(pluginInfo.getName()) || AD_PLUGIN_NAME_FOR_TEST.equals(pluginInfo.getName())) {
+                                Version version = ADVersionUtil.fromString(pluginInfo.getVersion());
+                                circle = adVersionCircles.computeIfAbsent(version, key -> new TreeMap<>());
+                                nodeAdVersions.computeIfAbsent(curNode.getId(), key -> ADVersionUtil.fromString(pluginInfo.getVersion()));
```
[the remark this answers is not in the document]
ylwu-amzn: fixed

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -141,4 +190,265 @@ public boolean build() {
...
+        Map.Entry<Version, TreeMap<Integer, DiscoveryNode>> versionTreeMapEntry = adVersionCircles.lastEntry();
+        if (versionTreeMapEntry == null) {
+            return Optional.empty();
+        }
+        TreeMap<Integer, DiscoveryNode> adVersionCircle = versionTreeMapEntry.getValue();
+        Map.Entry<Integer, DiscoveryNode> entry = adVersionCircle.higherEntry(modelHash);
+        return Optional.ofNullable(Optional.ofNullable(entry).orElse(adVersionCircle.firstEntry())).map(x -> x.getValue());
+    }
+
+    /**
```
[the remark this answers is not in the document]
ylwu-amzn: fixed

src/main/java/org/opensearch/ad/transport/CronRequest.java
```diff
@@ -61,12 +60,4 @@ public CronRequest(String requestId, DiscoveryNode... nodes) {
     public String getRequestId() {
         return this.requestId;
     }
-
```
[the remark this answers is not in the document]
ylwu-amzn: Plan to remove request id to make it compatible.

src/main/java/org/opensearch/ad/transport/CronNodeRequest.java
```diff
@@ -60,8 +60,5 @@ public String getRequestId() {
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
-        if (Bwc.supportMultiCategoryFields(out.getVersion())) {
```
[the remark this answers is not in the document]
ylwu-amzn: Plan to remove request id to make it compatible.

src/main/java/org/opensearch/ad/cluster/ADVersionUtil.java
```diff
@@ -0,0 +1,52 @@
...
+        StringBuilder normalizedVersion = new StringBuilder();
+        normalizedVersion.append(versions[0]);
+        for (int i = 1; i < VERSION_SEGMENTS; i++) {
+            normalizedVersion.append(".");
+            normalizedVersion.append(versions[i]);
+        }
+        return normalizedVersion.toString();
+    }
+
+    public static boolean versionCompatible(Version adVersion) {
```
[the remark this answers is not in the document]
ylwu-amzn: sure, this is to check if current local version is compatible with the `adVersion` passed in.

src/main/java/org/opensearch/ad/transport/ForwardADTaskRequest.java
```diff
@@ -48,19 +51,46 @@
...
+    private Integer availableTaskSLots;
     private ADTaskAction adTaskAction;
 
-    public ForwardADTaskRequest(AnomalyDetector detector, DetectionDateRange detectionDateRange, User user, ADTaskAction adTaskAction) {
+    public ForwardADTaskRequest(
+        AnomalyDetector detector,
+        DetectionDateRange detectionDateRange,
+        User user,
+        ADTaskAction adTaskAction,
+        Version remoteAdVersion
```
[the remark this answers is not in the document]
ylwu-amzn: Lead node has the same AD version of local node, so just use local AD version as remote AD version. Will create a new constructor function without remoteAdVersion to make it clear.

src/main/java/org/opensearch/ad/transport/ForwardADTaskRequest.java
```diff
@@ -48,19 +51,46 @@
...
+        ADTaskAction adTaskAction,
+        Integer availableTaskSLots,
+        Version remoteAdVersion
+    ) {
         this.detector = detector;
         this.detectionDateRange = detectionDateRange;
         this.user = user;
+        this.availableTaskSLots = availableTaskSLots;
         this.adTaskAction = adTaskAction;
+        if (!ADVersionUtil.versionCompatible(remoteAdVersion)) {
```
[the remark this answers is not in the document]
ylwu-amzn: sure, fixed

src/main/java/org/opensearch/ad/model/ADTaskType.java
```diff
@@ -27,27 +27,34 @@
...
 public enum ADTaskType {
-    REALTIME_SINGLE_ENTITY,
+    @Deprecated
+    HISTORICAL,
+    REALTIME_SINGLE_FLOW,
     REALTIME_HC_DETECTOR,
-    HISTORICAL_SINGLE_ENTITY,
+    HISTORICAL_SINGLE_FLOW,
     HISTORICAL_HC_DETECTOR,
     HISTORICAL_HC_ENTITY;
```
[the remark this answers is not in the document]
ylwu-amzn: `HISTORICAL_HC_DETECTOR`: detector level task to track overall state, init progress, error etc. for HC detector
`HISTORICAL_HC_ENTITY` : entity level task to track just one specific entity's state, init progress, error etc.

src/main/java/org/opensearch/ad/model/ADTaskType.java
```diff
@@ -27,27 +27,34 @@
...
 import com.google.common.collect.ImmutableList;
 
 public enum ADTaskType {
-    REALTIME_SINGLE_ENTITY,
+    @Deprecated
+    HISTORICAL,
+    REALTIME_SINGLE_FLOW,
     REALTIME_HC_DETECTOR,
-    HISTORICAL_SINGLE_ENTITY,
+    HISTORICAL_SINGLE_FLOW,
```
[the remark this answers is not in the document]
ylwu-amzn: fixed

src/main/java/org/opensearch/ad/model/ADEntityTaskProfile.java
```diff
@@ -0,0 +1,287 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
ylwu-amzn: sure, removed

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -141,4 +190,265 @@ public boolean build() {
...
+     * Get AD version.
+     * @param nodeId node id
+     * @return AD version
+     */
+    public Version getAdVersion(String nodeId) {
+        return nodeAdVersions.get(nodeId);
+    }
+
+    /**
+     * Get all eligible data nodes which AD version known in AD version based hash ring.
```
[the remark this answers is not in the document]
ylwu-amzn: fixed

src/main/java/org/opensearch/ad/model/ADTaskProfile.java
```diff
@@ -70,86 +71,34 @@
     private Integer thresholdModelTrainingDataSize;
     private Long modelSizeInBytes;
     private String nodeId;
-    private Entity entity;
```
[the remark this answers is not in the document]
ylwu-amzn: `Entity` has been moved to `ADEntityTaskProfile`. One example
```
{
  "ad_task" : {
    "ad_task" : {
      "task_id" : "Y6PDdXsBt5oPIND7s995",
      "last_update_time" : 1629768367742,
      "state" : "RUNNING",
      "detector_id" : "b_e-S3sB16-OL5YLC88d",
      "task_progress" : 0.0,
      "init_progress" : 0.0,
      "execution_start_time" : 1629768364920,
      "is_latest" : true,
      "task_type" : "HISTORICAL_HC_DETECTOR",
      "coordinating_node" : "ZqF1rBmzTF6hRqSOT_NU8w",
      "detector" : {
        "name" : "hc3",
        "description" : "nab_ec2_cpu_utilization_24ae8d",
        "time_field" : "timestamp",
        "indices" : [
          "nab_ec2_cpu_utilization_24ae8d"
        ],
        "filter_query" : {
          "match_all" : {
            "boost" : 1.0
          }
        },
        "detection_interval" : {
          "period" : {
            "interval" : 1,
            "unit" : "Minutes"
          }
        },
        "window_delay" : {
          "period" : {
            "interval" : 1,
            "unit" : "Minutes"
          }
        },
        "shingle_size" : 1,
        "schema_version" : 0,
        "feature_attributes" : [
          {
            "feature_id" : "bve-S3sB16-OL5YLC88E",
            "feature_name" : "F1",
            "feature_enabled" : true,
            "aggregation_query" : {
              "f_1" : {
                "sum" : {
                  "field" : "value"
                }
              }
            }
          }
        ],
        "last_update_time" : 1629063351068,
        "category_field" : [
          "type"
        ]
      },
      "detection_date_range" : {
        "start_time" : 1628450905000,
        "end_time" : 1629055705000
      }
    },
    "node_id" : "ZqF1rBmzTF6hRqSOT_NU8w",
    "task_id" : "Y6PDdXsBt5oPIND7s995",
    "task_type" : "HISTORICAL_HC_DETECTOR",
    "detector_task_slots" : 2,
    "total_entities_count" : 20,
    "pending_entities_count" : 18,
    "running_entities_count" : 2,
    "running_entities" : [
      "error9",
      "error8"
    ],
    "entity_task_profiles" : [
      {
        "shingle_size" : 1,
        "rcf_total_updates" : 3000,
        "threshold_model_trained" : true,
        "threshold_model_training_data_size" : 0,
        "model_size_in_bytes" : 1144784,
        "node_id" : "qBJK7rjcQFWRy6qqFY3ZmA",
        "entity" : [
          {
            "name" : "type",
            "value" : "error9"
          }
        ],
        "task_id" : "ZaPDdXsBt5oPIND7vN9J",
        "task_type" : "HISTORICAL_HC_ENTITY"
      },
      {
        "shingle_size" : 1,
        "rcf_total_updates" : 3000,
        "threshold_model_trained" : true,
        "threshold_model_training_data_size" : 0,
        "model_size_in_bytes" : 1144784,
        "node_id" : "UYW7pC60TMaQuoMZoLw-Iw",
        "entity" : [
          {
            "name" : "type",
            "value" : "error8"
          }
        ],
        "task_id" : "ZqPDdXsBt5oPIND7vN90",
        "task_type" : "HISTORICAL_HC_ENTITY"
      }
    ]
  }
}

```

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -141,4 +190,265 @@ public boolean build() {
...
+     * @param function consumer function
+     * @param listener action listener
+     * @param <T> listener response type
+     */
+    public <T> void getOwningNodeWithSameLocalAdVersion(
+        String modelId,
+        Consumer<Optional<DiscoveryNode>> function,
+        ActionListener<T> listener
+    ) {
+        buildCirclesOnAdVersions(ActionListener.wrap(r -> {
```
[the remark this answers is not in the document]
ylwu-amzn: You mean this function `buildCirclesOnAdVersions(ActionListener<Boolean> actionListener)` ? This function will not return owning node, it just builds AD version hash rings. Will pass `false` to `actionListener.onResponse` if fail to acquire AD version hash ring updating semaphore.

src/main/java/org/opensearch/ad/transport/CronNodeRequest.java
```diff
@@ -48,7 +47,8 @@ public CronNodeRequest(String requestId) {
 
     public CronNodeRequest(StreamInput in) throws IOException {
         super(in);
-        if (Bwc.supportMultiCategoryFields(in.getVersion())) {
+        // The request id is added since AD 1.1
```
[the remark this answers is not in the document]
ylwu-amzn: We can use different logic to write response to output stream based on remote node version. Check `ADTaskProfileTransportAction#nodeOperation` and `RCFResultTransportAction#doExecute`

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -528,25 +531,25 @@ private void updateRealtimeTask(AnomalyResultResponse response, String detectorI
...
+                updateLatestRealtimeTask(
+                    jobParameter,
                     detectorId,
                     null,
-                    response.getRcfTotalUpdates(),
+                    r.getTotalUpdates(),
                     response.getDetectorIntervalInMinutes(),
                     response.getError()
                 );
+            }, e -> { log.error("Failed to get init progress profile for " + detectorId, e); }));
```
ohltyler, line 542: error message doesn't make sense here? Maybe change to something like `Failed to update latest realtime task for " + detectorId, e`

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+                try (XContentParser parser = createXContentParserFromRegistry(xContentRegistry, searchHit.getSourceRef())) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                    AnomalyDetectorJob job = AnomalyDetectorJob.parse(parser);
+                    detectorJobs.add(job);
+                } catch (IOException e) {
+                    logger.error("Fail to parse AD job " + searchHit.getId(), e);
+                }
+            }
+            logger.info("Total AD jobs to backfill realtime task: {}", detectorJobs.size());
+            backfillRealtimeTask(detectorJobs, true);
```
ohltyler, line 151: Can you explain why there would be multiple jobs, and what the reason is to migrate multiple over, rather than just the latest?

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
+            .query(new MatchAllQueryBuilder())
+            .size(MAX_DETECTOR_UPPER_LIMIT);
+        SearchRequest searchRequest = new SearchRequest(ANOMALY_DETECTOR_JOB_INDEX).source(searchSourceBuilder);
+        client.search(searchRequest, ActionListener.wrap(r -> {
+            if (r == null || r.getHits().getTotalHits() == null || r.getHits().getTotalHits().value == 0) {
+                logger.info("No anomaly detector job found, no need to migrate");
+                return;
+            }
+            ConcurrentLinkedQueue<AnomalyDetectorJob> detectorJobs = new ConcurrentLinkedQueue<>();
```
ohltyler, line 138: Can you explain how/why there would be multiple jobs for a detector? And if so, why migrate all of them, instead of just the latest?

src/main/java/org/opensearch/ad/model/DetectorProfile.java
```diff
@@ -50,7 +49,7 @@
     private InitProgressProfile initProgress;
     private Long totalEntities;
     private Long activeEntities;
-    private Map<String, ADTaskProfile> adTaskProfiles;
+    private ADTaskProfile adTaskProfile;
```
ohltyler, line 52: Can you explain why multiple profiles were passed before?

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -577,9 +612,142 @@ public Integer getTopEntityCount(String detectorId) {
...
+            .computeIfAbsent(detectorId, key -> new ADTaskSlotLimit(taskSlots, taskSlots));
+        adTaskSlotLimit.setDetectorTaskSlots(taskSlots);
+    }
+
+    /**
+     * Scale up detector task slots.
+     * @param detectorId detector id
+     * @param delta scale delta
+     */
+    public synchronized void scaleUpDetectorTaskSlots(String detectorId, int delta) {
```
ohltyler, line 639: Maybe add some error handling or logging for failure, if the scaling up failed? Same goes for `scaleDownHCDetectorTaskSlots`

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -577,9 +612,142 @@ public Integer getTopEntityCount(String detectorId) {
...
+
+    /**
+     * Get total task slots on this node.
+     * @return total task slots
+     */
+    public int getTotalDetectorTaskSlots() {
+        int totalTaskSLots = 0;
+        logger
+            .info(
+                "yyyyyyyyyyyyyyyyyyyyyyyyyyyy detectorTaskSlotLimit size: {}, {}",
```
ohltyler, line 737: Clean this log message

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -606,21 +783,47 @@ private ADHCBatchTaskCache getExistingHCTaskCache(String detectorId) {
...
 
-    private ADHCBatchTaskCache getHCTaskCache(String detectorId) {
+    private ADHCBatchTaskCache getOrCreateHCTaskCache(String detectorId) {
         return hcTaskCaches.computeIfAbsent(detectorId, id -> new ADHCBatchTaskCache());
     }
 
+    /**
+     * Check if there is any HC task running on current node.
+     * @param detectorId detector id
+     * @return true if find detector id in any entity task or HC cache
```
ohltyler, line 796: Would you need to check the task states as well? Or only running tasks will be in the cache?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -405,6 +517,235 @@ protected void forwardStaleRunningEntitiesToCoordinatingNode(
...
+            logger.debug("Release checking task slot semaphore on lead node for detector {}", detectorId);
+        });
+        hashRing.getNodesWithSameLocalAdVersion(nodes -> {
+            int maxAdTaskSlots = nodes.length * maxAdBatchTaskPerNode;
+            ADStatsRequest adStatsRequest = new ADStatsRequest(nodes);
+            adStatsRequest
+                .addAll(ImmutableSet.of(AD_USED_BATCH_TASK_SLOT_COUNT.getName(), AD_DETECTOR_ASSIGNED_BATCH_TASK_SLOT_COUNT.getName()));
+            client.execute(ADStatsNodesAction.INSTANCE, adStatsRequest, ActionListener.wrap(adStatsResponse -> {
+                int totalUsedTaskSlots = 0; // Total entity tasks running on worker nodes
+                int totalAssignedTaskSlots = 0; // Total assigned task slots on coordinating nodes
```
ohltyler, line 566: May be helpful to add some comments on the difference and meaning of each of these types of task slots

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: will remove this

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -405,6 +517,235 @@ protected void forwardStaleRunningEntitiesToCoordinatingNode(
...
+                        availableAdTaskSlots
+                    );
+                    return;
+                }
+
+                long dataStartTime = detectionDateRange.getStartTime().toEpochMilli();
+                long dataEndTime = detectionDateRange.getEndTime().toEpochMilli();
+
+                if (detector.isMultientityDetector()) {
+                    if (detector.isMultiCategoryDetector()) {
```
ohltyler, line 607: May be helpful to add some comments on the different searches that happen based on if the detector is single entity vs. multi entity (single category) vs. multi entity (multi category)

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: I think it's ok to keep as it as 
1. These two functions `scaleUpDetectorTaskSlots` and `scaleDownHCDetectorTaskSlots` is straight forward and simply set value in cache.
2. If any exception, the client using them will handle.

src/main/java/org/opensearch/ad/transport/ForwardADTaskTransportAction.java
```diff
@@ -79,33 +80,58 @@ protected void doExecute(Task task, ForwardADTaskRequest request, ActionListener
...
 
         String entityValue = adTaskManager.convertEntityToString(adTask);
 
         switch (adTaskAction) {
+            case APPLY_FOR_TASK_SLOTS:
+                logger.debug("Received APPLY_FOR_TASK_SLOTS action for detector {}", detectorId);
+                adTaskManager.checkTaskSlots(adTask, detector, detectionDateRange, user, ADTaskAction.START, transportService, listener);
+                break;
+            case CHECK_AVAILABLE_TASK_SLOTS:
+                logger.debug("Received SCALE_TASK_SLOTS action for detector {}", detectorId);
```
ohltyler, line 94: nit: the logged action value doesn't match the case value

src/main/java/org/opensearch/ad/model/DetectorProfile.java
ylwu-amzn: Different way to organize task profile. Previous design treat detector level and entity level task profile as the same `ADTaskProfile`, that's not BWC as the AD 1.0 code is using `ADTaskProfile`. So now I changed it as `ADTaskProfile` to make it BWC with 1.0. And put both detector&entity level profiles in it.

ylwu-amzn: Check example https://github.com/opensearch-project/anomaly-detection/pull/170#discussion_r694416515

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
ylwu-amzn: good catch, will fix

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
ylwu-amzn: That's not for one detector. We need to migrate all AD jobs for all old detectors. One detector can only have only one AD job as the job doc id is detector id.

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Only running tasks will be in cache.

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
ohltyler: Got it. I think I confused myself when testing it out and seeing the `AD data migration done` for each detector, thinking it was running this method per detector. I think this was some earlier version of the changes and I can't reproduce now. Can ignore.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: sure, will add

src/main/java/org/opensearch/ad/transport/ForwardADTaskTransportAction.java
ylwu-amzn: Good catch, will fix.

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -263,6 +297,50 @@ public void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> add
...
+     *    ring. If hash ring changes, it's possible that the next job run will use a different
+     *    node to run RCF partition. Then we need to restore model on the new node and clean up
+     *    old model partitions on old node. That model migration between nodes may bring heavy
+     *    load to cluster.
+     *
+     * @return true if it's eligible to rebuild hash ring
+     */
+    private boolean eligibleToRebuildHashRingWithCooldown() {
+        // Check if there is any node change event
+        if (nodeChangeEvents.isEmpty() && !adVersionCirclesWithCooldown.isEmpty()) {
```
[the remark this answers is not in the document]
ylwu-amzn: Sure, good point

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -263,6 +297,50 @@ public void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> add
         }
     }
 
+    private void rebuildAdVersionCirclesWithCooldown() {
+        // Check if it's eligible to rebuild hash ring with cooldown
+        if (eligibleToRebuildHashRingWithCooldown()) {
+            LOG.info("Rebuild AD version hash ring with cooldown");
+            adVersionCirclesWithCooldown = new TreeMap<>(adVersionCircles);
+            nodeChangeEvents.poll();
```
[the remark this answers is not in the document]
ylwu-amzn: It's possible that multiple threads add new event to `nodeChangeEvents`, but this is the only place to consume/poll the event and there is only one thread poll it as we are using `adVersionCircleInProgress` semaphore (1) to control.

Looks good to read size and dequeue same amount, can save some hash ring rebuilds. Will change.

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -237,7 +268,10 @@ public void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> add
                 }
                 LOG.info("All nodes in AD version hash ring: {}", nodeAdVersions);
 
-                if (adVersionCircles.size() > 0 && adVersionCircles.lastEntry().getKey().after(Version.V_1_0_0)) {
+                // rebuild AD version hash ring with cooldown after all new node added.
+                rebuildAdVersionCirclesWithCooldown();
+
+                if (adVersionCircles.size() > 0 && adVersionCircles.lastEntry().getKey().onOrAfter(Version.V_1_1_0)) {
```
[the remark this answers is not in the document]
ylwu-amzn: sure, added

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -456,7 +466,13 @@ public ADTaskCancellationState cancelByDetectorId(String detectorId, String reas
...
     public boolean isCancelled(String taskId) {
         ADBatchTaskCache taskCache = getBatchTaskCache(taskId);
-        return taskCache.isCancelled();
+        String detectorId = taskCache.getDetectorId();
+        ADHCBatchTaskCache hcTaskCache = hcTaskCaches.get(detectorId);
+        boolean hcDetectorStopped = false;
+        if (hcTaskCache != null) {
+            hcDetectorStopped = hcTaskCache.getHistoricalAnalysisCancelled();
+        }
+        return taskCache.isCancelled() || hcDetectorStopped;
```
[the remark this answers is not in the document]
ylwu-amzn: Will add more comments.

If a new entity task comes after cancel event, then we have no chance to set it as cancelled. So we need to check `hcDetectorStopped` for HC detector to know if the HC detector cancelled or not. For single entity detector, it has just 1 task, just need to check `taskCache.isCancelled`

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -172,6 +166,8 @@ public void buildCirclesOnAdVersionsDirectly() {
      * 1. Delete removed nodes from AD version hash ring.
      * 2. Add new nodes to AD version hash ring
      *
+     * If fail to acquire semaphore to update AD version hash ring, will return false to
```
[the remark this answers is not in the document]
ylwu-amzn: Will add more comments and tune the variable name

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -197,16 +212,15 @@ public void buildCirclesOnAdVersionsDirectly() {
...
      * @param addedNodeIds added node ids
      * @param actionListener action listener
      */
-    public void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> addedNodeIds, ActionListener<Boolean> actionListener) {
-        if (!adVersionCircleInProgress.tryAcquire()) {
-            LOG.info("AD version hash ring change in progress, return.");
-            actionListener.onResponse(false);
-            return;
-        }
+    private void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> addedNodeIds, ActionListener<Boolean> actionListener) {
```
ylwu-amzn, line 223: @kaituo , per your ask, I moved the `adVersionCircleInProgress.tryAcquire()` block to functions which calling buildCirclesOnAdVersions:  `HashRing#buildCirclesOnAdVersions(ActionListener)` and `HashRing#buildCirclesOnAdVersions(DiscoveryNodes.Delta, ActionListener)`

**Pros**: This can return earlier if doesn't get semaphore, so we can save some cost, like the set comparison.
**Cons**: If someone wants to build new function on top of buildCirclesOnAdVersions, should pay attention to acquire `adVersionCircleInProgress` semaphore first.

kaituo: Thanks for the accommodation.  The cons is troublesome indeed.  And current code build on demand on some code path, not for all of them.  This will increase maintenance cost and not efficient.  I wrote a draft version (forgive my grammar or naming error/discrepancy, just to show you what I have in mind).  Could you check? Specifically,

First, I removed buildCirclesOnAdVersions(DiscoveryNodes.Delta delta, ActionListener<Boolean> listener) and depend on buildCirclesOnAdVersions(ActionListener<Boolean> actionListener).  So we have one public build method that can be called in all of the places.  That's what the original HashRIng code does. This ease maintenance and avoids possible errors mentioned in your cons.
Second, the only difference between real time and historical is the cooldown.  Other things are the same.  So I check nodeChangeEvents.isEmpty() in buildCirclesOnAdVersions.
Third, I put the logic of cooldown logic inside buildAndGetOwningNodeRealTime. We still have 2 hash rings.  buildAndGetOwningNodeRealTime is in charge of updating real time hash ring. 
Fourth, do extra inProgress.availablePermits() >= 1 check inside buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> addedNodeIds, ActionListener<Boolean> actionListener).  The objective is to fail fast and double check misuse of inProgress.

```
public class HashRing {
    ...
    private TreeMap<Version, TreeMap<Integer, DiscoveryNode>> circle, realTimeCircle;
    private Semaphore inProgress, realTimeInProgress;
    // the UTC epoch milliseconds of the most recent successful update
    private long lastUpdate, realTimeLastUpdate;
    ...
    private ConcurrentLinkedQueue<Boolean> nodeChangeEvents;

    public boolean buildCirclesOnAdVersions(ActionListener<Boolean> actionListener) {
        
        if (nodeChangeEvents.isEmpty() && !circle.isEmpty()) {
            return false;
        }

        if (!inProgress.tryAcquire()) {
            LOG.info("Hash ring change in progress, return.");
            return false;
        }

        DiscoveryNode[] eligibleDataNodes = nodeFilter.getEligibleDataNodes();
        Set<String> eligibleNodeIds = new HashSet<>();
        for (DiscoveryNode node : eligibleDataNodes) {
            eligibleNodeIds.add(node.getId());
        }
        Set<String> currentNodeIds = nodeAdVersions.keySet();
        Set<String> removedNodeIds = Sets.difference(currentNodeIds, eligibleNodeIds);
        Set<String> addedNodeIds = Sets.difference(eligibleNodeIds, currentNodeIds);
        buildCirclesOnAdVersions(removedNodeIds, addedNodeIds, actionListener);
    }

    private void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> addedNodeIds, ActionListener<Boolean> actionListener) {
       if (inProgress.availablePermits() >= 1) {
              // we need inProgress to have 0 permits.  If not, fail fast, throw exception
              ...
        }
        try {
            ...
                        int size = nodeChangeEvents.size();
                        TreeMap<Integer, DiscoveryNode> circle = null;
                        for (PluginInfo pluginInfo : plugins.getPluginInfos()) {
                            if (AD_PLUGIN_NAME.equals(pluginInfo.getName()) || AD_PLUGIN_NAME_FOR_TEST.equals(pluginInfo.getName())) {
                                Version version = ADVersionUtil.fromString(pluginInfo.getVersion());
                                circle = adVersionCircles.computeIfAbsent(version, key -> new TreeMap<>());
                                nodeAdVersions.put(curNode.getId(), ADVersionUtil.fromString(pluginInfo.getVersion()));
                                break;
                            }
                        }
                        if (circle != null) {
                            for (int i = 0; i < VIRTUAL_NODE_COUNT; i++) {
                                circle.put(Murmur3HashFunction.hash(curNode.getId() + i), curNode);
                            }
                        }
			            circle = newCircle;
			            while (size-- > 0) {
			                Boolean poll = nodeChangeEvents.poll();
			                if (poll == null) {
			                    break;
			                }
			            }
			            lastUpdate = clock.millis();
                    }
                }
                LOG.info("All nodes in AD version hash ring: {}", nodeAdVersions);

                ...
            }, e -> {
                adVersionCircleInProgress.release();
                actionListener.onFailure(e);
                LOG.error("Fail to get node info to build AD version hash ring", e);
            }));
        } catch (Exception e) {
            LOG.error("Failed to build AD version circles", e);
            adVersionCircleInProgress.release();
            actionListener.onFailure(e);
        }
    }

    private void buildCirclesOnAdVersionsDirectly() {
    	buildCirclesOnAdVersions(ActionListener.wrap(r->.., e->...))
    }

    /**
     * Compute the owning node of modelID using consistent hashing
     * @param key example: http-latency-rcf-1
     * @return the owning node of the modeID
     */
    public Optional<DiscoveryNode> buildAndGetOwningNodeRealTime(String key) {
        buildCirclesOnAdVersionsDirectly();

        if (realTimeInProgress.tryAcquire() && realTimeLastUpdate != lastUpdate && clock.millis() - realTimeLastUpdate > coolDownPeriod.getMillis()) {
            try {
                // recursive deep copy of circle
	        	realTimeCircle = ... 
	        	realTimeLastUpdate = lastUpdate
            } finally {
            	realTimeInProgress.release();
            }
        }

        return getOwningNode(realTimeCircle, key);
    }

    public Optional<DiscoveryNode> buildAndGetOwningNode(String key) {
        buildCirclesOnAdVersionsDirectly();

        return getOwningNode(circle, key);
    }

    public Optional<DiscoveryNode> getOwningNode(TreeMap<Integer, DiscoveryNode> hashRing, String key) {
    	int modelHash = Murmur3HashFunction.hash(key);
        Map.Entry<Integer, DiscoveryNode> entry = hashRing.higherEntry(modelHash);

        // The method can return an empty Optional. Say two concurrent getOwningNode requests to
        // the hash ring before it's been built. The first one starts building it,
        // turning on inProgress. The second one returns from build and continues on to
        // the rest of hashing and look up while the ring is still being built and thus empty.
        // The second getOwningNode request returns an empty Optional in this case.
        return Optional.ofNullable(Optional.ofNullable(entry).orElse(hashRing.firstEntry())).map(x -> x.getValue());
    }
}

```

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -300,9 +316,19 @@ public void buildCirclesOnAdVersions(Set<String> removedNodeIds, Set<String> add
     private void rebuildAdVersionCirclesWithCooldown() {
         // Check if it's eligible to rebuild hash ring with cooldown
         if (eligibleToRebuildHashRingWithCooldown()) {
-            LOG.info("Rebuild AD version hash ring with cooldown");
+            LOG.info("Rebuild AD version hash ring with cooldown, nodeChangeEvents size {}", nodeChangeEvents.size());
+            int size = nodeChangeEvents.size();
             adVersionCirclesWithCooldown = new TreeMap<>(adVersionCircles);
-            nodeChangeEvents.poll();
+            // It's possible that multiple threads add new event to nodeChangeEvents,
```
kaituo, line 322: The copy in line 321 is not a complete deep copy since adVersionCircles is a map of map. Check the following example, where s2 does a copy like yours and s3 is a recursive copy.

```
jshell> TreeMap<String, TreeMap<Integer, Object>> s = new TreeMap<>()
s ==> {}

jshell> TreeMap<Integer, Object> insert = new TreeMap<>()
insert ==> {}

jshell> insert.put(1, new Object())
$3 ==> null

jshell> s.put("1", insert)
$4 ==> null

jshell> TreeMap<String, TreeMap<Integer, Object>> s2 = new TreeMap<>(s)
s2 ==> {1={1=java.lang.Object@17f052a3}}

jshell> s2
s2 ==> {1={1=java.lang.Object@17f052a3}}

jshell> s
s ==> {1={1=java.lang.Object@17f052a3}}

jshell> insert.put(2, new Object())
$9 ==> null

jshell> s
s ==> {1={1=java.lang.Object@17f052a3, 2=java.lang.Object@2e5d6d97}}

jshell> s2
s2 ==> {1={1=java.lang.Object@17f052a3, 2=java.lang.Object@2e5d6d97}}

jshell> for (Map.Entry<String, TreeMap<Integer, Object>> entry: s.entrySet()) {
   ...>   s3.put(new String(entry.getKey()), new TreeMap<>(entry.getValue()));
   ...> }

jshell> s3
s3 ==> {1={1=java.lang.Object@17f052a3, 2=java.lang.Object@2e5d6d97}}

jshell> s
s ==> {1={1=java.lang.Object@17f052a3, 2=java.lang.Object@2e5d6d97}}

jshell> insert.put(3, new Object())
$16 ==> null

jshell> s
s ==> {1={1=java.lang.Object@17f052a3, 2=java.lang.Object@2e5d6d97, 3=java.lang.Object@6267c3bb}}

jshell> s3
s3 ==> {1={1=java.lang.Object@17f052a3, 2=java.lang.Object@2e5d6d97}}
```

ylwu-amzn: Good point, will fix

src/main/java/org/opensearch/ad/cluster/HashRing.java
ylwu-amzn: Summary of our offline discussion. Tuned hash ring method/variable name to make it clear and add semaphore check before building hash ring.
1. If we use two semaphore, the realtime may copy in middle of historical hash ring updating. That means some node may have less virtual node than other node in realtime hash ring. Then the load will be skewed for realtime as we depends on hash ring totally to distribute load.
2. We should update realtime hash ring on demand based on event queue. But for some edge cases like the node is busy with GC, the node change event may not be consumed, so the event queue is still empty, that means we may miss that node in hash ring for this case. For historical analysis, if we miss some node, it will run slower. So we check removed/added node every time to make sure the historical analysis is always up to date. And historical analysis will sleep for some time (5 seconds by default) between pieces, comparing with this sleep time, the performance loss of checking removed/added node can be ignored.

ylwu-amzn:
CI workflow failed for commit 16 as OpenSearch changed `PluginInfo` constructor in this PR https://github.com/opensearch-project/OpenSearch/pull/848/files, a new field `customFolderName` added

> public PluginInfo(String name, String description, String version, Version opensearchVersion, String javaVersion,
                      String classname, **String customFolderName**, List<String> extendedPlugins, boolean hasNativeController)

kaituo:
@ylwu-amzn The overall PR looks good to me.  Will wait for you to fix the build.

ohltyler:
Will check the updated and comments from the last time I looked at this PR. As discussed offline, will you be adding testing in a follow-up PR @ylwu-amzn ?

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
```diff
@@ -0,0 +1,292 @@
...
+                    if (r.isAcknowledged()) {
+                        logger.info("Created {} with mappings.", CommonName.DETECTION_STATE_INDEX);
+                        migrateDetectorInternalStateToRealtimeTask();
+                    } else {
+                        String error = "Create index " + CommonName.DETECTION_STATE_INDEX + " with mappings not acknowledged";
+                        logger.warn(error);
+                    }
+                }, e -> {
+                    if (ExceptionsHelper.unwrapCause(e) instanceof ResourceAlreadyExistsException) {
+                        migrateDetectorInternalStateToRealtimeTask();
```
ohltyler, line 116: How would this exception occur? Is this to handle some race condition where it gets created some time after the `if` check on line 102?

src/main/java/org/opensearch/ad/cluster/ADVersionUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    public static Version fromString(String adVersion) {
+        if (AD_PLUGIN_VERSION_FOR_TEST.equals(adVersion)) {
+            return Version.CURRENT;
+        }
+        return Version.fromString(normalizeVersion(adVersion));
+    }
+
+    public static String normalizeVersion(String adVersion) {
+        if (adVersion == null) {
+            throw new IllegalArgumentException("Wrong AD version " + adVersion);
```
ohltyler, line 31: Minor: change the exception to logically match the fact that `adVersion` was null?

ylwu-amzn:
> Will check the updated and comments from the last time I looked at this PR. As discussed offline, will you be adding testing in a follow-up PR @ylwu-amzn ?

Yes, the test will be in another PR.

src/main/java/org/opensearch/ad/cluster/ADDataMigrator.java
ylwu-amzn: Yes, while we migrate data, it's possible that user run some historical detector and it will create detection state index. Then we will see such exception.

src/main/java/org/opensearch/ad/cluster/ADVersionUtil.java
ylwu-amzn: sure, will tune error message

ohltyler (approved):

kaituo (approved):

PR #174, opened 2021-08-13
train ercf model

wnbts:

### Description
This pull request training anomaly detection models with ercf.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -34,6 +34,7 @@ import java.time.Instant;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -329,85 +330,41 @@ private void trainModelFromDataSegments(List<double[][]> dataPoints, Entity enti
...
-        List<double[]> allScores = new ArrayList<>();
-        int totalLength = 0;
-        // get continuous data points and send for training
-        for (double[][] continuousDataPoints : dataPoints) {
-            double[] scores = trainRCFModel(continuousDataPoints, rcf);
-            allScores.add(scores);
-            totalLength += scores.length;
-        }
+        ExtendedRandomCutForest ercf = new ExtendedRandomCutForest(
+            RandomCutForest
```
ylwu-amzn, line 334: Why pass `RandomCutForest.builder` into `ExtendedRandomCutForest` rather than `RandomCutForest` ?

wnbts: Because that's the current api available. There is no reason another constructor with a rcf cannot be added. When it's added, the client code here can change to use that.

ylwu-amzn (approved):
LGTM

LiuJoyceC (approved):

PR #179, opened 2021-08-16
get checkpoint for ercf

wnbts:

### Description
This pr creats the checkpoints for ercf models.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
@@ -82,6 +82,9 @@ import org.opensearch.index.reindex.DeleteByQueryAction;
 import org.opensearch.index.reindex.DeleteByQueryRequest;
 import org.opensearch.index.reindex.ScrollableHitSource;
 
+import com.amazon.randomcutforest.ERCF.ERCFMapper;
+import com.amazon.randomcutforest.ERCF.ERCFState;
+import com.amazon.randomcutforest.ERCF.ExtendedRandomCutForest;
 import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.config.Precision;
 import com.amazon.randomcutforest.serialize.json.v1.V1JsonToV2StateConverter;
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -337,10 +349,68 @@ public String toCheckpoint(EntityModel model) {
...
+            }
+        } finally {
+            if (isCheckout) {
+                try {
+                    serializeRCFBufferPool.returnObject(buffer);
+                } catch (Exception e) {
+                    logger.warn("Failed to return buffer to pool", e);
+                }
+            }
+        }
```
LiuJoyceC, line 384: If a `RuntimeException` triggers the `catch` block, the `finally` block will still run afterwards. Is it the correct behavior to still invoke `serializeRCFBufferPool.returnObject(buffer)` after already invoking `serializeRCFBufferPool.invalidateObject(buffer)`?

wnbts: That might not be correct. I changed the behavior to not return it in this case. Thanks for pointing it out!

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -337,10 +349,69 @@ public String toCheckpoint(EntityModel model) {
...
+            }
+        } finally {
+            if (requireCheckin) {
+                try {
+                    serializeRCFBufferPool.returnObject(buffer);
+                } catch (Exception e) {
+                    logger.warn("Failed to return buffer to pool", e);
+                }
+            }
+        }
```
LiuJoyceC, line 385: [Non-blocking, can still approve PR] Since the logic inside the `finally` block is not intended to be run if the `catch` block is triggered, it would be simpler to just remove the `finally` block and move the logic in it into the `try` block.

LiuJoyceC (approved):
Added an additional suggestion to simplify the logic, but is not required for approval.

PR #180, opened 2021-08-16
score from ercf

wnbts:

### Description
This pr adds ercf scoring.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/ModelManager.java
+++ b/src/main/java/org/opensearch/ad/ml/ModelManager.java
@@ -63,6 +63,7 @@ import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 
 import com.amazon.randomcutforest.ERCF.AnomalyDescriptor;
+import com.amazon.randomcutforest.ERCF.ExtendedRandomCutForest;
 import com.amazon.randomcutforest.RandomCutForest;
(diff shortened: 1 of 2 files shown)


LiuJoyceC (approved):

src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -838,33 +839,41 @@ public ThresholdingResult getAnomalyResultForEntity(
...
+                q.stream().forEach(s -> ercf.process(s));
+                q.clear();
+            });
+            result = toResult(ercf.process(feature));
+        } else {
+            RandomCutForest rcf = model.getRcf();
+            ThresholdingModel threshold = model.getThreshold();
+            if (rcf == null || threshold == null) {
+                return new ThresholdingResult(0, 0, 0);
+            }
```
LiuJoyceC, line 858: For all of the other cases, instead of directly returning a result, you set the variable `result` and at the end of the function call:
```
modelState.setLastUsedTime(clock.instant());
return result;
```
Why is the return logic different here? Should we still make sure `modelState.setLastUsedTime(clock.instant())` is invoked before returning?

wnbts: This is the current behavior that doesn't need to be changed since the behavior is ok and this branch for rcf and thresholding will be shortly removed to leave ercf only.

PR #158, opened 2021-08-05
Adding mixed cluster, rolling upgrade, restart upgrade bwc tests

VachaShah:
Signed-off-by: Vacha <vachshah@amazon.com>

### Description
Adding bwc tests and the supporting test framework for bwc. This is part of [automating backwards compatibility support for the OpenSearch project](https://github.com/opensearch-project/opensearch-build/issues/90). The bwc scenarios covered are:

1. Mixed cluster
2. Rolling Upgrade
3. Restart Upgrade

To run the mixed cluster test: `./gradlew adBwcCluster#mixedClusterTask -DmixedCluster=123 -Dtests.security.manager=false`

To run the rolling upgrade test: `./gradlew adBwcCluster#rollingUpgradeClusterTask -DrollingUpgradeCluster=123 -Dtests.security.manager=false`

To run the restart upgrade test: `./gradlew adBwcCluster#fullRestartClusterTask -DfullRestartCluster=123 -Dtests.security.manager=false`

These tests will be hooked up to CI as part of a separate PR.
### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #181, opened 2021-08-17
Add bwc tests to run with CI

VachaShah:
### Description
Adding bwc tests introduced in #158 to run with CI for `anomaly-detection`. Also adding the test run commands in the developer guide.
 
### Issues Resolved
[opensearch-build#222](https://github.com/opensearch-project/opensearch-build/issues/222)
 
### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -72,6 +72,26 @@ jobs:
           cp ./build/distributions/*.zip ../src/test/resources/job-scheduler
           echo "Copied ./build/distributions/*.zip to ../src/test/resources/job-scheduler ..."
           ls ../src/test/resources/job-scheduler
+          echo "Creating ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT ..."
+          mkdir -p ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+          pwd
+          echo "Copying ./build/distributions/*.zip to ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT ..."
+          ls ./build/distributions/
+          cp ./build/distributions/*.zip ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+          echo "Copied ./build/distributions/*.zip to ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT ..."
+          ls ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+
+      - name: Assemble anomaly-detection
+        run: |
+          ./gradlew assemble -Dopensearch.version=1.1.0-SNAPSHOT
+          echo "Creating ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT ..."
+          mkdir -p ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT
+          pwd
+          echo "Copying ./build/distributions/*.zip to ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT ..."
+          ls ./build/distributions/
+          cp ./build/distributions/*.zip ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT
+          echo "Copied ./build/distributions/*.zip to ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT ..."
+          ls ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT    
 
       - name: Build and Run Tests
         run: |
@@ -123,6 +143,21 @@ jobs:
             echo "Security plugin is NOT available"
             ./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster"
           fi
+
+      - name: Mixed cluster backwards compatibility tests
+        run: |
+          echo "Running mixed cluster backwards compatibility tests ..."
+          ./gradlew adBwcCluster#mixedClusterTask -DmixedCluster=123 -Dtests.security.manager=false
+
+      - name: Rolling upgrade backwards compatibility tests
+        run: |    
+          echo "Running rolling upgrade backwards compatibility tests ..."
+          ./gradlew adBwcCluster#rollingUpgradeClusterTask -DrollingUpgradeCluster=123 -Dtests.security.manager=false
+
+      - name: Full cluster restart backwards compatibility tests
+        run: |     
+          echo "Running full restart cluster backwards compatibility tests ..."
+          ./gradlew adBwcCluster#fullRestartClusterTask -DfullRestartCluster=123 -Dtests.security.manager=false
 
       - name: Upload Coverage Report
         uses: codecov/codecov-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -72,6 +72,26 @@ jobs:
...
+          ls ./build/distributions/
+          cp ./build/distributions/*.zip ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+          echo "Copied ./build/distributions/*.zip to ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT ..."
+          ls ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+
+      - name: Assemble anomaly-detection
+        run: |
+          ./gradlew assemble -Dopensearch.version=1.1.0-SNAPSHOT
+          echo "Creating ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT ..."
+          mkdir -p ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT
```
saratvemulapalli, line 88: Do we need this?
Gradle plugin automatically can pick up the plugin from the build directory.
Take a look at: https://github.com/opensearch-project/anomaly-detection/blob/main/build.gradle#L235

.github/workflows/CI.yml
```diff
@@ -124,6 +144,21 @@ jobs:
             ./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster"
           fi
 
+      - name: Mixed cluster backwards compatibility tests
+        run: |
+          echo "Running mixed cluster backwards compatibility tests ..."
+          ./gradlew adBwcCluster#mixedClusterTask -DmixedCluster=123 -Dtests.security.manager=false
```
saratvemulapalli, line 150: I understand we are using mixedCluster variable to run some specific tests. 
Can we use a boolean flag instead of an arbitrary number like 123?
```suggestion
          ./gradlew adBwcCluster#mixedClusterTask -DmixedCluster=true -Dtests.security.manager=false
```

.github/workflows/CI.yml
```diff
@@ -124,6 +144,21 @@ jobs:
...
+
+      - name: Rolling upgrade backwards compatibility tests
```
saratvemulapalli, line 152: This is nice to separate out dedicated tests but is there a way to run all BWC tests with a single command?

saratvemulapalli (commented):
Couple of questions, most of the changes look good to me.

.github/workflows/CI.yml
VachaShah: There is currently no task to run that, I can add one that runs all three of them together.

.github/workflows/CI.yml
VachaShah: This is needed since the `upgradeNodeAndPluginToNextVersion` method is designed to take in a list of `Provider<RegularFile>` for installing plugins on the node during the upgrade. The `plugin` method is separate which has two implementations, one for taking in a file and one for taking in the path which is used for the node.

saratvemulapalli: Got it. That makes sense. I think it worth overloading that function to have another method to take care of this. 
We should avoid as much as possible manually copying artifacts for tests. 

We can goahead with this change, but can you also make a change in OpenSearch to add this functionality?

.github/workflows/CI.yml
saratvemulapalli: Yeah definitely, I think its worth adding that task and will help us when we would want to publish nightly builds and run these tests in one go.

.github/workflows/CI.yml
VachaShah: So this method has the file list as input since the `job-scheduler` artifact would be a zip file and since AD has a dependency on that, we have to supply it at the same time so that the newer version plugins can be added when the node is upgraded. The new functionality can be useful for plugins that don't have any other plugins as dependency.

.github/workflows/CI.yml
VachaShah: Makes sense, updated it to `true` for the flag.

.github/workflows/CI.yml
```diff
@@ -72,6 +72,26 @@ jobs:
...
+          cp ./build/distributions/*.zip ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+          echo "Copied ./build/distributions/*.zip to ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT ..."
+          ls ../src/test/resources/org/opensearch/ad/bwc/job-scheduler/1.1.0.0-SNAPSHOT
+
+      - name: Assemble anomaly-detection
+        run: |
+          ./gradlew assemble -Dopensearch.version=1.1.0-SNAPSHOT
+          echo "Creating ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT ..."
+          mkdir -p ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.1.0.0-SNAPSHOT
+          pwd
```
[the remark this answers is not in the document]
VachaShah: Removed it.

.github/workflows/CI.yml
VachaShah: Definitely in agreement, I will work on adding a single task which can be invoked to run the bwc tests after this PR.

saratvemulapalli (approved):
Thanks @VachaShah for this change!
This is great.

I would like to get AD team to also take a look.
@ylwu-amzn can you review this change?

DEVELOPER_GUIDE.md
```diff
@@ -41,6 +41,9 @@ Currently we just put RCF jar in lib as dependency. Plan to publish to Maven and
 4. ` ./gradlew :integTest --tests="**.test execute foo"` runs a single integration test class or method
 5. `./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=true -Duser=admin -Dpassword=admin` launches integration tests against a local cluster and run tests with security
 6. `./gradlew spotlessApply` formats code. And/or import formatting rules in `.eclipseformat.xml` with IDE.
+7. `./gradlew adBwcCluster#mixedClusterTask -DmixedCluster=true -Dtests.security.manager=false` launches a cluster with three nodes of bwc version of OpenSearch with anomaly-detection and job-scheduler and tests backwards compatibility by upgrading one of the nodes with the current version of OpenSearch with anomaly-detection and job-scheduler creating a mixed cluster.
```
ylwu-amzn, line 44: Will the three nodes run three different OpenSearch versions? Or 1 node runs old version, other 2 nodes run new version?

DEVELOPER_GUIDE.md
```diff
@@ -41,6 +41,9 @@ Currently we just put RCF jar in lib as dependency. Plan to publish to Maven and
...
+8. `./gradlew adBwcCluster#rollingUpgradeClusterTask -DrollingUpgradeCluster=true -Dtests.security.manager=false` launches a cluster with three nodes of bwc version of OpenSearch with anomaly-detection and job-scheduler and tests backwards compatibility by performing rolling upgrade of each node with the current version of OpenSearch with anomaly-detection and job-scheduler.
+9. `./gradlew adBwcCluster#fullRestartClusterTask -DfullRestartCluster=true -Dtests.security.manager=false` launches a cluster with three nodes of bwc version of OpenSearch with anomaly-detection and job-scheduler and tests backwards compatibility by performing a full restart on the cluster upgrading all the nodes with the current version of OpenSearch with anomaly-detection and job-scheduler.
```
ylwu-amzn, line 46: Do we have some historical data of old version ready before full restart ? For AD, that will be create some detectors and run it to generate some AD result; then fully restart and test. 

Same question for other 2 BWC tests: mixed cluster and rolling upgrade.

DEVELOPER_GUIDE.md
VachaShah: The old version cluster will have 3 nodes of the old (same) version of OpenSearch and then one node will be upgraded to current version (for example 1.1) along with anomaly-detection and job-scheduler upgrade to 1.1 on that node. 2 other nodes will still be on the old version, this creates a mixed cluster.

DEVELOPER_GUIDE.md
VachaShah: AD can definitely add more assertions to the tests according to the various scenarios.

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #186, opened 2021-08-20
Implement custom pmml model support.

wqixian:
### Description
Implemented support for custom models in .pmml format for real-time, single-entity AD jobs.

Overview: some community user asked for this functionality. Since the ML plugin has supported such functionality (still in development branch), AD can make prediction API calls to the ML plugin which will run the actual tasks. The goal of this PR is to add the internal logic for supporting custom pmml models.

Code changes: added an optional parameter `mlModelId` for anomaly detector, which will be used to determine whether the detector is using a pmml custom model, and if so, the newly added `PMMLResultTransportAction` will be created, in which the ML model id will be passed to the ML plugin to find the model and perform inference jobs. In `AnomalyResultTransportAction`, the above logic is added, and a `PMMLResultListener` is added, which is used to handle the results from pmml transport calls.

Testing: some tests have been added, but due to limited time, they are not enough to test all the new functionalities. Classes that need more tests are listed in the "jacocoExclusions" in the build.gradle files.

Future tasks and improvements:

1. Create one instance of the ML node client during plugin initialization, and use that instead of creating a new one for each pmml transport action.
2. Some transport actions haven’t adapted to the new detectors using custom models. For example, in index anomaly detector, search the passed-in ml model id (if any) to make sure it exists in the ML plugin. Another example is, for the transport actions that are related to the cases mention in 3 (the following task), such as preview, need to make sure they adapt and work with the new features (before and after those cases are supported).
3. The historical and multi-entity cases should be covered.
4. We can add an optional field “boolean isUsingCustomModel” in ADTaskmanager - updateLatestRealtimeTask, so we can pass this in from the newly added functions in AnomalyDetector, and check whether we are using custom models or RCF to avoid having to check the RCF total update numbers for custom models. That way we don’t rely on the dummy value passed in as rcfTotalUpdates to AnomalyResultResponse.
5. The constructor of anomaly detectors has one limitation: we don’t support the case that detectorType is null but mlModelId is not null.
6. If the value conversions are not ideal, we can try to find better ways to improve those. One possibility is that the confidence can come from the number of estimators (trees) in the isolation forest, if the model has such attributes. Otherwise it can remain as 1 (dummy value). Another possibility is that to find a better mathematical way to convert the decision function to the anomaly grade in a non-linear way, etc.
7. The "initializing" state for custom detectors won't end. Need to check more details and fix.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,7 @@ buildscript {
         opensearch_build = opensearch_version.replaceAll(/(\.\d)([^\d]*)$/, '$1.0$2')
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
         job_scheduler_version = System.getProperty("job_scheduler.version", opensearch_build)
+        ml_version = System.getProperty("ml.version", "1.0.0.0")
     }
 
     repositories {
@@ -529,6 +530,7 @@ List<String> jacocoExclusions = [
         'org.opensearch.ad.transport.GetAnomalyDetectorResponse',
         'org.opensearch.ad.transport.ADBatchAnomalyResultRequest',
         'org.opensearch.ad.transport.ADBatchTaskRemoteExecutionTransportAction',
+        'org.opensearch.ad.transport.PMMLResultTransportAction'
 ]
 
 jacocoTestCoverageVerification {
@@ -572,6 +574,7 @@ dependencies {
     compileOnly "org.opensearch.plugin:opensearch-scripting-painless-spi:${opensearch_version}"
     compileOnly "org.opensearch:opensearch-job-scheduler-spi:${job_scheduler_version}"
     compile "org.opensearch:common-utils:${common_utils_version}"
+    compile "org.opensearch.ml:opensearch-ml-client:${ml_version}"
     compile "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     compile group: 'com.google.guava', name: 'guava', version:'29.0-jre'
     compile group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
--- a/build.gradle
+++ b/build.gradle
@@ -530,6 +530,8 @@ List<String> jacocoExclusions = [
         'org.opensearch.ad.transport.GetAnomalyDetectorResponse',
         'org.opensearch.ad.transport.ADBatchAnomalyResultRequest',
         'org.opensearch.ad.transport.ADBatchTaskRemoteExecutionTransportAction',
+
+        // TODO: custom model support caused coverage drop
         'org.opensearch.ad.transport.PMMLResultTransportAction'
 ]
 
--- a/src/main/java/org/opensearch/ad/ml/PMMLResult.java
+++ b/src/main/java/org/opensearch/ad/ml/PMMLResult.java
@@ -9,28 +9,15 @@
  * GitHub history for details.
  */
 
-/*
- * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License").
- * You may not use this file except in compliance with the License.
- * A copy of the License is located at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * or in the "license" file accompanying this file. This file is distributed
- * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
- * express or implied. See the License for the specific language governing
- * permissions and limitations under the License.
- */
-
 package org.opensearch.ad.ml;
 
 import java.util.Objects;
 
-// A reference for the format to expect for external call responses from the ML plugin.
-// This class is not used now. In PMML result transport action, we simply convert the results from the ML plugin
-// into values needed for the anomaly result response.
+/**
+ * A reference for the format to expect for external call responses from the ML plugin.
+ * This class is not used now. In PMML result transport action, we simply convert the results from the ML plugin
+ * into values needed for the anomaly result response.
+ */
 public class PMMLResult {
     private final boolean outlier;
     private final double decisionFunction;
--- a/src/main/java/org/opensearch/ad/ml/PMMLResult.java
+++ b/src/main/java/org/opensearch/ad/ml/PMMLResult.java
@@ -1,0 +1,81 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package org.opensearch.ad.ml;
+
+import java.util.Objects;
+
+// A reference for the format to expect for external call responses from the ML plugin.
+// This class is not used now. In PMML result transport action, we simply convert the results from the ML plugin
+// into values needed for the anomaly result response.
+public class PMMLResult {
+    private final boolean outlier;
+    private final double decisionFunction;
+
+    /**
+     * Constructor with all arguments.
+     *
+     * @param outlier whether the point is an outlier
+     * @param decisionFunction value that decides outlier, with negative means outlier, and positive means inlier
+     */
+    public PMMLResult(boolean outlier, double decisionFunction) {
+        this.outlier = outlier;
+        this.decisionFunction = decisionFunction;
+    }
+
+    /**
+     * Returns the outlier boolean.
+     *
+     * @return the outlier boolean
+     */
+    public boolean getOutlier() {
+        return outlier;
+    }
+
+    /**
+     * Returns the decision function double.
+     *
+     * @return the decision function double
+     */
+    public double getDecisionFunction() {
+        return decisionFunction;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        PMMLResult that = (PMMLResult) o;
+        return Objects.equals(this.outlier, that.outlier) && Objects.equals(this.decisionFunction, that.decisionFunction);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(outlier, decisionFunction);
+    }
+}
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -99,6 +99,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     public static final String CATEGORY_FIELD = "category_field";
     public static final String USER_FIELD = "user";
     public static final String DETECTOR_TYPE_FIELD = "detector_type";
+    public static final String ML_MODEL_ID_FIELD = "ml_model_id";
 
     private final String detectorId;
     private final Long version;
@@ -117,6 +118,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     private final List<String> categoryFields;
     private User user;
     private String detectorType;
+    private String mlModelId;
 
     /**
      * Constructor function.
@@ -173,6 +175,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
             lastUpdateTime,
             categoryFields,
             user,
+            null,
             null
         );
     }
@@ -195,6 +198,48 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         List<String> categoryFields,
         User user,
         String detectorType
+    ) {
+        this(
+            detectorId,
+            version,
+            name,
+            description,
+            timeField,
+            indices,
+            features,
+            filterQuery,
+            detectionInterval,
+            windowDelay,
+            shingleSize,
+            uiMetadata,
+            schemaVersion,
+            lastUpdateTime,
+            categoryFields,
+            user,
+            detectorType,
+            null
+        );
+    }
+
+    public AnomalyDetector(
+        String detectorId,
+        Long version,
+        String name,
+        String description,
+        String timeField,
+        List<String> indices,
+        List<Feature> features,
+        QueryBuilder filterQuery,
+        TimeConfiguration detectionInterval,
+        TimeConfiguration windowDelay,
+        Integer shingleSize,
+        Map<String, Object> uiMetadata,
+        Integer schemaVersion,
+        Instant lastUpdateTime,
+        List<String> categoryFields,
+        User user,
+        String detectorType,
+        String mlModelId
     ) {
         if (Strings.isBlank(name)) {
             throw new IllegalArgumentException("Detector name should be set");
@@ -240,6 +285,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         this.categoryFields = categoryFields;
         this.user = user;
         this.detectorType = detectorType;
+        this.mlModelId = mlModelId;
     }
 
     public AnomalyDetector(StreamInput input) throws IOException {
@@ -263,6 +309,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
             user = null;
         }
         detectorType = input.readOptionalString();
+        mlModelId = input.readOptionalString();
         if (input.readBoolean()) {
             this.uiMetadata = input.readMap();
         } else {
@@ -297,6 +344,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
             output.writeBoolean(false); // user does not exist
         }
         output.writeOptionalString(detectorType);
+        output.writeOptionalString(mlModelId);
         if (uiMetadata != null) {
             output.writeBoolean(true);
             output.writeMap(uiMetadata);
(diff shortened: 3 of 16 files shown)


build.gradle
```diff
@@ -529,6 +530,7 @@ List<String> jacocoExclusions = [
         'org.opensearch.ad.transport.GetAnomalyDetectorResponse',
         'org.opensearch.ad.transport.ADBatchAnomalyResultRequest',
         'org.opensearch.ad.transport.ADBatchTaskRemoteExecutionTransportAction',
+        'org.opensearch.ad.transport.PMMLResultTransportAction'
```
ylwu-amzn, line 533: Add some todo like line 512

src/main/java/org/opensearch/ad/ml/PMMLResult.java
```diff
@@ -0,0 +1,81 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ylwu-amzn, line 13: Don't need this old copyright header for new file.

src/main/java/org/opensearch/ad/ml/PMMLResult.java
```diff
@@ -0,0 +1,81 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package org.opensearch.ad.ml;
+
+import java.util.Objects;
+
+// A reference for the format to expect for external call responses from the ML plugin.
```
ylwu-amzn, line 31: Use javadoc for this class?
```
/** 
 * xxx
 */
```
https://stackoverflow.com/questions/211041/do-you-use-javadoc-for-every-method-you-write

> Plus, javadoc is way better than simple "//comment":
It is recognized by IDE and used to display a pop-up when you move your cursor on top of one of your - javadoc-ed - function. For instance, a constant - that is private static final variable -, should have a javadoc, especially when its value is not trivial. Case in point: regexp (its javadoc should includes the regexp in its non-escaped form, what is purpose is and a literal example matched by the regexp)
It can be parsed by external tools (like xdoclet)

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -263,6 +309,7 @@ public AnomalyDetector(StreamInput input) throws IOException {
             user = null;
         }
         detectorType = input.readOptionalString();
+        mlModelId = input.readOptionalString();
```
ylwu-amzn, line 312: Suggest to put new fields to the end of this method for backward compatibility. In old AD code, we don't have this field, the output stream will not have this optional string, so the input stream may fail to read.
Change to this and add to last line, similar for `writeTo`
```
if (input.available() > 0) {
    mlModelId = input.readOptionalString();
}
```

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -1059,6 +1083,104 @@ private void handleThresholdResult() {
...
+                    return;
+                }
+                // convert the results from PMML to AD specific format
+                double anomalyGrade = handleGrade(pmmlResult.getOutlier(), pmmlResult.getDecisionFunction());
+                double anomalyScore = handleScore(pmmlResult.getDecisionFunction());
+                // some values are dummy:
+                // 1. Confidence doesn't really change since the ML plugin uploaded models are immutable.
+                double confidence = 1;
+                // 2. PMML models don't have rcfTotalUpdates.
+                long rcfTotalUpdates = 1000;
```
ylwu-amzn, line 1146: Explain more about this hard code value. Current AD needs to calculate RCF model init progress with RCF total updates: 
`init_progress = RCF_total_updates / 128 `

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -1059,6 +1083,104 @@ private void handleThresholdResult() {
...
+            if (!outlier) {
+                return 0;
+            }
+            return -2 * decisionFunction;
+        }
+
+        // Decision function double from the ML plugin: [-0.5, 0.5]. Negative means anomalous.
+        // Anomaly score in AD is strictly positive. More positive means more anomalous. 1 is a typical cutoff.
+        // Thus, the values can be converted linearly: [-0.5, 0.5] to (0, 2].
+        private double handleScore(double decisionFunction) {
```
ylwu-amzn, line 1174: Anomaly score is not used on frontend, how about we keep the original `decisionFunction` value?

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
```diff
@@ -0,0 +1,148 @@
...
+        return features;
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(adID);
+        out.writeString(mlModelID);
+        out.writeVInt(featureNames.length);
+        for (String name : featureNames) {
```
ylwu-amzn, line 100: Use `out.writeOptionalArray` ?

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    protected void doExecute(Task task, PMMLResultRequest request, ActionListener<PMMLResultResponse> listener) {
+
+        if (adCircuitBreakerService.isOpen()) {
+            listener.onFailure(new LimitExceededException(request.getAdID(), CommonErrorMessages.MEMORY_CIRCUIT_BROKEN_ERR_MSG));
+            return;
+        }
+
+        try {
+            MachineLearningClient mlClient = new MachineLearningNodeClient(client);
```
ylwu-amzn, line 86: Can we use singleton of `MachineLearningClient` to avoid create new client for every request? You can refer to `AnomalyDetectorPlugin#createComponents`

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
```diff
@@ -0,0 +1,137 @@
...
+            sendPredictionRequest(mlClient, request, listener);
+        } catch (Exception e) {
+            LOG.error(e);
+            listener.onFailure(e);
+        }
+    }
+
+    // utilize the ml client layer to make prediction calls
```
ylwu-amzn, line 94: Use javadoc style? similar for other places

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
```diff
@@ -0,0 +1,137 @@
...
+    public void sendPredictionRequest(
+        MachineLearningClient mlClient,
+        PMMLResultRequest request,
+        ActionListener<PMMLResultResponse> listener
+    ) {
+        DataFrame inputData = loadDataFrame(request.getFeatureNames(), request.getFeatures());
+        mlClient.predict("pmml", null, inputData, request.getMlModelID(), ActionListener.wrap(response -> {
```
ylwu-amzn, line 101: Create constants for "pmml", so others can easily know and will not make mistake

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    // unload data frame to a map of result names and result values
+    public Map<String, Object> unloadDataFrame(DataFrame results) throws AnomalyDetectionException {
+        if (results == null || results.size() == 0) {
+            throw new AnomalyDetectionException("null or empty response from the ML plugin");
+        }
+        Map<String, Object> result = new HashMap<>();
+        ColumnMeta[] header = results.columnMetas();
+        Row prediction = results.getRow(0);
```
ylwu-amzn, line 128: How about adding DataFrame example, so others can understand this parsing logic easily?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -465,12 +516,15 @@ public static AnomalyDetector parse(
...
+                case ML_MODEL_ID_FIELD:
+                    mlModelId = parser.text();
+                    break;
                 default:
                     parser.skipChildren();
                     break;
             }
         }
-        return new AnomalyDetector(
+        AnomalyDetector anomalyDetector = new AnomalyDetector(
```
ohltyler, line 527: What is the need for this? Why can't we use the new constructor?

src/main/java/org/opensearch/ad/transport/PMMLResultAction.java
```diff
@@ -0,0 +1,40 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
ohltyler, line 25: nit: this header can be removed as well. Only need SPDX for new files

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.xcontent.ToXContentObject;
+import org.opensearch.common.xcontent.XContentBuilder;
+
+public class PMMLResultRequest extends ActionRequest implements ToXContentObject {
+    private String adID;
+    private String mlModelID;
+    private String[] featureNames;
+    private double[] features;
```
ohltyler, line 47: nit: how about making the name more verbose, like `featureValues`? Since you are already passing `featureNames`

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
```diff
@@ -0,0 +1,148 @@
...
+        out.writeVInt(features.length);
+        for (double feature : features) {
+            out.writeDouble(feature);
+        }
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        ActionRequestValidationException validationException = null;
+        if (adID == null || Strings.isEmpty(adID)) {
```
ohltyler, line 112: Rather than setting a `validationException` var and returning it at the end, can we just return within each `if` block? That way it will fail faster without having to traverse the other conditions. Also, it will let you utilize previous conditions to do less checks (e.g., don't need the null checks on line 124).

ohltyler: Could change to a cascading ifs too using `else if`

ohltyler:
One overall thought about the `PMML`-centric naming - the PMML format is more of an implementation detail, and could be more decoupled from what's actually being done, which is calling the ML plugin to fetch results from some custom model (that just happens to be _in_ .pmml format). For example,  `CustomModel` or `MLModel` could be a little more extensible and generalizable. AD shouldn't care what the format of the models are in the ML plugin. Consider future changes where the ML plugin changes to add or support some different format; then the naming conventions here wouldn't make much sense.

Additionally, the results shouldn't be coupled with the format - rather, they are more tied to the algorithm/model being used - in this case, `sklearn`'s `IsolationForest`s produce a `boolean` outlier and `double` decision function.

build.gradle
wqixian: Got it

src/main/java/org/opensearch/ad/ml/PMMLResult.java
wqixian: Understood

src/main/java/org/opensearch/ad/ml/PMMLResult.java
wqixian: Right, thanks!

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
wqixian: Definitely, thanks! For `writeTo`, I put `output.writeOptionalString(mlModelId);` at last line.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
wqixian: Got it

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
wqixian: During indexing anomaly results, I thought that non-positive values are considered to be erroneous (AnomalyDetectorJobRunner.java -- line 486), so I have to convert this value as well. Could you check and let me know if I misunderstood? Thanks

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
wqixian: I used this way to stick to how it's done in the RCF result request. I will dm you to discuss some details about this.

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
wqixian: Got it!

src/main/java/org/opensearch/ad/transport/PMMLResultAction.java
wqixian: Got it!

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
wqixian: If you check below, there is a line `anomalyDetector.setMlModelId(mlModelId);`. Because of the issue mentioned in this PR's "Description - Future tasks and improvements #5" and to support BWC, I think this is the best way to put it now.

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
wqixian: For these functions that I want to keep private (should I change to private?), is the casual format of comments good? Or always Javadoc style?

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
wqixian: I think this is just for testing use, the actual checks happen elsewhere, not in the request.

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
wqixian: Got it. Added a static final string at top of this transport action.

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
wqixian: Added in the comments!

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
wqixian: I have added this in this PR's "Description - Future tasks and improvements - 1"

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
ylwu-amzn: From our discussion, `writeOptionalArray` can only work for `Writeable` class, but String is not. It's ok to keep as is.

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
ylwu-amzn: Suggest to use java doc for method, no matter private or public. If this class should only not be used outside of the class, you should set it as private.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Sure, you can add some todo here. No need to fix right now considering your internship will end soon.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Good callout. You can add this to comment so others know clearly why we have to convert to positive scores

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
ylwu-amzn: Sure, please add TODO in code as well, so others know we should tune this part when read code.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
wqixian: Got it!

src/main/java/org/opensearch/ad/transport/PMMLResultTransportAction.java
wqixian: Got it! Will do

src/main/java/org/opensearch/ad/transport/PMMLResultRequest.java
ohltyler: Ok, just a minor thing for code readability & efficiency.

ohltyler:
I see the check is failing due to compilation failures because of missing ML-related dependencies. Ideally these new files & tests should only be executed when the ML plugin is available, and skipped when not. Any idea how notifications plugin handles this?

wqixian:
> I see the check is failing due to compilation failures because of missing ML-related dependencies. Ideally these new files & tests should only be executed when the ML plugin is available, and skipped when not. Any idea how notifications plugin handles this?

Since the ML is still a private project that is under development, the checks will be failing until we can publish ML to maven local as other packages. The notifications plugin is added in the common-utils package, so any plugin can import the interface by importing that package. Later we can either publish ML locally, or add a ML interface (as what notifications plugin is doing) in common-utils.

ylwu-amzn:
Close this PR as this is not under development now.

PR #187, opened 2021-08-23
Enable shingle in HCAD

kaituo:
### Description
Previously, we hardcoded HCAD (including real-time and historical) shingle size to be one due to time constraints when releasing HCAD. This PR enabled shingle. By default, the shingle size is eight but can be customized. I removed shingling-related code (even though I didn't allow shingling, but there is some half-finished shingling code) in HCAD as RCF can create shingles by itself (turned on via the internalShinglingEnabled configuration).

This PR also:
1. upgrade RCF to 2.0.1.
2. Removed MULTI_ENTITY_NUM_TREES with NUM_TREES as both HCAD and single-stream detectors have the same number of trees.
3.Fixed a bug in ModelManager.score method: I should compute anomaly grade and confidence before updating rcf and threshold model.

Testing done:
1. added unit tests.
2. Manual testing to verify HCAD single stream and historical workflow still work, and the RCF models have desired shingle size.
 
### Check List


src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -313,7 +261,88 @@ private void coldStart(
             return;
         }
 
-        nodeStateManager.getAnomalyDetector(detectorId, onGetDetector(modelId, entity, detectorId, modelState, listener));
+        boolean earlyExit = true;
```
sudiptoguha, line 264: This block seems to be just a shift of the existing block. Is that correct?

sudiptoguha (approved):

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: yeah, I moved the enclosing logic of getting a detector to the caller, thus the callee does not need to do that (and you saw a shift of existing block).

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -346,7 +380,12 @@ private void trainModelFromDataSegments(List<double[][]> dataPoints, Entity enti
             // vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize
             // overlapping x3, x4, and only store x5, x6.
             .shingleSize(shingleSize)
-            .build();
+            .internalShinglingEnabled(true);
```
ylwu-amzn, line 383: Does this mean we can just feed un-shingled data to RCF and RCF will do shingling inside?

kaituo: yes.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -587,7 +586,7 @@ public TimeConfiguration getWindowDelay() {
     }
 
     public Integer getShingleSize() {
-        return isMultientityDetector() ? DEFAULT_MULTI_ENTITY_SHINGLE : shingleSize;
+        return shingleSize;
```
ylwu-amzn, line 589: General question for supporting shingle size: do we need to re-train model if user changed shingle size? The old checkpoint is using different shingle size, can we restore model from checkpoint and feed feature data with different shingle size into it?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
ylwu-amzn: Got it, is this disabled by default? Not sure if this will impact current logic as we are feeding shingled data into RCF.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
kaituo: For existing detector, their shingle size is recorded in the config: https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/model/AnomalyDetector.java#L319

Example:

```
{
    "_id": "rjCrensBDqnenptucFOY",
    "_version": 1,
    "_primary_term": 1,
    "_seq_no": 0,
    "anomaly_detector": {
        "name": "Second Test Detector",
        "description": "ok rate",
        "time_field": "@timestamp",
        "indices": [
            "server-metrics"
        ],
        "filter_query": {
            "match_all": {
                "boost": 1.0
            }
        },
        "detection_interval": {
            "period": {
                "interval": 1,
                "unit": "Minutes"
            }
        },
        "window_delay": {
            "period": {
                "interval": 10,
                "unit": "Seconds"
            }
        },
        "shingle_size": 1,
        "schema_version": 2,
        "feature_attributes": [
            {
                "feature_id": "deny_max",
                "feature_name": "deny max",
                "feature_enabled": true,
                "aggregation_query": {
                    "deny_max": {
                        "max": {
                            "field": "deny"
                        }
                    }
                }
            }
        ],
        "ui_metadata": {
            "aabb": {
                "ab": "bb"
            }
        },
        "last_update_time": 1629850661013,
        "category_field": [
            "host"
        ]
    },
    "anomaly_detector_job": null,
    "realtime_detection_task": null,
    "historical_analysis_task": null
}
```

When reading it out, it is still one: https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/model/AnomalyDetector.java#L457

So the detector is still using shingle size 1.  No conversion or bwc for them.  

Is it correct?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: I am changing only HCAD real time to have internalShinglingEnabled. We fed unshingled data to old HCAD realtime RCF models.

For HCAD historical, the model is still using shingled data.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
ylwu-amzn: Oh, my question is about RCF model checkpoint. Before AD 1.1, the RCF model checkpoint is for model with shingle size 1. Not sure if we will delete all model checkpoint when user stop detector and update. If not, the model checkpoint may not be restored and used for new shingle size.

PR #195, opened 2021-08-29
block updating detector category field

ylwu-amzn:
### Description
If detector category field changed, frontend may not be able to render AD result for different detector types correctly. For example, if detector changed from HC to single entity detector, AD result page may show multiple anomaly result points on the same time point if there are multiple entities have anomaly results. If single-category HC changed category field from IP to error type, the AD result page may show both IP and error type in top entities list. That's confusing. So we decide to block updating detector category field.
 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
+++ b/src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
@@ -29,6 +29,7 @@ package org.opensearch.ad.rest.handler;
 import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG;
 import static org.opensearch.ad.model.ADTaskType.HISTORICAL_DETECTOR_TASK_TYPES;
 import static org.opensearch.ad.model.AnomalyDetector.ANOMALY_DETECTORS_INDEX;
+import static org.opensearch.ad.util.ParseUtils.listEqualsWithoutConsideringOrder;
 import static org.opensearch.ad.util.RestHandlerUtils.XCONTENT_WITH_TYPE;
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
@@ -38,7 +39,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
@@ -243,7 +243,7 @@ public class IndexAnomalyDetectorActionHandler {
             // If single-category HC changed category field from IP to error type, the AD result page may show both IP and error type
             // in top N entities list. That's confusing.
             // So we decide to block updating detector category field.
-            if (!Objects.equals(existingDetector.getCategoryField(), anomalyDetector.getCategoryField())) {
+            if (!listEqualsWithoutConsideringOrder(existingDetector.getCategoryField(), anomalyDetector.getCategoryField())) {
                 listener
                     .onFailure(new OpenSearchStatusException(CommonErrorMessages.CAN_NOT_CHANGE_CATEGORY_FIELD, RestStatus.BAD_REQUEST));
                 return;
--- a/src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
+++ b/src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
@@ -38,6 +38,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/ad/transport/ADBatchAnomalyResultRequest.java
```diff
@@ -70,8 +70,6 @@ public ActionRequestValidationException validate() {
         AnomalyDetector detector = adTask.getDetector();
         if (detector == null) {
             validationException = addValidationError("Detector can't be null", validationException);
-        } else if (detector.isRealTimeDetector()) {
-            validationException = addValidationError("Can't run batch task for realtime detector", validationException);
```
ohltyler, line 74: Why remove this validation?

ylwu-amzn: This method is deprecated `isRealTimeDetector`. We don't differentiate realtime and historical detector from AD1.1.

ohltyler: Oh right, I saw that was removed. Makes sense

ohltyler (dismissed):

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
```diff
@@ -236,32 +237,26 @@ private void onGetAnomalyDetectorResponse(GetResponse response) {
...
-            // We have separate flows for realtime and historical detector currently. User
-            // can't change detector from realtime to historical, vice versa.
-            if (existingDetector.isRealTimeDetector() != anomalyDetector.isRealTimeDetector()) {
+            // If detector category field changed, frontend may not be able to render AD result for different detector types correctly.
+            // For example, if detector changed from HC to single entity detector, AD result page may show multiple anomaly
+            // result points on the same time point if there are multiple entities have anomaly results.
+            // If single-category HC changed category field from IP to error type, the AD result page may show both IP and error type
+            // in top N entities list. That's confusing.
+            // So we decide to block updating detector category field.
+            if (!Objects.equals(existingDetector.getCategoryField(), anomalyDetector.getCategoryField())) {
```
kaituo, line 246: Seems Object.equals cannot be used here.  Example:

```
jshell> List<String> a = new ArrayList<>()
a ==> []

jshell> a.add("a1")
$2 ==> true

jshell> a.add("a2")
$3 ==> true

jshell> List<String> b = new ArrayList<>()
b ==> []

jshell> b.add("a2")
$5 ==> true

jshell> b.add("a1")
$6 ==> true

jshell> Objects.equals(a, b)
$7 ==> false
```

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -181,6 +181,50 @@ public void testCreateAnomalyDetector() throws Exception {
...
+        AnomalyDetector detector = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
+        String indexName = detector.getIndices().get(0);
+        TestHelpers.createIndex(client(), indexName, TestHelpers.toHttpEntity("{\"name\": \"test\"}"));
+        Response response = TestHelpers
+            .makeRequest(client(), "POST", TestHelpers.AD_BASE_DETECTORS_URI, ImmutableMap.of(), TestHelpers.toHttpEntity(detector), null);
+        assertEquals("Create anomaly detector failed", RestStatus.CREATED, TestHelpers.restStatus(response));
+        Map<String, Object> responseMap = entityAsMap(response);
+        String id = (String) responseMap.get("_id");
+        System.out.println(id);
+        // long version = (long) responseMap.get("_version");
```
kaituo, line 194: remove debug code

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
ylwu-amzn: That's by design. Block any changes on category fields including the order to avoid any potential issues like we depends on the order to process differently in future. Any concern if we don't allow user change the category field order?

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
ylwu-amzn: sure

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
ylwu-amzn: As AD backend doesn't rely on the order of category field, I'm ok to change to comparing list ignoring the order to give some flexibility to API user.

kaituo (approved):

src/main/java/org/opensearch/ad/util/ParseUtils.java
```diff
@@ -702,4 +705,24 @@ public static SearchSourceBuilder batchFeatureQuery(
...
+        Set<String> set1 = new HashSet<>();
+        Set<String> set2 = new HashSet<>();
+        if (list1 != null) {
+            set1.addAll(list1);
+        }
+        if (list2 != null) {
+            set2.addAll(list2);
+        }
+        return Objects.equals(set1, set2);
+    }
```
ohltyler, line 727: Seems fine to do this, but what about just sorting with `Collections.sort()` and then using the regular `Object.equals()`, rather than having to maintain this helper function?

ylwu-amzn: Here I want to just check if the list equals or not rather than change the lists themselves. `Collections.sort()` is one possible option, but it will change the item order in place. That means after executing this comparing method, the item order will change in original list. Another concern is if the list passed in is immutable, `Connections.sort()` will throw exception.

ohltyler: Got it, makes sense.

ohltyler (approved):

PR #201, opened 2021-09-02
reset realtime task state

ylwu-amzn:

### Description
Reset realtime task state when get latest detector task. If job is disabled, we should set realtime task state as stopped.

 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
+++ b/src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
@@ -66,12 +66,17 @@ import com.google.common.collect.ImmutableList;
 
 public class ADTaskCacheManager {
     private final Logger logger = LogManager.getLogger(ADTaskCacheManager.class);
-    private final Map<String, ADBatchTaskCache> taskCaches;
+
     private volatile Integer maxAdBatchTaskPerNode;
     private volatile Integer maxCachedDeletedTask;
     private final MemoryTracker memoryTracker;
     private final int numberSize = 8;
     public static final int TASK_RETRY_LIMIT = 3;
+
+    // This field is to record all batch tasks. Both single entity detector task
+    // and HC entity task will be cached in this field.
+    // Key: task id
+    private final Map<String, ADBatchTaskCache> taskCaches;
 
     // We use this field to record all detector level tasks which running on the
     // coordinating node to resolve race condition. We will check if
@@ -80,6 +85,7 @@ public class ADTaskCacheManager {
     // task in cache. For other tasks, we find the detector id exists,
     // that means there is already one task running for this detector,
     // so we will reject the task.
+    // Key: detector id; Value: detector level task id
     private Map<String, String> detectorTasks;
 
     // Use this field to cache all HC tasks. Key is detector id
@@ -91,7 +97,7 @@ public class ADTaskCacheManager {
 
     // This field is to cache all realtime tasks. Key is detector id
     private Map<String, ADRealtimeTaskCache> realtimeTaskCaches;
-    // This field is to cache all detectors' task slot and task lane limit
+    // This field is to cache all detectors' task slot and task lane limit. Key is detector id
     private Map<String, ADTaskSlotLimit> detectorTaskSlotLimit;
 
     /**
@@ -399,6 +405,18 @@ public class ADTaskCacheManager {
             taskCache.clear();
             hcTaskCaches.remove(detectorId);
         }
+        List<String> tasksOfDetector = getTasksOfDetector(detectorId);
+        for (String taskId : tasksOfDetector) {
+            remove(taskId);
+        }
+        if (tasksOfDetector.size() > 0) {
+            logger
+                .warn(
+                    "Removed historical AD task from cache for detector {}, taskId: {}",
+                    detectorId,
+                    Arrays.toString(tasksOfDetector.toArray(new String[0]))
+                );
+        }
         if (detectorTasks.containsKey(detectorId)) {
             detectorTasks.remove(detectorId);
             logger.info("Removed detector from AD task coordinating node cache, detectorId: " + detectorId);
@@ -681,6 +699,7 @@ public class ADTaskCacheManager {
         if (adTaskSlotLimit != null && delta > 0) {
             int newTaskSlots = taskSlots - delta;
             if (newTaskSlots > 0) {
+                taskSlots = newTaskSlots;
                 logger.info("Scale down task slots of detector {} from {} to {}", detectorId, taskSlots, newTaskSlots);
                 adTaskSlotLimit.setDetectorTaskSlots(newTaskSlots);
                 return newTaskSlots;
--- a/src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
+++ b/src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
@@ -381,7 +381,7 @@ public class ADTaskCacheManager {
      *
      * @param detectorId detector id
      */
-    public void removeDetector(String detectorId) {
+    public void removeHistoricalTaskCache(String detectorId) {
         ADHCBatchTaskCache taskCache = hcTaskCaches.get(detectorId);
         if (taskCache != null) {
             // this will happen only on coordinating node. When worker nodes left,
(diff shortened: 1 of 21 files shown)


src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1132,52 +1200,35 @@ private void stopHistoricalAnalysis(
...
-        }, e -> logger.error("Failed to update task state as STOPPED for task " + taskId, e)));
+        cleanDetectorCache(adTask, transportService, () -> {
+            String taskId = adTask.getTaskId();
+            Map<String, Object> updatedFields = ImmutableMap.of(STATE_FIELD, ADTaskState.STOPPED.name());
+            updateADTask(taskId, updatedFields, ActionListener.wrap(r -> {
+                adTask.setState(ADTaskState.STOPPED.name());
+                if (function != null) {
+                    function.execute();
+                }
+                if (ADTaskType.HISTORICAL_HC_DETECTOR.name().equals(adTask.getTaskType())) {
```
ohltyler, line 1223: Why is only the historical HC task type considered here? What about RT HC entity tasks?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1099,6 +1150,23 @@ private void resetHistoricalDetectorTaskState(ADTask adTask, AnomalyDetectorFunc
...
 
+    private boolean isTaskStopped(String taskId, AnomalyDetector detector, ADTaskProfile taskProfile) {
+        boolean taskStopped = false;
+        String detectorId = detector.getDetectorId();
+        if (taskProfile == null || !Objects.equals(taskId, taskProfile.getTaskId())) {
+            logger.debug("AD task not found for task {} detector {}", taskId, detectorId);
+            // If no node is running this task, reset it as STOPPED.
+            taskStopped = true;
+        } else if (!detector.isMultientityDetector() && taskProfile.getNodeId() == null) {
+            logger.debug("AD task not running for single flow detector {}, task {}", detectorId, taskId);
```
kaituo, line 1161: single entity or stream to be consistent?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1020,49 +1019,101 @@ public void stopDetector(
         }));
     }
 
-    private void resetLatestHistoricalDetectorTaskState(
+    private <T> void resetLatestHistoricalDetectorTaskState(
```
kaituo, line 1022: this method may also reset realtime tasks state.  Could you change the name?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1020,49 +1019,101 @@ public void stopDetector(
...
 
-    private void resetHistoricalDetectorTaskState(ADTask adTask, AnomalyDetectorFunction function, TransportService transportService) {
+    private <T> void resetRealtimeDetectorTaskState(
+        List<ADTask> runningRealtimeTasks,
+        AnomalyDetectorFunction function,
+        TransportService transportService,
+        ActionListener<T> listener
+    ) {
+        if (isNullOrEmpty(runningRealtimeTasks)) {
+            function.execute();
```
kaituo, line 1057: Does function guarantee to return using listener?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: For realtime anomaly detection, we only create detector level task, no entity level realtime task.

kaituo: could you add a comment for this?

ylwu-amzn: sure

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: sure, will change to single entity

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: good catch, will change

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: Yes, the listener will return in code block wrapped in function.

kaituo: Checked the call graph, this function will be called by ADTaskManager.getAndExecuteOnLatestADTask.  Since getAndExecuteOnLatestADTask is public, users can call it anyway they want and may miss the listener return.

I checked a few callers of ADTaskManager.getAndExecuteOnLatestADTask.  Does https://github.com/ylwu-amzn/anomaly-detection-2/blob/reset_realtime_task_state/src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java#L257 return listener?

ylwu-amzn: listener is one `field` of class `IndexAnomalyDetectorActionHandler`, I see listener returned.

kaituo: I see.  Can you add comments to remind the caller to return listener?

ylwu-amzn: sure, added

ohltyler (approved):

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -681,6 +699,7 @@ public synchronized int scaleDownHCDetectorTaskSlots(String detectorId, int delt
         if (adTaskSlotLimit != null && delta > 0) {
             int newTaskSlots = taskSlots - delta;
             if (newTaskSlots > 0) {
+                taskSlots = newTaskSlots;
```
kaituo, line 702: why do we need this line? after this, newTaskSlots and taskSlots are equal and the log the next time does not seem to make sense.

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -66,20 +66,26 @@
...
     private volatile Integer maxAdBatchTaskPerNode;
     private volatile Integer maxCachedDeletedTask;
     private final MemoryTracker memoryTracker;
     private final int numberSize = 8;
     public static final int TASK_RETRY_LIMIT = 3;
 
+    // This field is to record all batch tasks. Both single entity detector task
+    // and HC entity task will be cached in this field.
+    // Key: task id
+    private final Map<String, ADBatchTaskCache> taskCaches;
```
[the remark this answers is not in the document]
ylwu-amzn: sure

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: This is to return the new task slots. Yes ,the log will not show correct old task slots, will move this line after the log line 703.

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -89,7 +89,7 @@
     private Map<String, String> detectorTasks;
 
     // Use this field to cache all HC tasks. Key is detector id
-    private Map<String, ADHCBatchTaskCache> hcTaskCaches;
+    private Map<String, ADHCBatchTaskCache> hcBatchTaskCaches;
```
kaituo, line 92: What are the differences between hcBatchTaskCaches and batchTaskCaches? Do they have overlap contents? Is it that hcBatchTaskCaches have detector level task cache, while batchTaskCaches has entity level task cache (if it is a HC detector) and detector level task cache (if it is single-stream detector)? if yes, then I am more confused. Why cannot we combine these two maps?

ylwu-amzn: Your understanding is correct. `hcBatchTaskCaches`: cache HC detector level information on coordinating node, will add more comments. I think the comment for `batchTaskCaches` is enough, you can read and suggest more if the comment is still not clear.

They don't have overlap contents. Combine these two maps will mix the coordinating node and worker node cache together. I don't prefer that way now. If no big concern or cons , how about we keep current design now?

kaituo (approved):

ohltyler (approved):

PR #196, opened 2021-09-01
[BACKPORT] [1.x] Backport all changes to 1.x

ohltyler:
### Description

This PR cherry-picks all of the changes from `main` branch to `1.x` branch. This is ok, since we currently have nothing in `main` that shouldn't go in the next `1.x` release (which will be `1.1`). Confirmed that there are no differences between `1.x` and `main` via `git diff`

When merging, will not squash, so that the commit history is retained.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #202, opened 2021-09-02
Fix flaky test and log level/msgs and enable auto-expand replication

kaituo:
### Description
This PR (hopefully, as I cannot reproduce the failure locally) fixed flaky tests in MultiEntityResultTests. The tests are flaky, maybe because we expect two pages in our pagination, but we may create more than two pages due to a race condition. Please read comments in MultiEntityResultTests for detail.

This PR also changes the log level of the updating real-time task log from info to debug. We don't need info as Opensearch prints the log repeatedly in each interval. Also, I changed the log message in ADTaskManager to match what the relevant code does.

This PR also enables auto-expand replication for AD job indexes. The job scheduler puts both primary and replica shards in the hash ring. Enabling auto-expand the number of replicas based on the number of data nodes (up to 20) in the cluster so that each node can become a coordinating node. Enabling auto-expanding is useful when customers scale out their cluster so that we can do adaptive scaling accordingly. Also, this PR changed the primary number of shards of the AD job index to 1 as the AD job index is small.

Testing done:
1. Checked that the AD job index setting change is effective and won't negatively impact normal e2e workflow.
 
### Issues Resolved
Flaky test run: https://github.com/opensearch-project/anomaly-detection/pull/196/checks?check_run_id=3487905317
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
+++ b/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
@@ -43,6 +43,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -60,6 +61,10 @@ import org.opensearch.action.admin.indices.create.CreateIndexResponse;
 import org.opensearch.action.admin.indices.delete.DeleteIndexRequest;
 import org.opensearch.action.admin.indices.mapping.put.PutMappingRequest;
 import org.opensearch.action.admin.indices.rollover.RolloverRequest;
+import org.opensearch.action.admin.indices.settings.get.GetSettingsAction;
+import org.opensearch.action.admin.indices.settings.get.GetSettingsRequest;
+import org.opensearch.action.admin.indices.settings.get.GetSettingsResponse;
+import org.opensearch.action.admin.indices.settings.put.UpdateSettingsRequest;
 import org.opensearch.action.support.GroupedActionListener;
 import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.ad.common.exception.EndRunException;
@@ -103,11 +108,17 @@ public class AnomalyDetectionIndices implements LocalNodeMasterListener {
     // The index name pattern to query all AD result, history and current AD result
     public static final String ALL_AD_RESULTS_INDEX_PATTERN = ".opendistro-anomaly-results*";
 
+    // minimum shards of the job index
+    public static int minJobIndexReplicas = 1;
+    // maximum shards of the job index
+    public static int maxJobIndexReplicas = 20;
+
     // package private for testing
     static final String META = "_meta";
     private static final String SCHEMA_VERSION = "schema_version";
 
     private ClusterService clusterService;
+    private final Client client;
     private final AdminClient adminClient;
     private final ThreadPool threadPool;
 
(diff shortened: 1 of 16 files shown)


ohltyler (approved):

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -874,7 +874,7 @@ public void stopDetector(
 
                 consumer.accept(detector);
             } catch (Exception e) {
-                String message = "Failed to start anomaly detector " + detectorId;
+                String message = "Failed to get anomaly detector " + detectorId;
```
ylwu-amzn, line 877: How about change the error message as "Failed to parse anomaly detector "

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -406,7 +406,22 @@ public void initAnomalyDetectorJobIndex(ActionListener<CreateIndexResponse> acti
         try {
             CreateIndexRequest request = new CreateIndexRequest(AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX)
                 .mapping(AnomalyDetector.TYPE, getAnomalyDetectorJobMappings(), XContentType.JSON);
-            choosePrimaryShards(request);
+            request
+                .settings(
+                    Settings
+                        .builder()
+                        // AD job index is small. 1 primary shard is enough
+                        .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
```
ylwu-amzn, line 414: Should we or is it possible to change the old AD job index setting to this new setting?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
kaituo: changed

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
kaituo: You meant changing the number of primary shards? if yes, OpenSearch does not support it.

ylwu-amzn: Oh, I mean the auto expand replica, so we can leverage more data nodes to run realtime job.

kaituo: changed.

kaituo:
@ohltyler The build failed due to flaky test org.opensearch.ad.rest.HistoricalAnalysisRestApiIT.  Yaliang fixed it in another PR. Will leave it as it is.

ohltyler (dismissed):

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -603,34 +630,99 @@ private void deleteIndexIteration(String[] toDelete) {
...
+            switch (adIndex) {
+                case JOB:
+                    updateJobIndexSettingIfNecessary(indexStates.computeIfAbsent(adIndex, IndexState::new), conglomerateListeneer);
+                    break;
+                default:
+                    // we don't have settings to update for other indices
+                    IndexState indexState = indexStates.computeIfAbsent(adIndex, IndexState::new);
+                    indexState.settingUpToDate = true;
+                    logger.info(new ParameterizedMessage("Mark [{}]'s setting up-to-date", adIndex.getIndexName()));
+                    conglomerateListeneer.onResponse(null);
```
ylwu-amzn, line 693: `delegateListeneer` in this method is `GroupedActionListener` with group size 2. 
`conglomerateListeneer` is `GroupedActionListener` with group size `updates.size()`. 
Do these two group listeners have same group size?

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadWorker.java
```diff
@@ -168,6 +170,10 @@ protected MultiGetRequest toBatchRequest(List<EntityFeatureRequest> toProcess) {
             // lazy init since we don't expect retryable requests to happen often
             Set<String> retryableRequests = null;
             Set<String> notFoundModels = null;
+            boolean printedUnexpectedFailure = false;
+            // contain requests that we will set the detector's exception to
+            // EndRunException (stop now = false)
+            Map<String, Exception> stopDetectorRequests = null;
```
ylwu-amzn, line 176: When will we clean up the exception cache? Add some comments about the key?

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
ylwu-amzn: Oh, never mind, misunderstood the group size, should be ok, as the `conglomerateListeneer` will only call `delegateListeneer` once all updates returns.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -786,8 +897,96 @@ public int getSchemaVersion(ADIndex index) {
      * @param index Index metadata
      * @return Whether the given index's mapping is up-to-date
      */
-    public Boolean isUpdated(ADIndex index) {
+    public Boolean isMappingUpdated(ADIndex index) {
```
ylwu-amzn, line 900: Seems this method is not being used now?

src/main/java/org/opensearch/ad/ratelimit/CheckpointReadWorker.java
ylwu-amzn: The code folded makes me think this is some class field. Ignore this comment.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
kaituo: yeah, removed

ohltyler (approved):

ylwu-amzn (approved):

PR #214, opened 2021-09-14
Update versions for build dependencies

ohltyler:

### Description
Now that core OpenSearch has `1.1` branch cut, we should switch from using the snapshot artifacts (with version `1.1.0-SNAPSHOT`) to regular artifacts (with version `1.1.0`), for both core and other dependencies (job-scheduler, common-utils). We add the `build.snapshots=false` arg for all locally published artifacts, so that we have version consistency in `build.gradle`, where all dependency versions are now `1.1.0`.
 
### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -47,7 +47,7 @@ jobs:
           path: common-utils
       - name: Build common-utils
         working-directory: ./common-utils
-        run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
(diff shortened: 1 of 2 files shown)


ohltyler:
The build failed due to the flaky test, but the dependency builds worked as expected.

dblock (changes requested):
IMO, Your CI should/could continue building -SNAPSHOT builds. We will never want to use these in a release and we're building the whole distribution from the top in Jenkins infra.

We need to clarify this somewhere in opensearch-build, too.

ohltyler:
Got it, will close this out. Will keep using 1.1 snapshots, and will bump to 1.2 snapshots when we bump the plugin version to 1.2.

PR #218, opened 2021-09-17
Removing default integtest.sh (1.x).

dblock:
Coming from https://github.com/opensearch-project/opensearch-build/issues/497, removing default integtest.sh.


ylwu-amzn:
> @ylwu-amzn Is 1.x branch supposed to be building? It's failing CI here.

It works, maybe your branch doesn't have latest code. How about merging main branch ?

PR #221, opened 2021-09-23
Replace rcf/kll model with ThresholdedRandomCutForest

kaituo:

### Description
This PR does the following things:
1. added the new binaries of RCF library.
2. removed various places of kll threshold model.  Only keep it in ModelManager and CheckpointDao for backward compatibility purposes.  Kll threshold model is still needed because 1)old nodes can send threshold result messages to new nodes; 2)we need to know how to convert from kll threshold model and rcf model to ThresholdedRandomCutForest.
3. Removed the constructor of EntityModel that takes rcf and threshold models as we won't run old rcf/kll models in HCAD and EntityModel is only used in HCAD.  
4. Remove model partitioning logic in single-stream workflow.  We have already disabled the code path before.  Removing the code for easier maintenance.
5. Change to use ThresholdedRandomCutForest in preview API.

Testing done:
1. Manual bwc testing
2. real-time single stream and HCAD detectors testing
3. Make sure all unit/integ/system tests pass.
 
### Check List


src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -707,51 +558,16 @@ public void maintenance(ActionListener<Void> listener) {
             .compact(true)
             .precision(Precision.FLOAT_32)
             .boundingBoxCacheFraction(AnomalyDetectorSettings.BATCH_BOUNDING_BOX_CACHE_RATIO)
-            // same with dimension for opportunistic memory saving
-            .shingleSize(shingleSize)
+            // for external shingling, rcf does not recognize shingle
+            .shingleSize(1)
```
wnbts, line 562: does this affect singles and multis?

kaituo: This affects single-stream real time detectors.  HCAD models' configuration is in EntityColdStarter.

wnbts: is this getPreviewResults called by multis?

kaituo: Sorry, I was looking at the wrong places.  Yes, this affects both single stream/HCAD preview.

src/main/java/org/opensearch/ad/ml/EntityModel.java
```diff
@@ -32,24 +32,14 @@
...
 
-    public EntityModel(Entity entity, Queue<double[]> samples, RandomCutForest rcf, ThresholdingModel threshold) {
-        this.entity = entity;
-        this.samples = samples;
-        this.rcf = rcf;
-        this.threshold = threshold;
-    }
-
     /**
      * Constructor with ERCF.
```
sudiptoguha, line 44: Change to tRCF? Not urgent/not important right now.

kaituo: changed.

src/main/java/org/opensearch/ad/ml/ModelManager.java
sudiptoguha: OK. So this shingleSize = 1 for the tRCF is a signal that shingling is external. So we are using external shingling in the HCAD preview as well? We are using internal shingling in HCAD right?

sudiptoguha: (information question so that we are all on same page)

kaituo: yes, we are using external shingling in HCAD preview, but internal shingling in HCAD real time detectors.

sudiptoguha (approved):

wnbts (approved):

src/main/java/org/opensearch/ad/ml/ModelManager.java
sudiptoguha: Wait this shingleSize will be passed to the forest, so we would have shingleSize as 1 again.

PR #223, opened 2021-09-23
Remove model partition code and refactor model id generation

kaituo:

### Description
This PR removed model partitioner as we don’t partition models any more. Also, the PR moved model id logic from ModelPartitioner/ModelManager to ModelIdMapper so we have a central place to single-stream detectors’ id generation.

Testing done:
1. Manually tested single-stream detectors run fine.
 
### Check List


src/main/java/org/opensearch/ad/ml/ModelIdMapper.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package org.opensearch.ad.ml;
+
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Utilities to map between single-stream models and ids.  We will have circuilar
```
ylwu-amzn, line 19: This class only used for single-flow detector models? If that's the case, how about rename the class as like "SingleFlowModelIdMapper" ?

kaituo: renamed

src/main/java/org/opensearch/ad/cluster/ADClusterEventListener.java
```diff
@@ -143,7 +144,7 @@ public void clusterChanged(ClusterChangedEvent event) {
                                     LOG.info(REMOVE_MODEL_MSG + " {}", modelId);
                                     modelManager
                                         .stopModel(
-                                            modelManager.getDetectorIdForModelId(modelId),
+                                            SingleStreamModelIdMapper.getDetectorIdForModelId(modelId),
```
ohltyler, line 147: Looks like this method was moved from `ModelManager` to `SingleStreamModelIdMapper`, but don't see it removed in this PR. Is it removed in a different PR?

ohltyler: And can you confirm that the usages of the methods moved to `SingleStreamModelIdMapper` have been updated in a different PR?

src/main/java/org/opensearch/ad/ml/SingleStreamModelIdMapper.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package org.opensearch.ad.ml;
+
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Utilities to map between single-stream models and ids.  We will have circuilar
```
ohltyler, line 19: nit: `circuilar` -> `circular`

src/main/java/org/opensearch/ad/cluster/ADClusterEventListener.java
kaituo: yes, it is updated in https://github.com/opensearch-project/anomaly-detection/pull/221

ylwu-amzn (approved):

ohltyler (approved):

PR #224, opened 2021-09-23
Deal with backward-compatibility in Checkpoint

kaituo:

### Description
As part of integrating with ThresholdedRandomCutForest, this PR deals with backward-compatibility in Checkpoint.  Specifically, if ThresholdedRandomCutForest is present at the checkpoint, we load it directly.  Otherwise, we need to look for RandomCutForest and Kll threshold models and convert them to ThresholdedRandomCutForest. The model id is still the same so that we can still find previous checkpoints. For single-stream detectors, the model is detector_id _model_rcf_0.  For HCAD, the model id is detector_id _model_entity_value.

Testing done:
1. Manual BWC testing.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/org/opensearch/ad/ml/CheckpointDao.java
@@ -27,7 +27,6 @@
 package org.opensearch.ad.ml;
 
 import java.io.IOException;
-import java.io.StringReader;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.time.Instant;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -305,21 +306,22 @@ private void saveModelCheckpointAsync(Map<String, Object> source, String modelId
...
+                        serializedModel.isPresent() ? serializedModel.get().length() : 0
                     )
                 );
             return source;
         }
         String detectorId = modelState.getDetectorId();
         source.put(DETECTOR_ID, detectorId);
-        source.put(FIELD_MODELV2, serializedModel);
+        // we cannot pass Optional as OpenSearch does not know how to serialize an Optional value
+        source.put(FIELD_MODELV2, serializedModel.get());
```
wnbts, line 324: when an optional is empty, get returns an exception.

wnbts (approved):

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
kaituo: yes, we returned early if that the optional is empty.  Check line 310~320.

wnbts: ok. that's good.

sudiptoguha (approved):

PR #225, opened 2021-09-23
add BWC test cases; fix flaky test

ylwu-amzn:
### Description
Add more BWC IT test cases. As BWC test framework doesn't provide way to specify which node to send request to, just test happy workflow here. Will add more test cases for edge cases in future.
 
**[Note]**: This PR is not to cover all happy/edge cases for AD BWC. That's just what I have written currently. Definitely we will/need to add more test cases to improve coverage. 

### Check List

### DIFF
--- a/src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
+++ b/src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
@@ -233,6 +233,7 @@ public abstract class HistoricalAnalysisRestTestCase extends AnomalyDetectorRest
         return adTaskProfile;
     }
 
+    // TODO: change response to pair
     protected List<Object> waitUntilTaskDone(String detectorId) throws InterruptedException {
         return waitUntilTaskReachState(detectorId, TestHelpers.HISTORICAL_ANALYSIS_DONE_STATS);
     }
--- a/src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
+++ b/src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
@@ -31,6 +31,7 @@ import static org.opensearch.ad.settings.AnomalyDetectorSettings.BATCH_TASK_PIEC
 import java.io.IOException;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -57,6 +58,7 @@ import com.google.common.collect.ImmutableMap;
 
 public abstract class HistoricalAnalysisRestTestCase extends AnomalyDetectorRestTestCase {
 
+    public static final int MAX_RETRY_TIMES = 200;
     protected String historicalAnalysisTestIndex = "test_historical_analysis_data";
     protected int detectionIntervalInMinutes = 1;
     protected int categoryFieldDocCount = 2;
@@ -231,14 +233,16 @@ public abstract class HistoricalAnalysisRestTestCase extends AnomalyDetectorRest
         return adTaskProfile;
     }
 
-    protected ADTaskProfile waitUntilTaskDone(String detectorId) throws InterruptedException {
+    protected List<Object> waitUntilTaskDone(String detectorId) throws InterruptedException {
         return waitUntilTaskReachState(detectorId, TestHelpers.HISTORICAL_ANALYSIS_DONE_STATS);
     }
 
-    protected ADTaskProfile waitUntilTaskReachState(String detectorId, Set<String> targetStates) throws InterruptedException {
+    protected List<Object> waitUntilTaskReachState(String detectorId, Set<String> targetStates) throws InterruptedException {
+        List<Object> results = new ArrayList<>();
         int i = 0;
         ADTaskProfile adTaskProfile = null;
-        while ((adTaskProfile == null || !targetStates.contains(adTaskProfile.getAdTask().getState())) && i < 60) {
+        // Increase retryTimes if some task can't reach done state
+        while ((adTaskProfile == null || !targetStates.contains(adTaskProfile.getAdTask().getState())) && i < MAX_RETRY_TIMES) {
             try {
                 adTaskProfile = getADTaskProfile(detectorId);
             } catch (Exception e) {
@@ -249,6 +253,8 @@ public abstract class HistoricalAnalysisRestTestCase extends AnomalyDetectorRest
             i++;
         }
         assertNotNull(adTaskProfile);
-        return adTaskProfile;
+        results.add(adTaskProfile);
+        results.add(i);
+        return results;
     }
 }
--- a/src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
+++ b/src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
@@ -60,6 +60,7 @@ public class ADBackwardsCompatibilityIT extends OpenSearchRestTestCase {
 
     private static final ClusterType CLUSTER_TYPE = ClusterType.parse(System.getProperty("tests.rest.bwcsuite"));
     private static final String CLUSTER_NAME = System.getProperty("tests.clustername");
+    private static final String MIXED_CLUSTER_TEST_ROUND = System.getProperty("tests.rest.bwcsuite_round");
     private String dataIndexName = "test_data_for_ad_plugin";
     private int detectionIntervalInMinutes = 1;
     private int windowDelayIntervalInMinutes = 1;
@@ -154,13 +155,13 @@ public class ADBackwardsCompatibilityIT extends OpenSearchRestTestCase {
                     // Create single entity historical detector and start historical analysis
                     createHistoricalAnomalyDetectorsAndStart();
 
-                    // Verify cluster has 3 detectors now
+                    // Verify cluster has correct number of detectors now
                     verifyAnomalyDetectorCount(TestHelpers.LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI, 3);
-                    // Verify cluster has 3 detectors with search detector API
+                    // Verify cluster has correct number of detectors with search detector API
                     assertEquals(3, countDetectors(client(), null));
 
                     // Verify all realtime detectors are running realtime job
-                    verifyRealtimeJobRunning();
+                    verifyAllRealtimeJobsRunning();
                     break;
                 case MIXED:
                     // We have no way to specify whether send request to old node or new node now.
@@ -168,15 +169,38 @@ public class ADBackwardsCompatibilityIT extends OpenSearchRestTestCase {
                     Assert.assertTrue(pluginNames.contains("opensearch-anomaly-detection"));
                     Assert.assertTrue(pluginNames.contains("opensearch-job-scheduler"));
 
-                    // Verify cluster has 3 detectors now
-                    verifyAnomalyDetectorCount(TestHelpers.LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI, 3);
-                    // Verify cluster has 3 detectors with search detector API
-                    assertEquals(3, countDetectors(client(), null));
+                    // Create single entity detector and start realtime job
+                    createRealtimeAnomalyDetectorsAndStart(SINGLE_ENTITY_DETECTOR);
+
+                    // Create single category HC detector and start realtime job
+                    createRealtimeAnomalyDetectorsAndStart(SINGLE_CATEGORY_HC_DETECTOR);
+
+                    int mixedClusterTestRound = 0;
+                    switch (MIXED_CLUSTER_TEST_ROUND) {
+                        case "first":
+                            mixedClusterTestRound = 1;
+                            break;
+                        case "second":
+                            mixedClusterTestRound = 2;
+                            break;
+                        case "third":
+                            mixedClusterTestRound = 3;
+                            break;
+                        default:
+                            break;
+                    }
+                    int numberOfDetector = 3 + 2 * mixedClusterTestRound;
+                    // Verify cluster has correct number of detectors now
+                    verifyAnomalyDetectorCount(TestHelpers.LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI, numberOfDetector);
+                    // Verify cluster has correct number of detectors with search detector API
+                    assertEquals(numberOfDetector, countDetectors(client(), null));
 
                     // Verify all realtime detectors are running realtime job
-                    verifyRealtimeJobRunning();
+                    verifyAllRealtimeJobsRunning();
                     break;
                 case UPGRADED:
+                    // This branch is for testing full upgraded cluster. That means all nodes in cluster are running
+                    // latest AD version.
                     Assert.assertTrue(pluginNames.contains("opensearch-anomaly-detection"));
                     Assert.assertTrue(pluginNames.contains("opensearch-job-scheduler"));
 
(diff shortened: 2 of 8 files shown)


src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +131,176 @@ public void testBackwardsCompatibility() throws Exception {
...
                     break;
                 case UPGRADED:
                     Assert.assertTrue(pluginNames.contains("opensearch-anomaly-detection"));
                     Assert.assertTrue(pluginNames.contains("opensearch-job-scheduler"));
-                    verifyAnomalyDetector(TestHelpers.AD_BASE_DETECTORS_URI);
+
+                    Map<ADRestTestUtils.DetectorType, String> detectors = new HashMap<>();
+
+                    // Create single entity detector and start realtime job
+                    List<String> singleEntityDetectorResults = createRealtimeAnomalyDetectorsAndStart(SINGLE_ENTITY_DETECTOR);
```
VachaShah, line 186: The `createRealtimeAnomalyDetectorsAndStart` ultimately calls the `startAnomalyDetector` which uses the `LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI` to start detector. If the cluster is `UPGRADED`, should we not use the `AD_BASE_DETECTORS_URI` instead?

VachaShah: And if this works for upgraded cluster, wouldn't this work for mixed cluster as well where you can send request to `LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI` for creating new detectors?

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
```diff
@@ -231,14 +233,16 @@ protected ADTaskProfile waitUntilGetTaskProfile(String detectorId) throws Interr
         return adTaskProfile;
     }
 
-    protected ADTaskProfile waitUntilTaskDone(String detectorId) throws InterruptedException {
+    protected List<Object> waitUntilTaskDone(String detectorId) throws InterruptedException {
```
ohltyler, line 236: nit: maybe use some Pair/Tuple class or `AbstractMap.SimpleEntry` rather than `List<Object>` here?

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
```diff
@@ -231,14 +233,16 @@ protected ADTaskProfile waitUntilGetTaskProfile(String detectorId) throws Interr
...
         return waitUntilTaskReachState(detectorId, TestHelpers.HISTORICAL_ANALYSIS_DONE_STATS);
     }
 
-    protected ADTaskProfile waitUntilTaskReachState(String detectorId, Set<String> targetStates) throws InterruptedException {
+    protected List<Object> waitUntilTaskReachState(String detectorId, Set<String> targetStates) throws InterruptedException {
```
ohltyler, line 240: same thought here; see previous comment

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
ylwu-amzn: How about we improve later? It's too tight to fix. Will add some todo.

ohltyler: Seems we are ok to merge later / post-1.1? Up to you, so I'm ok either way.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +131,176 @@ public void testBackwardsCompatibility() throws Exception {
...
+
+                    // Create multi category HC detector and start realtime job
+                    List<String> multiCategoryHCResults = createRealtimeAnomalyDetectorsAndStart(MULTI_CATEGORY_HC_DETECTOR);
+                    detectors.put(MULTI_CATEGORY_HC_DETECTOR, singleEntityDetectorResults.get(0));
+                    // Start historical analysis for multi category HC detector
+                    startHistoricalAnalysisOnNewNode(multiCategoryHCResults.get(0), ADTaskType.HISTORICAL_HC_DETECTOR.name());
+
+                    // Verify cluster has 6 detectors now
+                    verifyAnomalyDetectorCount(TestHelpers.AD_BASE_DETECTORS_URI, 6);
+                    // Verify cluster has 3 detectors with search detector API
```
ohltyler, line 205: nit: "3 detectors" -> "6 detectors"

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: > The createRealtimeAnomalyDetectorsAndStart ultimately calls the startAnomalyDetector which uses the LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI to start detector. If the cluster is UPGRADED, should we not use the AD_BASE_DETECTORS_URI instead?

Testing for new URL `AD_BASE_DETECTORS_URI` has been covered in other REST test class like `AnomalyDetectorRestApiIT`. Here BWC is just for testing if old API still compatible on new cluster. We can add more testing to cover new URIs, but don't think that's our first priority in current phase as it has already been covered. 

> And if this works for upgraded cluster, wouldn't this work for mixed cluster as well where you can send request to LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI for creating new detectors?

For AD old version(<=1.0) we only support single flow and single-category HC detectors. For AD new version (>=1.1), we support new detector type: multi-category HC detector. So for multi-catetory HC detector, we should dispatch to new node, otherwise it will fail on old node. As we have no way to dispatch request to specific node, I would push this case to next release.

Ideally we should be able to create single flow and single-category HC detectors on mixed cluster no matter the request served by old or new node. But the mixed cluster has some tricky part for random detector name. I'm working on fixing that. But today is our cut off day for 1.1. So I wish to push what we have to 1.1 first so we can cover some BWC happy case at least automatically. Agree that we should cover more cases, but we should also develop agilely rather than waiting for all cases covered.

VachaShah: That sounds good, thanks for the explanation @ylwu-amzn!

VachaShah (dismissed):
LGTM from bwc perspective!

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +131,176 @@ public void testBackwardsCompatibility() throws Exception {
...
+                assertEquals(RestStatus.OK, TestHelpers.restStatus(stopHistoricalAnalysisResponse));
+            } catch (Exception e) {
+                if (!ExceptionUtil.getErrorMessage(e).contains("No running task found")) {
+                    throw e;
+                }
+            }
+            Map<String, Object> jobAndTask = getDetectorWithJobAndTask(client(), detectorId);
+            AnomalyDetectorJob job = (AnomalyDetectorJob) jobAndTask.get(ANOMALY_DETECTOR_JOB);
+            ADTask historicalAdTask = (ADTask) jobAndTask.get(HISTORICAL_ANALYSIS_TASK);
+            if (!job.isEnabled() && historicalAdTask.isDone()) {
```
ohltyler, line 267: If you're already asserting that the responses from stopping the RT and historical task are true, should there be asserts that the job is disabled and the historical task is done as well?

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +131,176 @@ public void testBackwardsCompatibility() throws Exception {
...
+        deleteRunningDetector(detectorId);
+        waitUntilTaskDone(client(), detectorId);
+        List<ADTask> adTasks = searchLatestAdTaskOfDetector(client(), detectorId, taskType);
+        assertEquals(1, adTasks.size());
+        assertEquals(taskId, adTasks.get(0).getTaskId());
+        int adResultCount = countADResultOfDetector(client(), detectorId, taskId);
+        assertTrue(adResultCount > 0);
+    }
+
+    private void startRealtimeJobForHistoricalDetectorOnNewNode() throws IOException {
```
ohltyler, line 285: these 2 methods include `OnNewNode` when that functionality hasn't been added yet. Maybe rename them, or add comments above them saying that these can't guarantee they are running on new nodes yet until that functionality's available?

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +131,176 @@ public void testBackwardsCompatibility() throws Exception {
...
+            String jobId = startAnomalyDetectorDirectly(client(), detectorId);
+            assertEquals(detectorId, jobId);
+            Map<String, Object> jobAndTask = getDetectorWithJobAndTask(client(), detectorId);
+            AnomalyDetectorJob detectorJob = (AnomalyDetectorJob) jobAndTask.get(ANOMALY_DETECTOR_JOB);
+            assertTrue(detectorJob.isEnabled());
+            runningRealtimeDetectors.add(detectorId);
+        }
+    }
+
+    private void verifyRealtimeJobRunning() throws IOException {
```
ohltyler, line 296: nit: maybe rename to indicate that this is verifying **all current** realtime jobs?

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -122,34 +321,128 @@ private String getUri() {
...
 
-    private void createBasicAnomalyDetector() throws Exception {
-        AnomalyDetector detector = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
-        String indexName = detector.getIndices().get(0);
-        TestHelpers.createIndex(client(), indexName, TestHelpers.toHttpEntity("{\"name\": \"test\"}"));
+    private List<String> createRealtimeAnomalyDetectorsAndStart(ADRestTestUtils.DetectorType detectorType) throws Exception {
+        switch (detectorType) {
+            case SINGLE_ENTITY_DETECTOR:
+                // Create single flow detector
+                Response singleFLowDetectorResponse = createAnomalyDetector(
```
ohltyler, line 328: nit: `singleFLow`- > `singleFlow`

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -122,34 +321,128 @@ private String getUri() {
...
+        try {
+            deleteDetector(client(), detectorId);
+        } catch (Exception e) {
+            assertTrue(ExceptionUtil.getErrorMessage(e).contains("running"));
+        }
+    }
+
+    private List<String> startAnomalyDetector(Response response, boolean historicalDetector) throws IOException {
+        // verify that the detector is created
+        assertEquals("Create anomaly detector failed", RestStatus.CREATED, TestHelpers.restStatus(response));
```
ohltyler, line 410: why asserting equals on a failure message?

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: sure, will fix

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: Historical task stopping is try the best, not 100% sure. Add assertion for historical task state will make the test flaky.

Realtime job stopping should be disabled once the request returns, but actually it's flaky as the index refresh needs some time. Will test if sleep for some time can fix. But from previous experience, sometimes the flaky test case can pass on dev host but will fail in Github workflow. That's why I just assert the stop request returns OK.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: "OnNewNode" means we can only execute these functions on new AD node, if these functions executed on old nodes, they will generate unexpected result.

These functions will run in `CASE UPGRADED` branch, which means fully upgraded cluster, all nodes in this cluster are running new AD version. If BWC test framework supports sending request to a specific node, we can run these functions on only new nodes in mixed cluster as well. But that's impossible to send request to a specific node via REST API now. That will be future work.
```
case "upgraded_cluster":
                    return UPGRADED;
```

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: sure, will fix

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: sure

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ohltyler: Got it, makes sense, since it relies on the index updates. I'm ok with leaving as-is.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ohltyler: Got it, makes sense. Could you add comments explaining that? So it will be easy to integrate these when the mixed-cluster-specific-node-version functionality is supported?

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
ylwu-amzn: I prefer to merge to 1.1, so we can test BWC automatically when infra team trigger CI workflow.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: sure

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: @VachaShah thanks for your help. I added starting realtime job for single-flow detector and single-category HC detector on mixed cluster by checking which round current test is running.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +132,203 @@ public void testBackwardsCompatibility() throws Exception {
...
+                    // Verify cluster has correct number of detectors now
+                    verifyAnomalyDetectorCount(TestHelpers.LEGACY_OPENDISTRO_AD_BASE_DETECTORS_URI, 3);
+                    // Verify cluster has correct number of detectors with search detector API
+                    assertEquals(3, countDetectors(client(), null));
+
+                    // Verify all realtime detectors are running realtime job
+                    verifyAllRealtimeJobsRunning();
                     break;
                 case MIXED:
+                    // We have no way to specify whether send request to old node or new node now.
```
VachaShah, line 167: Is this comment relevant since you added tests based on rounds?

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
VachaShah: No problem! Thanks for adding.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
```diff
@@ -83,25 +132,203 @@ public void testBackwardsCompatibility() throws Exception {
...
+                            break;
+                        case "second":
+                            mixedClusterTestRound = 2;
+                            break;
+                        case "third":
+                            mixedClusterTestRound = 3;
+                            break;
+                        default:
+                            break;
+                    }
```
VachaShah, line 191: You can move this out to a method.

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: Oh, this should be inside the TODO next line, will move to TODO

src/test/java/org/opensearch/ad/bwc/ADBackwardsCompatibilityIT.java
ylwu-amzn: sure, done

VachaShah (approved):

ohltyler (approved):
LGTM!

PR #226, opened 2021-09-23
Adjust transport messaging for single stream detectors

kaituo:

### Description
As part of integrating with ThresholdedRandomCutForest, this PR adjusts transport messaging for single stream detectors.  Previously, we had one remote call to a RandomCutForest model and one remote call to a Kll Threshold model.  After integrating with ThresholdedRandomCutForest, we don’t need the 2nd call to the Kll threshold model any more.  This PR changed to send one remote call to a ThresholdedRandomCutForest model and deal with the backward compatibility in RCFResultResponse.

Testing done:
1. Manual BWC testing.
2. Tested that single-stream detection still works.
 
### Check List


wnbts (approved):

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -510,16 +504,16 @@ private void handleException(Exception e) {
                 return;
             } else {
                 // HC logic ends and single entity logic starts here
-                String thresholdModelID = modelPartitioner.getThresholdModelId(adID);
-                Optional<DiscoveryNode> asThresholdNode = hashRing.getOwningNodeWithSameLocalAdVersionForRealtimeAD(thresholdModelID);
-                if (!asThresholdNode.isPresent()) {
-                    listener.onFailure(new InternalFailure(adID, "Threshold model node is not available."));
+                String rcfModelID = ModelIdMapper.getRcfModelId(adID, 0);
```
ylwu-amzn, line 507: Minor: We are going to use only 1 model partition for single-flow detector. How about adding some comments here, so others can understand what 0 means?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -558,50 +552,70 @@ private void handleException(Exception e) {
...
+            if (!featureOptional.getProcessedFeatures().isPresent()) {
+                Optional<Exception> exception = coldStartIfNoCheckPoint(detector);
+                if (exception.isPresent()) {
+                    listener.onFailure(exception.get());
+                    return;
+                }
+
+                if (!featureOptional.getUnprocessedFeatures().isPresent()) {
+                    // Feature not available is common when we have data holes. Respond empty response
+                    // so that alerting will not print stack trace to avoid bloating our logs.
```
ylwu-amzn, line 566: Not quite get what does this line mean. You mean alerting plugin may print too much log if we don't return empty response?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -773,225 +782,70 @@ private boolean invalidQuery(SearchPhaseExecutionException ex) {
...
+                                response.getConfidence(),
+                                response.getRCFScore(),
+                                featureInResponse,
+                                response.getTotalUpdates(),
+                                detector.getDetectorIntervalInMinutes()
+                            )
+                        );
                 } else {
                     LOG.warn(NULL_RESPONSE + " {} for {}", modelID, rcfNodeID);
+                    listener.onResponse(null);
```
ylwu-amzn, line 829: Can you confirm if the `listener` can handle `null` response?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: added

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -436,22 +431,21 @@ private void handleException(Exception e) {
             long dataStartTime = request.getStart() - delayMillis;
             long dataEndTime = request.getEnd() - delayMillis;
 
-            Optional<Exception> previousException = stateManager.fetchExceptionAndClear(adID);
+            // HC logic starts here
+            if (anomalyDetector.isMultientityDetector()) {
+                Optional<Exception> previousException = stateManager.fetchExceptionAndClear(adID);
```
ylwu-amzn, line 436: I see previous exception checking is now moved inside this if block `if (anomalyDetector.isMultientityDetector()) {`.  Does that mean the previous exception only work for HC detector?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: This is from the first PRs and I never changed.  Changed to:

// Feature not available is common when we have data holes. Respond empty response
// and don't log to avoid bloating our logs.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -436,22 +431,21 @@ private void handleException(Exception e) {
...
-                Exception exception = previousException.get();
-                LOG.error("Previous exception of {}: {}", adID, exception);
-                if (exception instanceof EndRunException) {
-                    EndRunException endRunException = (EndRunException) exception;
-                    if (endRunException.isEndNow()) {
-                        listener.onFailure(exception);
-                        return;
+                if (previousException.isPresent()) {
+                    Exception exception = previousException.get();
+                    LOG.error("Previous exception of {}: {}", adID, exception);
```
ylwu-amzn, line 440: This line will not print exception stack trace. Is this by design?

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: good catch.  Changed to "listener.onFailure(new InternalFailure(adID, NO_MODEL_ERR_MSG));"

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: In the main branch, it is inside HCAD block: https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java#L477-L491

Lai wanted to pursue another design and made the original change.  But due to limited time, I followed a different design after discussing with Lai and moved the code back inside the HCAD block.

HCAD won't wait for response and return right away.  That's we check previous exceptions.  Single-stream would wait for the response and thus doesn't need to check previous exceptions.

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
kaituo: Good catch.  Changed to 

LOG.error(new ParameterizedMessage("Previous exception of [{}]", adID), exception);

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Got it, make sense.

ylwu-amzn (approved):

PR #182, opened 2021-08-17
Limit the max models shown on the stats and profile API

kaituo:
### Description
Previously, the number of models we show on the stats/profile API is unbounded. The unbounded models can cause potential memory/network issues if a vast number of models exist. This PR sets an upper bound and makes the bound a dynamic setting. Due to the above change, the models may not be complete, and thus I added a model_count to the stats/profile API to show the actual models present.

This PR also fixed bugs:
First, AnomalyDetectorProfileRunner used a listener instead of delegateListener, which can cause the profile API to return early before collecting all outputs.
Second, I forgot to support the multi-category fields in AnomalyDetectorProfileRunner.profileEntityStats. This PR adds support there. In detail, profile API needs to report total entities. We can use cardinality aggregation for a single-category field. But we cannot do that for multi-category fields as it requires scripting to generate run time fields, which is expensive. We work around the problem by using a composite query to find the first 10_000 buckets. Generally, traversing all buckets/combinations can't be done without visiting all matches, which is costly for data with many entities. Given that it is often enough to have a lower bound of the number of entities, such as "there are at least 10000 entities", the default is set to 10,000. That is, requests will count the total entities up to 10,000.
Third, I fixed a potential null pointer exception in ADTaskManager.

Testing done:
1. e2e workflow tests by verifying profile/stats API shows correct output, and the limit is in effect.
2. backward compatibility tests to make sure the newly added model_count does not cause issues.
3. added unit tests.

Example:

stats API

```
{
	"anomaly_detectors_index_status": "green",
	"anomaly_detection_state_status": "green",
	"single_entity_detector_count": 0,
	"detector_count": 2,
	"multi_entity_detector_count": 0,
	"anomaly_detection_job_index_status": "green",
	"models_checkpoint_index_status": "green",
	"anomaly_results_index_status": "green",
	"nodes": {
		"-rjw2JaoQ5WNrMdWnbd7nA": {
			"ad_execute_request_count": 0,
			"models": [{
					"detector_id": "4_cyVXsBPFUfO-zDBGpR",
					"model_type": "entity",
					"last_used_time": 1629222319564,
					"model_id": "4_cyVXsBPFUfO-zDBGpR_entity_3S8k6q_DLFhw3hboko3dfw",
					"last_checkpoint_time": 1629222140992,
					"entity": [{
							"name": "host",
							"value": "server_1"
						},
						{
							"name": "service",
							"value": "app_3"
						}
					]
				},
				{
[...]

PR #228, opened 2021-09-24
Introduce ThresholdedRandomCutForest into historical analysis code

kaituo:

### Description
This PR does the following.

1.Replace THRESHOLD_MODEL_TRAINING_SIZE with NUM_MIN_SAMPLES so that we don’t have different variables for the same thing.
2.Integrate with ThresholdedRandomCutForest.
3.Remove memory to store training data for the Kll threshold model as there is only one model ThresholdedRandomCutForest now.
4.Adjust model size calculation to consider new models.

There are rebased changes from main (most of src/main/java/org/opensearch/ad/task/ADTaskManager.java changes). Yaliang should be able to recognize it :)

Testing done:
1. tested single-stream and HCAD historical analyses.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
+++ b/src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
@@ -411,8 +411,6 @@ public final class AnomalyDetectorSettings {
             Setting.Property.Dynamic
         );
 
-    public static int THRESHOLD_MODEL_TRAINING_SIZE = 128;
-
     // Maximum number of old AD tasks we can keep.
     public static int MAX_OLD_AD_TASK_DOCS = 1000;
     public static final Setting<Integer> MAX_OLD_AD_TASK_DOCS_PER_DETECTOR = Setting
@@ -464,7 +462,7 @@ public final class AnomalyDetectorSettings {
     public static final Setting<Integer> MAX_RUNNING_ENTITIES_PER_DETECTOR_FOR_HISTORICAL_ANALYSIS = Setting
         .intSetting(
             "plugins.anomaly_detection.max_running_entities_per_detector_for_historical_analysis",
-            1,
+            2,
             1,
             1000,
             Setting.Property.NodeScope,
@@ -788,4 +786,9 @@ public final class AnomalyDetectorSettings {
     // such as "there are at least 10000 entities", the default is set to 10,000. That is, requests will count the
     // total entities up to 10,000.
     public static final int MAX_TOTAL_ENTITIES_TO_TRACK = 10_000;
+
+    // ======================================
+    // AD Index setting
+    // ======================================
+    public static int MAX_UPDATE_RETRY_TIMES = 10_000;
 }
--- a/src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
+++ b/src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
@@ -29,7 +29,6 @@ package org.opensearch.ad.task;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.NUM_MIN_SAMPLES;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.NUM_SAMPLES_PER_TREE;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.NUM_TREES;
-import static org.opensearch.ad.settings.AnomalyDetectorSettings.THRESHOLD_MODEL_TRAINING_SIZE;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.TIME_DECAY;
 
 import java.util.ArrayDeque;
@@ -40,15 +39,13 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
-import org.opensearch.ad.ml.HybridThresholdingModel;
-import org.opensearch.ad.ml.ThresholdingModel;
 import org.opensearch.ad.model.ADTask;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 
-import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.config.Precision;
+import com.amazon.randomcutforest.parkservices.ThresholdedRandomCutForest;
 
 /**
  * AD batch task cache which will mainly hold these for one task:
@@ -61,12 +58,10 @@ import com.amazon.randomcutforest.config.Precision;
 public class ADBatchTaskCache {
     private final String detectorId;
     private final String taskId;
-    private RandomCutForest rcfModel;
-    private ThresholdingModel thresholdModel;
+    private ThresholdedRandomCutForest rcfModel;
     private boolean thresholdModelTrained;
     private Deque<Map.Entry<Long, Optional<double[]>>> shingle;
     private AtomicInteger thresholdModelTrainingDataSize = new AtomicInteger(0);
-    private double[] thresholdModelTrainingData;
     private AtomicBoolean cancelled = new AtomicBoolean(false);
     private AtomicLong cacheMemorySize = new AtomicLong(0);
     private String cancelReason;
@@ -84,7 +79,7 @@ public class ADBatchTaskCache {
         this.shingle = new ArrayDeque<>(shingleSize);
         int dimensions = detector.getShingleSize() * detector.getEnabledFeatureIds().size();
 
-        rcfModel = RandomCutForest
+        rcfModel = ThresholdedRandomCutForest
             .builder()
             .dimensions(dimensions)
             .numberOfTrees(numberOfTrees)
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -311,13 +311,17 @@ public void startDetector(
         ActionListener<AnomalyDetectorJobResponse> listener
     ) {
         getDetector(detectorId, (detector) -> {
-            if (validateDetector(detector, listener)) { // validate if detector is ready to start
+            if (!detector.isPresent()) {
```
ylwu-amzn, line 314: Not for this line but general question for this class. Seems just merged code from main branch, no specific code change related with this new model integration, right?

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -1087,26 +1085,12 @@ private void detectAnomaly(
...
-                        threshold.train(adTaskCacheManager.getThresholdModelTrainingData(taskId));
-                        adTaskCacheManager.setThresholdModelTrained(taskId, true);
-                    }
-                } else {
-                    grade = threshold.grade(score);
-                    confidence = threshold.confidence();
-                    if (score > 0) {
-                        threshold.update(score);
-                    }
+                AnomalyDescriptor descriptor = trcf.process(point, 0);
```
ylwu-amzn, line 1088: How about adding some comments to explain what 0 means?

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
```diff
@@ -95,23 +90,11 @@ protected ADBatchTaskCache(ADTask adTask) {
...
             .precision(Precision.FLOAT_32)
             .boundingBoxCacheFraction(AnomalyDetectorSettings.BATCH_BOUNDING_BOX_CACHE_RATIO)
-            // same with dimension for opportunistic memory saving
-            // Usually, we use it as shingleSize(dimension). When a new point comes in, we will
-            // look at the point store if there is any overlapping. Say the previously-stored
-            // vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize
-            // overlapping x3, x4, and only store x5, x6.
-            .shingleSize(shingleSize)
+            // for external shingling, rcf does not recognize shingle
+            .shingleSize(1)
```
ylwu-amzn, line 94: Does this mean even user tune shingle size on frontend, we always use shingle size 1 ?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
kaituo: The following lines of ADTaskManager are related to new model integration:

old file: 189~212 vs new file: 195~196

old file: 258~268 vs new file: 226~228

old file: 342~362 vs new file: 303~331

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
kaituo: yes, added

// 0 is placeholder for timestamp. In the future, we will add
// data time stamp there.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: Don't see difference for line `195~196` in new file.
Are these line numbers for class `ADTaskCacheManager.java` ?

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
kaituo: shingle tuning will impact dimension.  It won't impact shingle size here, but impact the overall dimension of the data.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
kaituo: PR

kaituo: Are we looking at the same thing?

ylwu-amzn: Yes, I have reviewed this class `ADTaskCacheManager.java`.  I'm asking this class `ADTaskManager.java`, what's the specific change for this new model integration.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: The timestamp means model's last update time?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
kaituo: oh.  For ADTaskManager, the related changes are:

old file: 2512~2544; new file: 2530~2562

ylwu-amzn: Got it, these changes looks good

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
kaituo: it can be the data start or end time in anomaly result index.  We can use either, just need to be consistent.

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
kaituo: added comments:

            // for external shingling, rcf does not recognize shingle.  Thus, shingle size
            // is 1 here.
            // shingle in detector config and shingle size here are different things.
            // shingle size in detector config impacts dimensions.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
sudiptoguha: +1

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -788,4 +786,9 @@ private AnomalyDetectorSettings() {}
     // such as "there are at least 10000 entities", the default is set to 10,000. That is, requests will count the
     // total entities up to 10,000.
     public static final int MAX_TOTAL_ENTITIES_TO_TRACK = 10_000;
+
```
sudiptoguha, line 789: "MAX_TOTAL_ENTITIES_TO_TRACK = 10_000;" seems like an upper bound. Should the variable be renamed re "MAX_BUCKETS"?? Or are we limiting to 10000 for now (which can be a deliberate decision as well)?

kaituo: It is Yaliang's code :) I'll leave the decision to @ylwu-amzn 

We are limiting to 10000 in historical detectors because historical detectors are more heavier than real time.  I guess it can be changed to a dynamic setting and let users configure.  Will leave it to Yaliang.

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
sudiptoguha: Yes -- this is for existing models. This can be changed subsequently (in a single threaded mode) when a full shingle is available. However for a new model we should use the actual shingle size. In line 79
AnomalyDetector detector = adTask.getDetector(); are we getting config or the model? Seems like config, right?

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
```diff
@@ -95,23 +90,14 @@ protected ADBatchTaskCache(ADTask adTask) {
...
-            // look at the point store if there is any overlapping. Say the previously-stored
-            // vector is x1, x2, x3, x4, now we add x3, x4, x5, x6. RCF will recognize
-            // overlapping x3, x4, and only store x5, x6.
-            .shingleSize(shingleSize)
+            // for external shingling, rcf does not recognize shingle. Thus, shingle size
+            // is 1 here.
+            // shingle in detector config and shingle size here are different things.
+            // shingle size in detector config impacts dimensions.
+            .shingleSize(1)
+            .anomalyRate(1 - AnomalyDetectorSettings.THRESHOLD_MIN_PVALUE)
```
sudiptoguha, line 98: anomaly rate should be close to zero. Where is AnomalyDetectorSettings.THRESHOLD_MIN_PVALUE?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
sudiptoguha: Ok. Makes sense. Let's not make hasty changes.

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
sudiptoguha: Ok, resolved. As lon g as we are using external shingling we should use shingleSize = 1 for the ThresholdedRCF model. If we ever switch to not have external shingle processing (which would be quite a bit of change) then we should revisit this.

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
sudiptoguha: Resolved. AnomalyDetectorSettings.THRESHOLD_MIN_PVALUE = 0.995 from here
https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java#L250

kaituo: Yeah, I am reusing a setting from Kll model.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
ylwu-amzn: @kaituo , seems `MAX_TOTAL_ENTITIES_TO_TRACK` is for entity profile of realtime HCAD, this is written by you, in [PR182](https://github.com/opensearch-project/anomaly-detection/pull/182/files#diff-dfb14a8e1cf93e3abc10cb12cdb3a9b0271728f9b8a8e1dd3c4868697ef7816dR795)

ylwu-amzn (approved):

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: oh, you are right.  I slept too little these days :) @ylwu-amzn

sudiptoguha (approved):

src/main/java/org/opensearch/ad/task/ADBatchTaskCache.java
sudiptoguha: if pulling in PR278 for RCF, Revert back to shingleSIze(shingleSize). Even though shingling is external -- this will keep the thresholded model closer to its final state and remove this distinction.

sudiptoguha (changes requested):
Change the shingleSIze 1 -> shingleSize = shingleSize

kaituo:
> Change the shingleSIze 1 -> shingleSize = shingleSize

Yeah, made change in another PR: https://github.com/opensearch-project/anomaly-detection/pull/238/commits/1731da200dcb5d36a26c53e04a9496eca4c0d32d

PR #231, opened 2021-09-24
Validation API - "Blocker" level validation 

amitgalitz:
### Description

The goal of the validation API is too find issues in detector configuration before the user is stuck and has already tried making the anomaly detector. The Validation API will be split into two parts (Blocker and Non-blocker). 

This PR includes the implementation for these "blocker" checks which are usually in the detector level, these are issues that if not fixed, completely block the creation of an AD. 

1. Add blocker level validation API.
2. Refactor detector config validation so that it is shared between create/update and validate API

 ### API Request

* Path: _opendistro/_anomaly_detection/detectors/**validate**/**<*aspect*>**
* Parameter (aspect): 
    * [default] detector
    * model
* Payload: Detector config

For both examples below the last param of detector can also be switched out with model which would run both the detector "blocker" validation and the model "non-blocker" validations on the detector configs. However currently there are no additional validation checks added for the model type (this will be added in future PRs).

Example where issue found: hc_field is text field

```
➜  ~ curl -X POST "localhost:9200/_opendistro/_anomaly_detection/detectors/_validate/detector" -H 'Content-Type: application/json' -d'
{
    "name":"cli-d-1-min-002",
    "description":"Test nab daily jumpsup",
    "time_field":"timestamp",
    "indices":[
        "test-index"
    ],
    "feature_attributes":[
        {
            "feature_name":"feature-1",
            "feature_enabled":true,
            "aggregation_query":{
                "value_sum":{
                    "sum":{
                        "field":"hc_field"
                    }
                }
            }
        },
        {
            "feature_name":"feature-2",
            "feature_enabled":true,
            "aggregation_query":{
                "value_sum_1":{
                    "sum":{
                        "field":"hc_field"
                    }
                }
            }
        }
    ],
    "detection_interval":{
        "period":{
            "interval":1,
            "unit":"Minutes"
        }
    }
}
'
```

Response:

```
{
  "detector": {
    "feature_attributes": {
      "message": "Feature has invalid query causing runtime exception: feature-1, Feature has invalid query causing runtime exception: feature-2",
      "sub_issues": {
        "feature-2": "Feature has invalid query causing runtime exception",
        "feature-1": "Feature has invalid query causing runtime exception"
      }
    }
  }
}

```
Example: validation passed:

```
➜  ~ curl -X POST "localhost:9200/_opendistro/_anomaly_detection/detectors/_validate/detector" -H 'Content-Type: application/json' -d'
{
    "name":"cli-d-1-min-002",
    "description":"Test nab daily jumpsup",
    "time_field":"timestamp",
    "indices":[
        "test-index"
    ],
    "feature_attributes":[
        {
            "feature_name":"feature-1",
            "feature_enabled":true,
            "aggregation_query":{
                "value_sum":{
                    "sum":{
                        "field":"feature-1"
                    }
                }
            }
        }
    ],
    "detection_interval":{
        "period":{
            "interval":1,
            "unit":"Minutes"
        }
    }
}
'
```

Response:

`{}`

### Further Design Details

REST-level:

* Abstracted IndexAnomalyDetectionAction logic into the new AbstractAnomalyDetectionActionHandler, moved all logic there. Main reason: keep the same functionality, but add a dry-run param to propagate down. This way, when dry-running with validation, we don’t actually make the index changes, but can still use the exact same logic
* Changed IndexAnomalyDetectorActionHandler to extend AbstractAnomalyDetectorActionHandler, passing dry-run to false, (since we want to actually make the index changes). Deleted rest of the logic from this class since it is in the extended AbstractAnomalyDetectorActionHandler
* Implemented ValidateAnomalyDetectorAction which extends AbstractAnomalyDetectorActionHandler, passing dry-run to true
* In AbstractAnomalyDetectionActionHandler logic for checking features has moved from the onGetAnomalyDetectorResponse method that was previously in IndexAnomalyDetectorActionHandler to new validateAnomalyDetectorFeatures method
    * New validateAnomalyDetectorFeatures now both validates for config/syntax of features and for runtime errors of detector features
* Most of the exceptions returned have changed with the new code which also effects create and update API. I created a new ADValidationException class that is now returned for most issues that used to be of type IllegalArgument

Transport-level:

- Added ValidateAnomalyDetectorTransportAction  which handles some of the security logic and parsing of ADValidationExceptions

 
### Check List


src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
```diff
@@ -0,0 +1,70 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
ohltyler, line 25: For all newly created files, this header isn't necessary. Only need the SPDX one

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
```diff
@@ -0,0 +1,162 @@
...
+public class DetectorValidationIssue implements ToXContentObject, Writeable {
+    private static final String MESSAGE_FIELD = "message";
+    private static final String SUGGESTED_FIELD_NAME = "suggested_value";
+    private static final String SUB_ISSUES_FIELD_NAME = "sub_issues";
+
+    private final ValidationAspect aspect;
+    private final DetectorValidationIssueType type;
+    private final String message;
+    private Map<String, String> subIssues;
+    private Object suggestion;
```
ohltyler, line 59: Why not String or a separate `Suggestion` class here if needed?

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
```diff
@@ -0,0 +1,82 @@
...
+            case AnomalyDetector.FILTER_QUERY_FIELD:
+                return FILTER_QUERY;
+            default:
+                throw new IllegalArgumentException("Unsupported validation type");
+        }
+    }
+
+    public static Set<DetectorValidationIssueType> getNames(Collection<String> names) {
+        return Name.getNameFromCollection(names, DetectorValidationIssueType::getName);
+    }
```
ohltyler, line 81: Don't see this used anywhere, can it be removed?

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public List<ReplacedRoute> replacedRoutes() {
+        return ImmutableList
+            .of(
+                // validate detector
+                new ReplacedRoute(
+                    RestRequest.Method.POST,
+                    String.format(Locale.ROOT, "%s/%s", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE),
+                    RestRequest.Method.POST,
+                    String.format(Locale.ROOT, "%s/%s", AnomalyDetectorPlugin.LEGACY_OPENDISTRO_AD_BASE_URI, VALIDATE)
```
ohltyler, line 79: @ylwu-amzn because this is a new API should it be necessary to support the legacy path?

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
```diff
@@ -0,0 +1,113 @@
...
+                ),
+                // validate detector with type
+                new ReplacedRoute(
+                    RestRequest.Method.POST,
+                    String.format(Locale.ROOT, "%s/%s/{%s}", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE, TYPE),
```
ohltyler, line 84: Can you give an example of an api call hitting this path? And can you add the list of possible `TYPE`s to the PR description? That may be helpful here

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -0,0 +1,821 @@
...
+    protected final AnomalyDetectorActionHandler handler = new AnomalyDetectorActionHandler();
+    protected final RestRequest.Method method;
+    protected final Client client;
+    protected final TransportService transportService;
+    protected final NamedXContentRegistry xContentRegistry;
+    protected final ActionListener<T> listener;
+    protected final User user;
+    protected final ADTaskManager adTaskManager;
+    protected final SearchFeatureDao searchFeatureDao;
+    protected final boolean isDryRun;
```
ohltyler, line 133: Because this instance variable is set upon creation, can we just utilize the value of `this.isDryRun` where it is actually needed, rather than passing it down through several layers of method calls?

Seems the majority of the methods are simply passing the var down to other methods, and it's only really needed in a few places:
- Actually indexing vs. not indexing on [lines 557-558](https://github.com/opensearch-project/anomaly-detection/blob/1659807b945810edcdaae644912b6d7b61dc0b6e/src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java#L557-L558))
- Parsing the responses after creating ([line 359](https://github.com/opensearch-project/anomaly-detection/blob/1659807b945810edcdaae644912b6d7b61dc0b6e/src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java#L359), [line 369](https://github.com/opensearch-project/anomaly-detection/blob/1659807b945810edcdaae644912b6d7b61dc0b6e/src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java#L369))

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
amitgalitz: POST _opendistro/_anomaly_detection/detectors/_validate/model 
or 
POST _opendistro/_anomaly_detection/detectors/_validate/detector will hit this API. 

The payload would be the same but if the type param is model then it will also execute the model "non-blocker" validation on top of the "blocker" detector validation. I didn't add a specific example of that to this PR because this PR doesn't have the implementation for additional model checks but will add something regarding this. Currently if the model type is used then it just runs the detector validation and no additional validation checks for the model at the moment. 

An example response that isn't implemented yet for the model type: 
```
{
  "model": {
    "detection_interval": {
      "suggested_value": {
        "period": {
          "interval": 120,
          "unit": "Minutes"
        }
      }
      "message": "use suggested value"
    }
  }
}
```

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
amitgalitz: Currently it is a place holder. It will most likely be changed to string, but it may change to other type if we need more complicated type for suggestion of model level validation. Will finalize soon as I finalize details for model validation design.

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -0,0 +1,821 @@
...
+                    errorMessage = UNKNOWN_SEARCH_QUERY_EXCEPTION_MSG + feature.getName();
+                }
+
+                logger.error(errorMessage, e);
+                multiFeatureQueriesResponseListener.onFailure(new OpenSearchStatusException(errorMessage, RestStatus.BAD_REQUEST, e));
+            }));
+        }
+    }
+
+    // public class Indexer<T> {
```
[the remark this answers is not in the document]
amitgalitz: fixed and removed, was a different implementation I was trying out and forgot to delete

src/main/java/org/opensearch/ad/rest/handler/ValidateAnomalyDetectorActionHandler.java
```diff
@@ -0,0 +1,138 @@
...
+            maxMultiEntityAnomalyDetectors,
+            maxAnomalyFeatures,
+            method,
+            xContentRegistry,
+            user,
+            null,
+            searchFeatureDao,
+            true
+        );
+        String normalizedTypes = StringUtils.isBlank(typeStr) ? ValidationAspect.MODEL.getName() : typeStr.trim().replaceAll("\\s", "");
```
ohltyler, line 118: Shouldn't this default to `detector` instead of `model`?

src/main/java/org/opensearch/ad/rest/handler/ValidateAnomalyDetectorActionHandler.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public void start() throws IOException {
+        super.start();
+        if (aspects.contains(ValidationAspect.MODEL)) {
+            validateModelConfig();
+        }
+    }
+
+    // TODO: add implementation for model config validation
```
ohltyler, line 133: Adding some more comments may be helpful here, like how all model config validation is non-blocker, maybe a link to this PR or other open issues about future plans to improve the validation

src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
amitgalitz: fixed

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
```diff
@@ -0,0 +1,162 @@
...
+ * subIssues are issues for each feature;
+ * suggestion is how to fix the issue/subIssues found
+ */
+public class DetectorValidationIssue implements ToXContentObject, Writeable {
+    private static final String MESSAGE_FIELD = "message";
+    private static final String SUGGESTED_FIELD_NAME = "suggested_value";
+    private static final String SUB_ISSUES_FIELD_NAME = "sub_issues";
+
+    private final ValidationAspect aspect;
+    private final DetectorValidationIssueType type;
```
ohltyler, line 41: Minor thought - is it necessary to have a separate class file for the type? Can it be implemented directly as an enum in this class for simplicity?

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
```diff
@@ -0,0 +1,273 @@
...
+        AnomalyDetector detector = request.getDetector();
+        ActionListener<ValidateAnomalyDetectorResponse> validateListener = ActionListener.wrap(response -> {
+            logger.debug("Result of validation process " + response);
+            // forcing response to be empty
+            listener.onResponse(new ValidateAnomalyDetectorResponse((DetectorValidationIssue) null));
+        }, exception -> {
+            if (exception instanceof ADValidationException) {
+                // ADValidationException is converted as validation issues returned as response to user
+                DetectorValidationIssue issue = parseADValidationException((ADValidationException) exception);
+                listener.onResponse(new ValidateAnomalyDetectorResponse(issue));
```
ohltyler, line 130: why call `onResponse` here instead of `onFailure`?

ohltyler: ^ Nvm, I think I understand - after validating, if any errors found, will propagate the issues and fixes via exceptions, which are then parsed into DetectorValidationIssues and a response is built from that. If the exception is caused by something else (illegal argument, etc.), it is considered a failure

To me, seems kind of weird to "wrap" the response via an exception then "unwrap"/parse it here. Is the limitation because it is implementing the abstract handler which originally was throwing exceptions because of a failed attempt to actually update/create the detector?

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
```diff
@@ -0,0 +1,162 @@
...
+ * the issue would be in `detector` aspect, not `model`;
+ * and its type is FEATURE_ATTRIBUTES, because it is related to feature;
+ * message would be the message from thrown exception;
+ * subIssues are issues for each feature;
+ * suggestion is how to fix the issue/subIssues found
+ */
+public class DetectorValidationIssue implements ToXContentObject, Writeable {
+    private static final String MESSAGE_FIELD = "message";
+    private static final String SUGGESTED_FIELD_NAME = "suggested_value";
+    private static final String SUB_ISSUES_FIELD_NAME = "sub_issues";
```
ohltyler, line 38: how about changing to something like `"feature_issues"`? I got confused reading later code wondering what the sub-issues were related to

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
```diff
@@ -0,0 +1,273 @@
...
+                errorMessage = originalErrorMessage;
+                break;
+            default:
+                logger.warn(String.format(Locale.ROOT, "Invalid AD validation exception type: %s", exception));
+                return null;
+        }
+        return new DetectorValidationIssue(exception.getAspect(), exception.getType(), errorMessage, subIssues, suggestion);
+    }
+
+    private Map<String, String> getFeatureSubIssuesFromErrorMessage(String errorMessage) {
```
ohltyler, line 200: Can you add some comments with an example of a string representation getting parsed into a map here?

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
```diff
@@ -0,0 +1,273 @@
...
+        Map<String, String> result = new HashMap<String, String>();
+        String featureSubIssueMessageSeparator = ", " + FEATURE_INVALID_MSG_PREFIX;
+
+        String[] subIssueMessagesSuffix = errorMessage.split(featureSubIssueMessageSeparator);
+
+        if (subIssueMessagesSuffix.length == 1) {
+            result.put(getFeatureNameFromErrorMessage(errorMessage), getFeatureIssueFromErrorMessage(errorMessage));
+            return result;
+        }
```
ohltyler, line 209: Can this be removed? Seems its handled in the for loop below already

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
ohltyler: + also not sure on where `FEATURE_INVALID_MSG_PREFIX`  would exist in the string, so an example of that would be helpful too

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
```diff
@@ -0,0 +1,273 @@
...
+        for (int i = 0; i < subIssueMessagesSuffix.length; i++) {
+            if (i == 0) {
+                // element at 0 doesn't have FEATURE_INVALID_MSG_PREFIX removed
+                result
+                    .put(
+                        getFeatureNameFromErrorMessage(subIssueMessagesSuffix[i]),
+                        getFeatureIssueFromErrorMessage(subIssueMessagesSuffix[i])
```
ohltyler, line 216: rather than 2 helper methods, how about using `String.split()` to get the name and corresponding issue?

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
ohltyler: Sure. If the suggestions are all `String`s for now I would recommend changing to `String` here, and could be abstracted out to something else later if it becomes more complex (when model/non-blocker validation is done), or could add a comment here to explain why it's an Object for now

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
ohltyler: Got it, have 2 questions:
1. Why does `model` run both detector/blocker validation and model/non-blocker validation? Shouldn't it just run the latter? If users wanted to run both aspects, they could just pass both types (e.g., `<path>/_validate/detector,model`)?
2. Because there is nothing for `model` yet, should that endpoint stay hidden until there's actually additional validation being done? Considering this may go into a separate release before non-blocker validation is done, we don't want to add user confusion there

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
amitgalitz: Changed this.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -164,31 +167,49 @@ public AnomalyDetector(
         User user
     ) {
         if (Strings.isBlank(name)) {
-            throw new IllegalArgumentException("Detector name should be set");
+            throw new ADValidationException("Detector name should be set", DetectorValidationIssueType.NAME, ValidationAspect.DETECTOR);
```
ylwu-amzn, line 170: AD frontend limit detector name "Detector name must contain 1-64 characters. Valid characters are a-z, A-Z, 0-9, -(hyphen) and _(underscore).". 
Feature name has the same limitation on frontend.

How about we add the same limitation at backend ?

src/main/java/org/opensearch/ad/model/ValidationAspect.java
```diff
@@ -0,0 +1,54 @@
...
+package org.opensearch.ad.model;
+
+import java.util.Collection;
+import java.util.Set;
+
+import org.opensearch.ad.Name;
+import org.opensearch.ad.constant.CommonName;
+
+public enum ValidationAspect implements Name {
+    DETECTOR(CommonName.DETECTOR),
```
ylwu-amzn, line 21: How about adding some java doc for these validation aspects?

src/main/java/org/opensearch/ad/model/MergeableList.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.model;
+
+import java.util.List;
+
+public class MergeableList<T> implements Mergeable {
```
ylwu-amzn, line 16: Don't see this class used except in `MergeableListTests.java` , can we remove this class?

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
```diff
@@ -0,0 +1,143 @@
...
+import org.opensearch.rest.BaseRestHandler;
+import org.opensearch.rest.BytesRestResponse;
+import org.opensearch.rest.RestChannel;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.RestStatus;
+import org.opensearch.rest.action.RestToXContentListener;
+
+import com.google.common.collect.ImmutableList;
+
+public class RestValidateAnomalyDetectorAction extends AbstractAnomalyDetectorAction {
```
ylwu-amzn, line 48: Add some java doc?

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
ylwu-amzn: I think we can just support the new path.

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorRequest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.unit.TimeValue;
+
+public class ValidateAnomalyDetectorRequest extends ActionRequest {
+
+    private final AnomalyDetector detector;
+    private final String typeStr;
```
ylwu-amzn, line 26: `typeStr` means the validation type, `detector` or `model`?

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
```diff
@@ -0,0 +1,260 @@
...
+        Consumer<AnomalyDetector> function
+    ) {
+        try {
+            // Check if user has backend roles
+            // When filter by is enabled, block users validating detectors who do not have backend roles.
+            if (filterByEnabled && !checkFilterByBackendRoles(requestedUser, listener)) {
+                return;
+            }
+            // Validate Detector
+            function.accept(null);
```
ylwu-amzn, line 112: Don't need to pass any parameter to consumer function here. How about use `AnomalyDetectorFunction`?

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -0,0 +1,719 @@
...
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.RestStatus;
+import org.opensearch.search.aggregations.AggregatorFactories;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.transport.TransportService;
+
+public abstract class AbstractAnomalyDetectorActionHandler<T extends ActionResponse> {
```
ylwu-amzn, line 83: How about adding java doc to explain in details about differences between create/update and validate? And what does dry run mean in this class?

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
amitgalitz: 1. Model runs both the detector/blocker and model/non-blocker validation because in order to make sure you can even run the model validation all the detector validation checks have to pass such as there being an existent data index, non negative interval and so on. 

2. I agree that it could be confusing at the moment to have the model param. At the moment I can just keep the detector param only and completely remove the model route.

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
amitgalitz: That makes sense, I changed it to String now and will change later as you said if it needs to be for non-blocker validation.

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorRequest.java
amitgalitz: Yes, will change to validationType

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
amitgalitz: Done

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
amitgalitz: removed it, simplified method entirely

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
amitgalitz: removed

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
amitgalitz: I think I want to keep this as a separate class file as it can potentially be changed and it keeps consistent with a lot of other enum classe files that were made.

src/main/java/org/opensearch/ad/model/ValidationAspect.java
amitgalitz: added

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
```diff
@@ -59,31 +62,19 @@ public String getName() {
...
-                new Route(RestRequest.Method.POST,
-                        String.format(Locale.ROOT, "%s/%s", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE)),
-                new Route(RestRequest.Method.POST,
-                        String.format(Locale.ROOT, "%s/%s/{%s}", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE, TYPE))
-        );
+        return ImmutableList
+            .of(
+                new Route(
+                    RestRequest.Method.POST,
+                    String.format(Locale.ROOT, "%s/%s", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE)
```
ylwu-amzn, line 69: If user use this URL without inputing the TYPE, will we just detect blocking issues?

src/main/java/org/opensearch/ad/model/MergeableList.java
ylwu-amzn: The link doesn't show me the exact line, you can just write which line is using this class, I searched `MergeableList` on this PR page, seems just used in MergeableListTests.java,  maybe missed some folded code.

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
amitgalitz: Yes it defaults to blocking issues only

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
```diff
@@ -0,0 +1,144 @@
...
+    public List<Route> routes() {
+        return ImmutableList
+            .of(
+                new Route(
+                    RestRequest.Method.POST,
+                    String.format(Locale.ROOT, "%s/%s", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE)
+                ),
+                new Route(
+                    RestRequest.Method.POST,
+                    String.format(Locale.ROOT, "%s/%s/{%s}", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, VALIDATE, TYPE)
```
kaituo, line 73: Cannot comment on https://github.com/opensearch-project/anomaly-detection/pull/231#discussion_r721536978

Will reply here.

> Do you think it's fair to completely remove the params for just the blocker release due to the fact that /_validate/detector will have the same output as /_validate, or to keep it to give user some information about the type of validation they are doing and make the addition of model validation a smoother experience later on?

It is fine to keep this end point.  I guess Tyler is suggesting not to expose the model type and only deal with detector type before you release non-blocker validation.

> Why does model run both detector/blocker validation and model/non-blocker validation? Shouldn't it just run the latter? If users wanted to run both aspects, they could just pass both types (e.g., <path>/_validate/detector,model)

I like Tyler's suggestion.  detector is for blocker part and model is for non-blocker part.  Profile/stats API use similar format.

amitgalitz: This makes sense to me, changed the implementation slightly so its similar to Profile API. Detector param for blocker part and Model Param will be for non-blocker part. However, to hide the model type at the moment for 1.2 release I removed it from the ValidationAspect class. So currently validation type can be left empty and it defaults to detector or if its anything else except configs I return an exception stating that the given validation type doesn't exist.

src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
```diff
@@ -0,0 +1,56 @@
...
+ * GitHub history for details.
+ */
+
+package org.opensearch.ad.common.exception;
+
+import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.ad.model.DetectorValidationIssueType;
+import org.opensearch.ad.model.ValidationAspect;
+
+public class ADValidationException extends AnomalyDetectionException {
```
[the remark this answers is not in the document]
amitgalitz: done

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -409,7 +444,17 @@ public static AnomalyDetector parse(
                     }
                     break;
                 case DETECTION_INTERVAL_FIELD:
-                    detectionInterval = TimeConfiguration.parse(parser);
+                    try {
+                        detectionInterval = TimeConfiguration.parse(parser);
+                    } catch (Exception e) {
+                        if (e instanceof IllegalArgumentException && e.getMessage().contains("should be non-negative")) {
```
ylwu-amzn, line 450: Why not just throw `ADValidationException` in `IntervalTimeConfiguration` ?

src/main/java/org/opensearch/ad/model/MergeableList.java
ylwu-amzn: Got it, looks just used it as general List, won't use the `merge` method. I guess the future change will use `MergeableList#merge` method. I'm ok with that.

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorResponse.java
```diff
@@ -0,0 +1,61 @@
...
+        return issue;
+    }
+
+    public ValidateAnomalyDetectorResponse(DetectorValidationIssue issue) {
+        this.issue = issue;
+    }
+
+    public ValidateAnomalyDetectorResponse(StreamInput in) throws IOException {
+        super(in);
+        if (in.available() > 0) {
```
ylwu-amzn, line 37: No error by adding this check. But I don't think we need to add this check here as we release validation feature the first time, this check is mainly for BWC.

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorResponse.java
```diff
@@ -0,0 +1,61 @@
...
+            issue = new DetectorValidationIssue(in);
+        }
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        if (issue != null) {
```
ylwu-amzn, line 44: How about we write false to output stream if issue is null? So we can check if `in.readBoolean()` is true in line 37.
If we need to add more fields to this class in future, for example we add filed "suggestion",  if issue is null , suggestion not null, will be hard to know if the value read from input stream is for issue or for suggestion.

src/main/java/org/opensearch/ad/model/DetectorValidationIssue.java
ohltyler: Sure, I'm ok to keep it separate.

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
amitgalitz: Sounds good, keeping detector param for now and model will just be added in the future

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorResponse.java
amitgalitz: changed this

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorResponse.java
amitgalitz: added

ohltyler (dismissed):
Latest changes LGTM. Make sure to update frontend changes to handle AD blocking the creation if validation fails.

src/main/java/org/opensearch/ad/util/RestHandlerUtils.java
```diff
@@ -143,6 +155,21 @@ private static String validateFeatures(List<Feature> features) {
         return errorMsgBuilder.toString();
     }
 
+    public static boolean isExceptionCausedByInvalidQuery(Exception ex) {
+        if (!(ex instanceof SearchPhaseExecutionException)) {
+            return false;
+        }
+        SearchPhaseExecutionException exception = (SearchPhaseExecutionException) ex;
+        // If all shards return bad request and failure cause is IllegalArgumentException, we
```
ylwu-amzn, line 163: Minor, "If all shards..." should be "If any shard.." ?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -70,16 +73,20 @@
     public static final String TYPE = "_doc";
     public static final String QUERY_PARAM_PERIOD_START = "period_start";
     public static final String QUERY_PARAM_PERIOD_END = "period_end";
+    public static final String PARSING_ISSUE = "query_parsing";
+    public static final String GENERAL_SETTINGS = "general_settings";
+    public static final String NAME_REGEX = "[a-zA-Z0-9._-]+";
```
ylwu-amzn, line 78: Frontend limit on detector name "Detector name must contain 1-64 characters. Valid characters are a-z, A-Z, 0-9, -(hyphen) and _(underscore).", we don't allow dot `.` in detector name.

Can you verify if this regex `[a-zA-Z0-9._-]+` will allow dot `.` in detector name?

amitgalitz: Good catch! This regex currently allows dot `.` in detector name. On the frontend we also currently allow dot `.` in detector name(https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/913be1b6457c92e1d2bb74e605afb31882652f7b/public/utils/constants.ts#L84). Based on message we give user this a bug we have on frontend also then. I missed this as I was referring to frontend regex to make sure my backend regex was correct. I will change this to not include dot `.` and also correspondingly change this on frontend on validation PR there.

src/main/java/org/opensearch/ad/util/RestHandlerUtils.java
amitgalitz: changed

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
ylwu-amzn: I see you have moved the detector name regex check to validate/create/update steps, that looks good to me. So if user has old detector created with invalid name via AD API, we can show these detectors on AD dashboard correctly but will throw exception when user update it.

amitgalitz: Yes exactly.

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
ohltyler: @amitgalitz minor ping on this ^ any thoughts? No blocker for me but curious on your thoughts. If agreed this is a good idea, can you open an issue to track for next release?

ohltyler (approved):
Done another round of review, LGTM. As confirmed offline, security works as expected now based on the changes made in the security PRs (which will need to get approved/merged).

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
amitgalitz: Hi I looked into this and since then I have added a few more check for indexing variable but utilizing `this.isDryRun` could reduce some unnecessary passing down of the var. I meant to come back to this but totally forgot while fixing other bugs. I will review again to make sure calling`this.isDryRun` is cleaner with new added checks and create an issue.

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #253, opened 2021-09-30
fix null AD version of non-data-type node

ylwu-amzn:

### Description
We only record AD version of data nodes in HashRing. But the test cluster created by infra team has remote_cluster_clients node which server external REST request. For task profile, the request will go to the remote_cluster_clients node first, then send broadcast to gather task profile on all data nodes. Data nodes will send back task profile based on the remote AD version ( remote_cluster_clients node), as we don’t record  remote_cluster_clients node’s AD version, data node will think it’s incompatible node then send null task profile back check https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/transport/ADTaskProfileNodeResponse.java#L69, https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/cluster/HashRing.java#L537

This issue only impacts the task profile API.
 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/cluster/ADClusterEventListener.java
+++ b/src/main/java/org/opensearch/ad/cluster/ADClusterEventListener.java
@@ -31,10 +31,8 @@ import java.util.concurrent.Semaphore;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionListener;
-import org.opensearch.ad.util.DiscoveryNodeFilterer;
 import org.opensearch.cluster.ClusterChangedEvent;
 import org.opensearch.cluster.ClusterStateListener;
-import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.node.DiscoveryNodes.Delta;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/ad/cluster/ADClusterEventListener.java
```diff
@@ -96,27 +84,8 @@ public void clusterChanged(ClusterChangedEvent event) {
...
-            for (DiscoveryNode addedNode : delta.addedNodes()) {
-                if (nodeFilter.isEligibleNode(addedNode)) {
-                    LOG.info(NODE_ADDED_MSG + " {}", addedNode.getId());
-                    dataNodeAdded = true;
-                    break;
-                }
-            }
-
-            if (dataNodeAdded || dataNodeRemoved) {
+            if (delta.removed() || delta.added()) {
```
ohltyler, line 87: is this log necessary? Wouldn't this already be logged in core during a cluster changed event like a node leaving or joining?

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -243,17 +240,21 @@ private void buildCircles(Set<String> removedNodeIds, Set<String> addedNodeIds,
         try {
             DiscoveryNode localNode = clusterService.localNode();
             if (removedNodeIds != null && removedNodeIds.size() > 0) {
-                LOG.info("Remove nodes from AD version hash ring: {}", Arrays.toString(removedNodeIds.toArray(new String[0])));
+                LOG.info("Node removed: {}", Arrays.toString(removedNodeIds.toArray(new String[0])));
```
ohltyler, line 243: + here too, can this be removed? or changed to `debug` level?

ylwu-amzn: The log just print when new node joins, generally it only prints several times. We'd better keep these log to help RCA. For example, this PR helped me when I debug the null AD version issue fixed in this PR.

src/main/java/org/opensearch/ad/cluster/ADClusterEventListener.java
ylwu-amzn: I'd prefer to keep this log to make AD debug easier, in case any bug happen we can tell apart from core issue or our own issue easily. And this log won't print too much, just print when cluster starts generally.

src/main/java/org/opensearch/ad/cluster/HashRing.java
```diff
@@ -243,17 +240,21 @@ private void buildCircles(Set<String> removedNodeIds, Set<String> addedNodeIds,
...
                 for (String nodeId : removedNodeIds) {
-                    removeNodeFromCircles(nodeId);
+                    ADNodeInfo nodeInfo = nodeAdVersions.remove(nodeId);
+                    if (nodeInfo != null && nodeInfo.isEligibleDataNode()) {
+                        removeNodeFromCircles(nodeId, nodeInfo.getAdVersion());
+                        LOG.info("Remove data node from AD version hash ring: {}", nodeId);
+                    }
```
ohltyler, line 249: Can you confirm that only data nodes are added to the hash ring? And we store all versions regardless of data node or not?

ylwu-amzn: Yes, that's correct. Hash ring only stores eligible data nodes so when we dispatch realtime RCF model or historical task, we only choose one from those data nodes. But we store all nodes' AD version to make sure if the request from non-data-type node, we can recognize the its AD version and send back correct response.

ohltyler (approved):

PR #268, opened 2021-10-15
Add extra fields to anomaly result index

kaituo:
### Description
ThresholdedRandomCutForest has exposed useful fields such as attribution and expected values when an anomaly is found.  This PR adds those fields to the result index.  if there is an anomaly, we write everything to result; otherwise, we only record score, grade, confidence, and threshold to save space.

Also, this PR did refactoring.
First, we have used a telescoping constructor pattern to write a lot of constructors in AnomalyResult for different combinations of parameters.  It is hard to write client code when there are many parameters (this PR adds 8 new parameters), and harder still to read it.  Instead, I reduce the number of constructors to 2: one for the exception/empty result case and another one for the normal case.
Second, I removed RcfResult and solely relied on ThresholdingResult since we don't differentiate rcf and thresholding models any more (related changes were made in the previous PRs).

Testing done:
1. Manually run preview API, historical and real time detectors.
2. Bwc tests.
 
Example result:

```
Normal:
{
	"_index": "opensearch-ad-plugin-result-a",
	"_type": "_doc",
	"_id": "1Dci43wBgEQAbjDnCzFG",
	"_score": null,
	"_source": {
		"detector_id": "kzcZ43wBgEQAbjDnhzGF",
		"schema_version": 5,
		"data_start_time": 1635898161367,
		"data_end_time": 1635898221367,
		"feature_data": [{
				"feature_id": "deny_max1",
				"feature_name": "deny max1",
				"data": 2322.0
			},
			{
				"feature_id": "deny_avg",
				"feature_name": "deny avg",
				"data": 1718.6666666666667
			},
			{
				"feature_id": "deny_min",
				"feature_name": "deny min",
				"data": 1375.0
			},
			{
				"feature_id": "deny_sum",
				"feature_name": "deny sum",
				"data": 5156.0
			},
			{
				"feature_id": "accept_max5",
				"feature_name": "accept max5",
				"data": 31198.0
			}
		],
		"execution_start_time": 1635898231577,
		"execution_end_time": 1635898231622,
		"anomaly_score": 1.8124904404395776,
		"anomaly_grade": 0.0,
		"confidence": 0.9802940756605277,
		"entity": [{
			"name": "service",
			"value": "app_3"
		}],
		"model_id": "kzcZ43wBgEQAbjDnhzGF_entity_app_3",
		"threshold": 1.2368549346675202
	},
	"sort": [
		1635898231577
	]
}

Anomaly:

{
	"detector_id": "fylE53wBc9MCt6q12tKp",
	"schema_version": 0,
	"data_start_time": 1635927900000,
	"data_end_time": 1635927960000,
	"feature_data": [{
			"feature_id": "deny_max1",
			"feature_name": "deny max1",
			"data": 2291.0
		},
		{
			"feature_id": "deny_avg",
			"feature_name": "deny avg",
			"data": 1677.3333333333333
		},
		{
			"feature_id": "deny_min",
			"feature_name": "deny min",
			"data": 1054.0
		},
		{
			"feature_id": "deny_sum",
			"feature_name": "deny sum",
			"data": 5032.0
		},
		{
			"feature_id": "accept_max5",
			"feature_name": "accept max5",
			"data": 11422.0
		}
	],
	"anomaly_score": 1.1986675882872033,
	"anomaly_grade": 0.26806225550178464,
	"confidence": 0.9607519742565531,
	"entity": [{
		"name": "service",
		"value": "app_4"
	}],
	"approx_anomaly_start_time": 1635927900000,
	"relevant_attribution": [{
			"feature_id": "deny_max1",
			"data": 0.03628638020431366
		},
		{
			"feature_id": "deny_avg",
			"data": 0.03384479053991436
		},
		{
			"feature_id": "deny_min",
			"data": 0.058812549572819096
		},
		{
			"feature_id": "deny_sum",
			"data": 0.10154576265526988
		},
		{
			"feature_id": "accept_max5",
			"data": 0.7695105170276828
		}
	],
	"expected_values": [{
		"likelihood": 1.0,
		"value_list": [{
				"feature_id": "deny_max1",
				"data": 2291.0
			},
			{
				"feature_id": "deny_avg",
				"data": 1677.3333333333333
			},
			{
				"feature_id": "deny_min",
				"data": 1054.0
			},
			{
				"feature_id": "deny_sum",
				"data": 6062.0
			},
			{
				"feature_id": "accept_max5",
				"data": 23379.0
			}
		]
	}],
	"threshold": 1.0993584705913992,
	"execution_end_time": 1635898427895,
	"execution_start_time": 1635898427803
}

If anomaly is detected late:

{
	"_index": "opensearch-ad-plugin-result-a",
	"_type": "_doc",
	"_id": "dDcl43wBgEQAbjDnClhb",
	"_version": 1,
	"_seq_no": 9938,
	"_primary_term": 1,
	"_score": null,
	"_source": {
		"detector_id": "kzcZ43wBgEQAbjDnhzGF",
		"confidence": 0.9746820962328963,
		"relevant_attribution": [{
				"feature_id": "deny_max1",
				"data": 0.07339452532666227
			},
			{
				"feature_id": "deny_avg",
				"data": 0.04934972719948845
			},
			{
				"feature_id": "deny_min",
				"data": 0.01803003656061806
			},
			{
				"feature_id": "deny_sum",
				"data": 0.14804918212089874
			},
			{
				"feature_id": "accept_max5",
				"data": 0.7111765287923325
			}
		],
		"task_id": "9Dck43wBgEQAbjDn4zEe",
		"threshold": 1.0,
		"model_id": "kzcZ43wBgEQAbjDnhzGF_entity_app_0",
		"schema_version": 5,
		"anomaly_score": 1.141419389056506,
		"execution_start_time": 1635898427803,
		"past_values": [{
				"feature_id": "deny_max1",
				"data": 905.0
			},
			{
				"feature_id": "deny_avg",
				"data": 479.0
			},
			{
				"feature_id": "deny_min",
				"data": 128.0
			},
			{
				"feature_id": "deny_sum",
				"data": 1437.0
			},
			{
				"feature_id": "accept_max5",
				"data": 8440.0
			}
		],
		"data_end_time": 1635883920000,
		"data_start_time": 1635883860000,
		"feature_data": [{
				"feature_id": "deny_max1",
				"feature_name": "deny max1",
				"data": 1360.0
			},
			{
				"feature_id": "deny_avg",
				"feature_name": "deny avg",
				"data": 990.0
			},
			{
				"feature_id": "deny_min",
				"feature_name": "deny min",
				"data": 608.0
			},
			{
				"feature_id": "deny_sum",
				"feature_name": "deny sum",
				"data": 2970.0
			},
			{
				"feature_id": "accept_max5",
				"feature_name": "accept max5",
				"data": 9670.0
			}
		],
		"expected_values": [{
			"likelihood": 1.0,
			"value_list": [{
					"feature_id": "deny_max1",
					"data": 905.0
				},
				{
					"feature_id": "deny_avg",
					"data": 479.0
				},
				{
					"feature_id": "deny_min",
					"data": 128.0
				},
				{
					"feature_id": "deny_sum",
					"data": 4847.0
				},
				{
					"feature_id": "accept_max5",
					"data": 15713.0
				}
			]
		}],
		"execution_end_time": 1635898427895,
		"anomaly_grade": 0.5514172746375128,
		"entity": [{
			"name": "service",
			"value": "app_0"
		}],
		"approx_anomaly_start_time": 1635883620000
	},
	"sort": [
		1635898427895
	]
}
```

### Check List


src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -171,34 +171,87 @@ public void getRcfResult(String detectorId, String modelId, double[] point, Acti
...
 
+    /**
+     * normalize total attribution to 1
+     *
+     * @param forest rcf accessor
+     * @param rawAttribution raw attribution scores
+     *
+     * @return normalized attribution
+     */
+    public double[] normalizeAttribution(RandomCutForest forest, double[] rawAttribution) {
```
ohltyler, line 215: This may be covered somewhere else, but want to confirm: how is it guaranteed that the array of attribution values is in the same order as the `List` of features for the detector?

ohltyler: same question for expected value.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -262,6 +393,44 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+            }
+            if (currentTimeAttribution != null) {
+                xContentBuilder.array(CURRENT_TIME_ATTRIBUTION_FIELD, currentTimeAttribution);
+            }
+            if (oldValues != null) {
+                xContentBuilder.array(OLD_VALUES_FIELD, oldValues);
+            }
+            // In OpenSearch, an array of arrays like [ 1, [ 2, 3 ]] is the equivalent of [ 1, 2, 3 ]
+            // Also, even though expectedValuesList if a two-dimensional array, it has only one nested
+            // array inside in current implementation.
```
ohltyler, line 419: Makes sense to have 2D array for future extensibility. Just thinking, in terms of index readability, it could get confusing if later on there is more fields added per expected value (like different confidences, etc.). When indexing, each new field per expected value would add `n` new values to the end of the array, where `n` is the expected value / feature length.

Since there's current efforts to expose the result index to the user, this could make it extra confusing for the user to understand the indexing strategy done here. Did you consider using nested fields, and just index an array of expected value objects? Currently, it would only contain a single `value` field, but if that was to be expanded on in the future, could just add more fields?

src/main/java/org/opensearch/ad/util/ParseUtils.java
```diff
@@ -714,4 +715,13 @@ public static boolean listEqualsWithoutConsideringOrder(List<String> list1, List
         }
         return Objects.equals(set1, set2);
     }
+
+    public static double[] parseDoubleArray(XContentParser parser) throws IOException {
```
ohltyler, line 719: Why is this method needed?

kaituo: It is used in parse double array out of XContent in AnomalyResult.parse method.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: Can you give me an example how nested fields work? Say I have an expected value array [1, 2, 3, 4, 5].  Tomorrow there are two expected value array with different confidence. I need to add [7, 8, 9, 10, 11].  How do I change mapping?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -89,42 +98,107 @@
...
+    "oldValues": [
+    17265.0,
+    4113.142857142857,
+    49.0,
+    86376.0,
+    64878.0
+    ],
+    
+    So oldValues is null when relativeIndex is 0 or startOfAnomaly is true.*/
+    private final double[] oldValues;
```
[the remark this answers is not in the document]
ohltyler: Got it, thanks for clarifying, makes sense. So seems like `oldValues` is more of just an internal way for storing "what values caused this anomaly", but not a cx-friendly value, whereas feature attribution and expected values are cx-friendly, and may possibly use `oldValues` as part of their input when determining their own values.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
ohltyler: I was thinking of it in such a way, that each object in the array would contain all of the expected value info for each corresponding feature.

For example, given features `f1` and `f2`, could have expected values like the following:
```
[
  // info about feature f1
  {
    expectedValue1: 5,
    confidence1: 0.95,
    expectedValue2: 7,
    confidence2: 0.82
  },

  // info about feature f2
  {
    expectedValue1: 20,
    confidence1: 0.90,
    expectedValue2: 16,
    confidence2: 0.80
  }  
]
```

Not sure if this is the right way to go either, though, since changing the mappings in the future may lead to other issues. Your solution is obviously simpler, just thinking about user readability in the future when we open up these results publicly.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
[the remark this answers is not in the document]
kaituo: feature attribution and expected values may not be cx-friendly too as they may correspond to oldValues not  input values.

kaituo:
> Another high-level thought - has there been discussion with UX about how to show expected values? I'd imagine it could be helpful to see those results for non-anomalous results as well, which would allow for showing an additional time series on the feature breakdown charts.
> 
> Simpler / short-term solution can just be showing it paired with anomalous results. And, I guess it could be added to non-anomalous results later on. Just curious

There are no expected values for non-anomalous results as we don't expect otherwise :)

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: Changing the mapping in the future would not be easy. Also, searching with nested fields can be difficult (e.g., recall our discussions about entity field and how complicated of the query behind our facet API is).  So I didn't start with nested fields.

If you think querying nested fields is not a problem, I am willing to go that route.  I am thinking of two ways to work around the changeability issue.  One is to reserve expectedValue1~expectedValue20 in the mapping.  As long as we don't allocate more than 20 expected values (unlikely), we are fine.

Another is to have sth like:

[
  {
    expectedValue: [1,2,3],
    likelihood: 0.95
  },
  {
    expectedValue: [11,12,13],
    likelihood: 0.90
  }  
]

What do you think?

kaituo:
> > > Another high-level thought - has there been discussion with UX about how to show expected values? I'd imagine it could be helpful to see those results for non-anomalous results as well, which would allow for showing an additional time series on the feature breakdown charts.
> > > Simpler / short-term solution can just be showing it paired with anomalous results. And, I guess it could be added to non-anomalous results later on. Just curious
> > 
> > 
> > There are no expected values for non-anomalous results as we don't expect otherwise :)
> 
> Is this because the model can't calculate the expected values for non-anomalous result or some other consideration like performance?

There can be 2 reasons:
First, if it is not an anomaly, input value is expected.  It would be strange if we provide an expected value but say this is not an anomaly.
Second, it is costly to compute expected value.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
sudiptoguha: I'd suggest the nested fields .. in the above the idea is f1 = 1, f2= 2, and f3 = 3 in a scenario with likelihood 0.95 and f1=11, f2=12, and f3=13 in another scenario with likelihood 0.90 (let's not worry about things summing to 1, etc. for this discussion). What this captures is that in either scenarios the f1-f3 gap is in a narrow range. That correlation is lost if we perform a field by field analysis -- separating the confidence, expected value of each field from other fields. If we were interested in single fields -- we should perform 3 different anomaly detections (one for each field). One performs a (multivariate) analysis on several fields together if only there is something to be gained (in this case the gain is the correlated/conditional expectation).

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -21,9 +24,20 @@
     private final double grade;
     private final double confidence;
     private final double rcfScore;
+    private long totalUpdates;
+    private boolean startOfAnomaly;
+    private boolean inHighScoreRegion;
+    private int relativeIndex;
+    private double[] currentTimeAttribution;
+    private double[] oldValues;
+    private double[][] expectedValuesList;
```
ylwu-amzn, line 33: How about add some java doc for some fields not easy to understand like `inHighScoreRegion`, `relativeIndex`, `oldValues`, `expectedValuesList` ?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -89,42 +98,107 @@
...
-        );
-    }
+    // sequence index (the number of updates to RCF) -- it is possible in imputation
+    // that
+    // the number of updates more than the input tuples seen by the overall program
+    private final Long totalUpdates;
+
+    // flag indicating if the anomaly is the start of an anomaly or part of a run of
+    // anomalies
+    private final Boolean startOfAnomaly;
```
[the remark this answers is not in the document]
ylwu-amzn: For HC detector, this only mean the start anomaly for this entity right? I haven't read all code, but seems we need to cache the last anomaly time to tell whether current anomaly is the start anomaly or not.

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
ylwu-amzn: Never mind, find the java doc in AnomalyResult.java.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -89,42 +98,107 @@
...
+
+    // flag indicating if the time stamp is in elevated score region to be
+    // considered as anomaly
+    private final Boolean inHighScoreRegion;
+
+    /**
+     * position of the anomaly vis a vis the current time (can be -ve) if anomaly is
+     * detected late, which can and should happen sometime; for shingle size 1; this
+     * is always 0
```
ylwu-amzn, line 117: If the shingle size is 4, for example shingle is `[0, 0, 1, 0]`, and this shingle detected as anomaly, but actually the anomaly is caused by the third item "1", then the `relativeIndex` will be 2 ?

sudiptoguha: it'll be -1 (relative to current time) in the library - one can always add (shingleSize - 1) to it to get 2 as well

src/main/java/org/opensearch/ad/model/AnomalyResult.java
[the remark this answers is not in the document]
sudiptoguha: Re "is the expected value calculated based on individual data point rather than a whole shingle? Feel confused that we tell a detector interval is anomalous or not based on shingle, but we calculate expected value/attribution based on individual data point (old value)?" ---> it's based on the shingle. In fact it is much harder to predict based on a single value :) without the context. That is why the relative position of *where in the shingle are we most likely to have a problem* is needed (relativeIndex).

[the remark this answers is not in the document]
sudiptoguha: So the "relevant" values should refer to that time slice. For most simple anomalies this is often 0. But there are very good reasons why it's hard to have a generic detector with 0 time lag (and not have a lot of mistakes).

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -21,9 +24,20 @@
     private final double grade;
     private final double confidence;
     private final double rcfScore;
+    private long totalUpdates;
+    private boolean startOfAnomaly;
+    private boolean inHighScoreRegion;
+    private int relativeIndex;
+    private double[] currentTimeAttribution;
```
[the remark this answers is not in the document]
ylwu-amzn: After reading more code, I feel this field named as "currentTimexxx" as the attribution may be not for current time, it may be for some oldValues ? So you name it "currentTimexxx" to indicate the attribution is for current interval aggregated data point, not for some old values?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -89,42 +98,107 @@
...
+    So oldValues is null when relativeIndex is 0 or startOfAnomaly is true.*/
+    private final double[] oldValues;
+
+    // expected values, currently set to maximum 1 expected. In the future, we
+    // might give different expected values with differently likelihood. So
+    // the two-dimensional array allows us to future-proof our applications.
+    // Also, expected values correspond to oldValues if present or current input
+    // point otherwise. If oldValues is present, it will take effort to show this
+    // on UX since we found an anomaly from the past (old values).
+    private final double[][] expectedValuesList;
```
ylwu-amzn, line 193: How about add some examples to help understand what the 2D array will looks like and what each dimension means?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
ylwu-amzn: The main goal is to save less result and avoid updating feature's expected value of same interval multiple times. Just some thought, maybe not quite fit the new anomaly detection expected value logic in model.

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -25,13 +31,44 @@
...
+     * is always 0
+     */
     private int relativeIndex;
-    private double[] currentTimeAttribution;
-    private double[] oldValues;
+
+    // a flattened version denoting the basic contribution of each input variable
+    private double[] relevantAttribution;
+
+    // oldValues is related to relativeIndex and startOfAnomaly. Read the same
```
ohltyler, line 53: nit: `oldValues` -> `pastValues`

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -25,13 +31,44 @@
...
+    // field comment on AnomalyResult.
+    private double[] pastValues;
+
+    // expected values, currently set to maximum 1 expected. In the future, we
+    // might give different expected values with differently likelihood. So
+    // the two-dimensional array allows us to future-proof our applications.
+    // Also, expected values correspond to oldValues if present or current input
+    // point otherwise. If oldValues is present, it will take effort to show this
+    // on UX since we found an anomaly from the past (old values).
```
ohltyler, line 62: `oldValues` -> `pastValues`

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -25,13 +31,44 @@
...
     private double[][] expectedValuesList;
+
+    // likelihood values for the list
+    private double[] likelihoodOfValues;
```
ohltyler, line 66: Is the idea there will be one likelihood value that spans a single set of expected values? In other words, for now, only one likelihood value should be expected? If so can you add some more comments here about that?

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -58,8 +95,9 @@ public ThresholdingResult(
         boolean inHighScoreRegion,
         int relativeIndex,
         double[] currentTimeAttribution,
```
ohltyler, line 97: nit: rename this to `relevantAttribution` also?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -98,31 +100,34 @@
...
-    private final Integer relativeIndex;
+    private final Instant approxAnomalyStartTime;
 
     // a flattened version denoting the basic contribution of each input variable
-    private final double[] currentTimeAttribution;
+    private final List<DataByFeatureId> relavantAttribution;
 
     /*
-    oldValues is related to relativeIndex and startOfAnomaly.
+    relevantValues is related to relativeIndex, startOfAnomaly and anomaly grade.
```
ohltyler, line 130: `relevantValues` -> `pastValues`

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -169,28 +174,47 @@
     71604.0,
     75236.0
     ],
-    Since relativeIndex is -2, we look back baseDimension * 2 and get the oldValues:
+    Since rcf returns relativeIndex is -2, we look back baseDimension * 2 and get the pastValues:
     "startOfAnomaly": true,
     "inHighScoreRegion": true,
-    "relativeIndex": -2,
-    "oldValues": [
+    "relevantValues": [
```
ohltyler, line 180: `relevantValues` -> `pastValues`

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
kaituo: Fixed

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
kaituo: fixed

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
kaituo: yes, will add comments

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
kaituo: renamed

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: fixed

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: fixed

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -169,28 +174,47 @@
...
-    private final double[] oldValues;
+    So pastValues is null when relativeIndex is 0 or startOfAnomaly is true
+    or the current shingle is not an anomaly.
+    
+    In the UX, if pastValues value is null, we can just show attribution/expected
+    value and it is implicit this is due to current input; if pastValues is not
+    null, it means the the attribution/expected values are from an old value
+    (e.g., 2 steps ago with data [1,2,3]) and we can add a text to explain that.
+    */
+    private final List<DataByFeatureId> pastValues;
```
ohltyler, line 196: had a new thought here, would it actually make sense to just include the relevant values here? (rename to `relevantValues`), and include the current values if `relativeIndex` is 0, else populate with past values?

Few reasons for this:
1. unifies the data in the results (always have values in here, regardless of if the anomaly is found late or not)
2. simplifies logic needed from external users building their own dashboards (+ frontend plugin) on getting the current values. it can be retrieved directly from here, rather than having to use the detector's interval and shingle size to calculate the list to retrieve

ohltyler: ^ this would be redundant, since the current values would already be what's included in the feature values, and no extra data fetching would be needed. the current `pastValues` makes more sense here. please disregard.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -253,13 +276,12 @@ public AnomalyResult(
...
         Integer schemaVersion,
         String modelId,
-        Long totalUpdates,
         Boolean startOfAnomaly,
         Boolean inHighScoreRegion,
-        Integer relativeIndex,
-        double[] currentTimeAttribution,
-        double[] oldValues,
-        double[][] expectedValuesList,
+        Instant approAnomalyStartTime,
```
ohltyler, line 281: `approAnomalyStartTime` -> `approxAnomalyStartTime`

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -253,13 +276,12 @@ public AnomalyResult(
...
+        List<DataByFeatureId> relavantAttribution,
```
ohltyler, line 282: `relavantAttribution` -> `relevantAttribution`

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -277,16 +299,172 @@ public AnomalyResult(
...
+        Instant executionEndTime,
+        String error,
+        Entity entity,
+        User user,
+        Integer schemaVersion,
+        String modelId,
+        Boolean startOfAnomaly,
+        Boolean inHighScoreRegion,
+        double[] relevantAttribution,
+        int reletiveIndex,
```
ohltyler, line 359: `reletiveIndex` -> `relativeIndex`

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: As discussed offline, it would be redundant if we have both feature_data and relevantValues since relevantValues is feature_data when there is no anomaly or relative index is 0. if removing feature_data, there would be bwc issue. Also, the front end needs a lot of modification.  if keeping feature_data, our index size will increase.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: fixed

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: fixed

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: fixed

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -316,34 +494,39 @@ public AnomalyResult(StreamInput input) throws IOException {
         this.taskId = input.readOptionalString();
         this.modelId = input.readOptionalString();
 
-        this.totalUpdates = input.readOptionalLong();
         this.startOfAnomaly = input.readOptionalBoolean();
         this.inHighScoreRegion = input.readOptionalBoolean();
-        this.relativeIndex = input.readOptionalInt();
+        // if anomaly is caued by current input, we don't show approximate time
```
ohltyler, line 499: `caued` -> `caused`

ohltyler: can you add a comment on line 124 that this can be `null` if `relativeIndex` is 0?

kaituo: fixed

src/main/resources/mappings/anomaly-results.json
```diff
@@ -106,6 +106,60 @@
     },
     "model_id": {
       "type": "keyword"
+    },
+    "startOfAnomaly": {
```
ylwu-amzn, line 110: Can we use snake case like `start_of_anomaly` ?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -63,13 +66,12 @@
...
     public static final String START_OF_ANOMALY_FIELD = "startOfAnomaly";
     public static final String IN_HIGH_SCORE_REGION_FIELD = "inHighScoreRegion";
-    public static final String RELATIVE_INDEX_FIELD = "relativeIndex";
-    public static final String CURRENT_TIME_ATTRIBUTION_FIELD = "currentTimeAttribution";
-    public static final String OLD_VALUES_FIELD = "oldValues";
-    public static final String EXPECTED_VAL_LIST_FIELD = "expectedValuesList";
+    public static final String APPROX_ANOMALY_START_FIELD = "approx_anomaly_start_time";
+    public static final String RELEVANT_ATTRIBUTION_FIELD = "relavantAttribution";
+    public static final String PAST_VALUE_FIELD = "past_value";
+    public static final String EXPECTED_VALUE_FIELD = "expected_value";
```
ohltyler, line 74: since these will be arrays, how about make them plural? `past_values` / `expected_values`?

src/main/java/org/opensearch/ad/model/DataByFeatureId.java
```diff
@@ -0,0 +1,114 @@
...
+
+            switch (fieldName) {
+                case FEATURE_ID_FIELD:
+                    featureId = parser.text();
+                    break;
+                case DATA_FIELD:
+                    data = parser.doubleValue();
+                    break;
+                default:
+                    break;
```
ylwu-amzn, line 79: Should we add `parser.skipChildren();` before break?

src/main/java/org/opensearch/ad/transport/RCFResultResponse.java
```diff
@@ -62,6 +64,7 @@ public RCFResultResponse(
         Integer relativeIndex,
         double[] oldValues,
```
ohltyler, line 65: `oldValues` -> `pastValues`

src/main/java/org/opensearch/ad/transport/RCFResultTransportAction.java
```diff
@@ -80,15 +80,16 @@ protected void doExecute(Task task, RCFResultRequest request, ActionListener<RCF
                                         result.getRcfScore(),
                                         result.getConfidence(),
                                         result.getForestSize(),
-                                        result.getCurrentTimeAttribution(),
+                                        result.getRelevantttribution(),
```
ohltyler, line 83: `getRelevantttribution()` -> `getRelevantAttribution()`

src/main/resources/mappings/anomaly-results.json
```diff
@@ -107,26 +107,56 @@
...
       "type": "boolean"
     },
     "inHighScoreRegion": {
       "type": "boolean"
     },
-    "relativeIndex": {
-      "type": "integer"
+    "approx_anomaly_start_time": {
+      "type": "date",
+      "format": "strict_date_time||epoch_millis"
```
ohltyler, line 118: can you explain why it can be `strict_date_time` or `epoch_millis`?

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -21,9 +30,51 @@
...
+    private boolean startOfAnomaly;
+
+    // flag indicating if the time stamp is in elevated score region to be
+    // considered as anomaly
+    private boolean inHighScoreRegion;
+
+    /**
+     * position of the anomaly vis a vis the current time (can be -ve) if anomaly is
+     * detected late, which can and should happen sometime; for shingle size 1; this
+     * is always 0
```
ylwu-amzn, line 46: I'm afraid others who don't read this PR's comments will have difficulty to understand this. How about add some example ?

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
ylwu-amzn: I remember @ohltyler had some concern about this part and we decided to change to some json format like 
```
{
 likehood: {feature id: expected value}
}
```

Maybe the change is for another part ?

src/main/resources/mappings/anomaly-results.json
kaituo: sure

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
ylwu-amzn: How about add more details in comments, like the expected value is only calculated for anomalous detection intervals, and will generate expected value for each feature if detector has multiple features?

If RCF model detects anomaly for some interval for multiple times, will we only persist the expected value for the first time? For example, shingle size 4, the first shingle is [t1, t2, t3, t4], second shingle is [t2, t3, t4, t5], RCF detect anomaly for t3 in first shingle, then detect anomaly again for t3 in second shingle, will RCF recalculate expected value for t3 in second shingle and will we store that recalculated expected value?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
kaituo: yes, changed

src/main/java/org/opensearch/ad/model/DataByFeatureId.java
kaituo: so skipChildren means if we saw some unknown field, we skip its children?

src/main/java/org/opensearch/ad/transport/RCFResultResponse.java
kaituo: fixed

src/main/java/org/opensearch/ad/transport/RCFResultTransportAction.java
kaituo: fixed

src/main/resources/mappings/anomaly-results.json
kaituo: I am following other date like execution_end_time

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
kaituo: added

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
kaituo: > I remember @ohltyler had some concern about this part and we decided to change to some json format like
> 
> ```
> {
>  likehood: {feature id: expected value}
> }
> ```
> 
> Maybe the change is for another part ?

The changes are done in AnomalyResult class.

kaituo: > How about add more details in comments, like the expected value is only calculated for anomalous detection intervals, and will generate expected value for each feature if detector has multiple features?

added details and comments.

> 
> If RCF model detects anomaly for some interval for multiple times, will we only persist the expected value for the first time? For example, shingle size 4, the first shingle is [t1, t2, t3, t4], second shingle is [t2, t3, t4, t5], RCF detect anomaly for t3 in first shingle, then detect anomaly again for t3 in second shingle, will RCF recalculate expected value for t3 in second shingle and will we store that recalculated expected value?

RCF will recalculate expected value for t3 and we will store that.

src/main/java/org/opensearch/ad/model/DataByFeatureId.java
ylwu-amzn: Yes, the unknown field and it's children should be ignored, this is what AnomalyDetector does to not parse unnecessary fields.

kaituo: fixed.

src/main/java/org/opensearch/ad/ml/ThresholdingResult.java
```diff
@@ -59,20 +99,97 @@ public double getRcfScore() {
...
+        return new ToStringBuilder(this)
+            .append("grade", grade)
+            .append("confidence", confidence)
+            .append("rcfScore", rcfScore)
+            .append("totalUpdates", totalUpdates)
+            .append("startOfAnomaly", startOfAnomaly)
+            .append("inHighScoreRegion", inHighScoreRegion)
+            .append("relativeIndex", relativeIndex)
+            .append("currentTimeAttribution", Arrays.toString(currentTimeAttribution))
+            .append("oldValues", Arrays.toString(oldValues))
```
sudiptoguha: This is changing right?

sudiptoguha: I saw the old -> past -> relevant discussion ..

kaituo: yes, changed to pastValues

kaituo: > I saw the old -> past -> relevant discussion ..

Yes, not an easy choice.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -210,6 +214,22 @@ public AnomalyDetectionIndices(
...
+            // we have already initiated the field
+            return;
+        }
+        String resultMapping = getAnomalyResultMappings();
+
+        Map<String, Object> asMap = XContentHelper.convertToMap(new BytesArray(resultMapping), false, XContentType.JSON).v2();
+        Object properties = asMap.get(CommonName.PROPERTIES);
+        if (properties instanceof Map) {
+            AD_RESULT_FIELD_CONFIGS = (Map<String, Object>) properties;
+        }
```
ylwu-amzn, line 232: If properties is not Map, that means the result index mapping JSON file is wrong, then we should throw exception?

kaituo: the exception will be caught by the caller's catch method and log the error and return false.  How about log an error if the index mapping json file is wrong to save another round of exception creation/caught?

ylwu-amzn: I'm ok to just log error and let the caller to throw exception, make sure the error message make sense to the user.

kaituo: as discussed offline, I logged the error in the initResultMapping method.

ylwu-amzn:
The code looks good to me, but have some concerns:

1. We may need some BWC effort if some field/model not stabilized, like Kaituo said `relevant_attribution` may add likehood dimension. We can’t update this field after release, may need to add a new field to support likehood in future, then user/AD frontend need to query two fields.
2. The field `in_high_score_region` indicates the anomaly score is high enough or not, not sure if we will add more score level like medium score region. If it’s possible, add a keyword/int filed `score_region` may be more flexible, otherwise, we. need to add more new fields in future.
3. The field `approx_anomaly_start_time` is confusing. It’s estimated time as TRCF doesn’t store time info. As we discussed offline, if the data is sparse, the data pints in one shingle won't be consecutive on time. For example if shingle size is 4, and interval t3 and t4 have no data, the shingle will be [t1, t2, t5, t6], if anomaly found for t2, anomaly then relative index is -3. We will calculate the `approx_anomaly_start_time` based on `t6 -3*interval`, it will be t3, not t2. It looks confusing. Maybe add more details in doc

ylwu-amzn (dismissed):

kaituo:
> The code looks good to me, but have some concerns:
> 
>     1. We may need some BWC effort if some field/model not stabilized, like Kaituo said `relevant_attribution` may add likehood dimension. We can’t update this field after release, may need to add a new field to support likehood in future, then user/AD frontend need to query two fields.
> 
relevant_attribution won't add likelihood dimension.  Expected value has the likelihood dimension.  Current mapping has already considered the possibility that we may have different sets of expected values with different likelihoods.

>     2. The field `in_high_score_region` indicates the anomaly score is high enough or not, not sure if we will add more score level like medium score region. If it’s possible, add a keyword/int filed `score_region` may be more flexible, otherwise, we need to add more new fields in future.
@sudiptoguha Can you help answer the question?

>     3. The field `approx_anomaly_start_time` is confusing. It’s estimated time as TRCF doesn’t store time info. As we discussed offline, if the data is sparse, the data pints in one shingle won't be consecutive on time. For example if shingle size is 4, and interval t3 and t4 have no data, the shingle will be [t1, t2, t5, t6], if anomaly found for t2, anomaly then relative index is -3. We will calculate the `approx_anomaly_start_time` based on `t6 -3*interval`, it will be t3, not t2. It looks confusing. Maybe add more details in doc.
Yeah, will do.

kaituo:
>     2. The field `in_high_score_region` indicates the anomaly score is high enough or not, not sure if we will add more score level like medium score region. If it’s possible, add a keyword/int filed `score_region` may be more flexible, otherwise, we need to add more new fields in future.
Discussed with Sudipto.  I am gonna remove the field in_high_score_region.

ylwu-amzn (approved):

kaituo:
have to push again to merge from mainline.

ylwu-amzn (approved):

sudiptoguha (approved):
Looks good. Just a nit -- should not be a blocker for merging; the top example in the PR header still shows the older views of normal and anomalous data (the field names have since changed in the anomalous; I still see 	"start_of_anomaly: true,
	in_high_score_region: true," it is also possible I'm not looking at the current commit) -- we will probably forget all the detailed discussion, and someone may just look at the top example in the PR.

kaituo:
> start_of_anomaly

You are right.  I removed start_of_anomaly and in_high_score_region from the commit message.

PR #276, opened 2021-10-27
support storing anomaly result to custom index

### LINKED ISSUE #270: Anomaly Detection: Provide option to store anomaly result to custom index
From ODFE 1.11, AD plugin moved all AD indices to system index list. That means user can’t query AD result index directly. We build search APIs for AD results, but user can’t use it to build dashboard.

Without the easy access to AD result data, we must build kinds of visualization to show AD result to user. But it’s hard and time consuming to develop custom visualization for kinds of use cases. 

We plan to add one option to let user specify which index they want to use to store the detector’s AD result. If not set, we will store in default AD result system index. User should own the index management and the data lifecycle by themselves, suggest using ISM plugin https://opensearch.org/docs/latest/im-plugin/ism/index/

ylwu-amzn:
### Description
From ODFE 1.11, AD plugin moved all AD indices to system index list. That means user can’t query AD result index directly. We build search APIs for AD results, but user can’t use it to build dashboard.

Without the easy access to AD result data, we must build kinds of visualization to show AD result to user. But it’s hard and time consuming to develop custom visualization for kinds of use cases.

We plan to add one option to let user specify which index they want to use to store the detector’s AD result. If not set, we will store in default AD result system index. User should own the index management and the data lifecycle by themselves, suggest using ISM plugin https://opensearch.org/docs/latest/im-plugin/ism/index/. And user should maintain the permission control of these result indices by themselves.
 
### Issues Resolved
#270
 
### Check List


kaituo (commented):
partial comments

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -50,18 +52,37 @@
...
     public static final String DATA_END_TIME_FIELD = "data_end_time";
-    private static final String EXECUTION_START_TIME_FIELD = "execution_start_time";
+    public static final String EXECUTION_START_TIME_FIELD = "execution_start_time";
     public static final String EXECUTION_END_TIME_FIELD = "execution_end_time";
     public static final String ERROR_FIELD = "error";
     public static final String ENTITY_FIELD = "entity";
     public static final String USER_FIELD = "user";
     public static final String TASK_ID_FIELD = "task_id";
     public static final String MODEL_ID_FIELD = "model_id";
+    public static final ImmutableMap<String, String> AD_RESULT_FIELD_CONFIGS = ImmutableMap
```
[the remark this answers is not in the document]
ylwu-amzn: Yeah, I planned to do that at first, but the index mapping object we got from Java may output a different String for example the sequence of some sub field changed compared to the index mapping file, it takes some more effort to sort out such problems/edge cases.

Prefer to do that in next phase, including the custom result index mapping upgrade.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -281,10 +294,132 @@ public boolean doesAnomalyDetectorJobIndexExist() {
...
+     * @return true if result index mapping is valid
+     */
+    public boolean isValidResultIndex(String resultIndex) {
+        IndexMetadata indexMetadata = clusterService.state().metadata().index(resultIndex);
+        Map<String, Object> indexMapping = indexMetadata.mapping().sourceAsMap();
+        String propertyName = "properties";
+        if (!indexMapping.containsKey(propertyName)) {
+            return false;
+        }
+        LinkedHashMap<String, Object> mapping = (LinkedHashMap<String, Object>) indexMapping.get(propertyName);
```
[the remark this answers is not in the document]
ylwu-amzn: will add some check

src/main/java/org/opensearch/ad/transport/handler/AnomalyIndexHandler.java
```diff
@@ -95,45 +93,50 @@ public AnomalyIndexHandler(
...
         this.clusterService = clusterService;
     }
 
-    /**
-     * Since the constructor needs to provide injected value and Guice does not allow Boolean to be there
-     * (claiming it does not know how to instantiate it), caller needs to manually set it to true if
-     * it want to save to a specific doc.
-     * @param fixedDoc whether to save to a specific doc Id
-     */
-    public void setFixedDoc(boolean fixedDoc) {
```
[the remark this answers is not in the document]
ylwu-amzn: sure

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -266,6 +279,7 @@ public void writeTo(StreamOutput output) throws IOException {
         } else {
             output.writeBoolean(false);
         }
+        output.writeOptionalString(resultIndex);
```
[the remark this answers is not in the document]
ylwu-amzn: `AnomalyDetector` used in these transport requests/responses
```
ADBatchAnomalyResultRequest, we only send to nodes with same AD version
ForwardADTaskRequest,  we only send to nodes with same AD version
IndexAnomalyDetectorRequest / IndexAnomalyDetectorResponse,  send to local node 
PreviewAnomalyDetectorRequest / PreviewAnomalyDetectorResponse, send to local node
GetAnomalyDetectorResponse,  request only sent to local node
```

src/main/java/org/opensearch/ad/model/AnomalyDetectorJob.java
```diff
@@ -156,6 +166,7 @@ public void writeTo(StreamOutput output) throws IOException {
         } else {
             output.writeBoolean(false); // user does not exist
         }
+        output.writeOptionalString(resultIndex);
```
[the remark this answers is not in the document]
ylwu-amzn: `AnomalyDetectorJob` only used in this transport response

```
GetAnomalyDetectorResponse, request only sent to local node
```

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -281,10 +294,132 @@ public boolean doesAnomalyDetectorJobIndexExist() {
...
+                }));
+            }, exception -> {
+                logger.error("Failed to write dummy AD result to result index " + resultIndex, exception);
+                listener.onFailure(exception);
+            }));
+        } catch (Exception e) {
+            logger.error("Failed to create detector with custom result index " + resultIndex, e);
+            listener.onFailure(e);
+        }
+    }
```
ohltyler, line 364: Seems weird to be indexing and deleting to the custom index. Why isn't the check on the index mappings in `isValidResultIndex` enough here?

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -281,10 +294,132 @@ public boolean doesAnomalyDetectorJobIndexExist() {
...
+    public boolean isValidResultIndex(String resultIndex) {
+        IndexMetadata indexMetadata = clusterService.state().metadata().index(resultIndex);
+        Map<String, Object> indexMapping = indexMetadata.mapping().sourceAsMap();
+        String propertyName = "properties";
+        if (!indexMapping.containsKey(propertyName)) {
+            return false;
+        }
+        LinkedHashMap<String, Object> mapping = (LinkedHashMap<String, Object>) indexMapping.get(propertyName);
+
+        boolean correctResultIndexMappig = true;
```
ohltyler, line 412: nit: `Mappig` -> `Mapping`

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -192,6 +198,10 @@ public AnomalyDetector(
         this.categoryFields = categoryFields;
         this.user = user;
         this.detectorType = isMultientityDetector(categoryFields) ? MULTI_ENTITY.name() : SINGLE_ENTITY.name();
+        this.resultIndex = Strings.trimToNull(resultIndex);
+        if (this.resultIndex != null && !this.resultIndex.startsWith(CUSTOM_RESULT_INDEX_PREFIX)) {
+            throw new IllegalArgumentException(INVALID_RESULT_INDEX_PREFIX);
+        }
```
ohltyler, line 204: @amitgalitz @ylwu-amzn will this custom-result-index check make it into the initial validation API?

src/main/java/org/opensearch/ad/model/AnomalyResult.java
```diff
@@ -223,9 +244,13 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
-            .field(DATA_START_TIME_FIELD, dataStartTime.toEpochMilli())
-            .field(DATA_END_TIME_FIELD, dataEndTime.toEpochMilli())
-            .field(CommonName.SCHEMA_VERSION_FIELD, schemaVersion);
+            .field(SCHEMA_VERSION_FIELD, schemaVersion);
+        if (dataStartTime != null) {
+            xContentBuilder.field(DATA_START_TIME_FIELD, dataStartTime.toEpochMilli());
+        }
+        if (dataEndTime != null) {
+            xContentBuilder.field(DATA_END_TIME_FIELD, dataEndTime.toEpochMilli());
+        }
```
ohltyler, line 253: Can you explain these added checks? What has changed to allow passing null data start/end times?

src/main/java/org/opensearch/ad/rest/RestSearchAnomalyResultAction.java
```diff
@@ -41,6 +52,27 @@ public RestSearchAnomalyResultAction() {
 
     @Override
     public String getName() {
-        return SEARCH_ANOMALY_DETECTOR_ACTION;
+        return SEARCH_ANOMALY_RESULT_ACTION;
+    }
+
+    @Override
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
```
ohltyler, line 59: Is a similar change needed on the delete anomaly result rest action?

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
```diff
@@ -164,22 +166,41 @@ public IndexAnomalyDetectorActionHandler(
...
+            // use default detector result index which is system index
+            if (resultIndex == null) {
+                createOrUpdateDetector();
+                return;
+            }
+
+            // user custom result index
+            anomalyDetectionIndices.initCustomResultIndexAndExecute(resultIndex, () -> createOrUpdateDetector(), listener);
+        } catch (Exception e) {
+            logger.error("Failed to create index " + resultIndex, e);
```
ohltyler, line 181: Maybe tune this error message? Seems the failures could be either issues creating the custom result index, as well as any errors found in `createOrUpdateDetector()`

ohltyler, line 181: Maybe tune this error message? Couldn't this encapsulate more than just a failure to create the index? For example, in `createOrUpdateDetector()`, it calls `prepareAnomalyDetectorIndexing()`,  which does a handful of other operations (validation, creation, etc.).

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -1115,17 +1118,81 @@ private void detectAnomaly(
             }
         }
 
+        String user;
+        List<String> roles;
+        if (adTask.getUser() == null) {
+            user = "";
+            roles = settings.getAsList("", ImmutableList.of("all_access", "AmazonES_all_access"));
```
ohltyler, line 1125: Why is it necessary to add these admin roles? The only other place I see it used is AnomalyDetectorJobRunner.

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
```diff
@@ -833,6 +833,15 @@ public boolean isHCTaskCoordinatingNode(String detectorId) {
...
         hcTaskCache.setTopEntityCount(count);
+        ADTaskSlotLimit adTaskSlotLimit = detectorTaskSlotLimit.get(detectorId);
+
+        if (count != null && adTaskSlotLimit != null) {
+            Integer detectorTaskSlots = adTaskSlotLimit.getDetectorTaskSlots();
+            if (detectorTaskSlots != null && detectorTaskSlots > count) {
+                logger.debug("Scale down task slots from {} to the same as top entity count {}", detectorTaskSlots, count);
+                adTaskSlotLimit.setDetectorTaskSlots(count);
+            }
+        }
```
ohltyler, line 844: Is this to fix some bug?

src/main/java/org/opensearch/ad/transport/ADResultBulkTransportAction.java
```diff
@@ -94,42 +94,48 @@ protected void doExecute(Task task, ADResultBulkRequest request, ActionListener<
...
-    private void addResult(BulkRequest bulkRequest, AnomalyResult result) {
+    private void addResult(BulkRequest bulkRequest, AnomalyResult result, String resultIndex) {
         try (XContentBuilder builder = jsonBuilder()) {
-            IndexRequest indexRequest = new IndexRequest(indexName).source(result.toXContent(builder, RestHandlerUtils.XCONTENT_WITH_TYPE));
+            String index = resultIndex == null ? indexName : resultIndex;
+            IndexRequest indexRequest = new IndexRequest(index).source(result.toXContent(builder, RestHandlerUtils.XCONTENT_WITH_TYPE));
             bulkRequest.add(indexRequest);
         } catch (IOException e) {
-            LOG.error(String.format(Locale.ROOT, "Failed to prepare bulk %s", indexName), e);
+            LOG.error(String.format(Locale.ROOT, "Failed to prepare bulk %s", resultIndex), e);
```
ohltyler, line 138: From line 134, seems `resultIndex` could be null, like in the default result index case. Can tune this error statement to be either `indexName` or `resultIndex`

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -11,32 +11,163 @@
...
+                    for (String index : targetIndices) {
+                        multiSearchRequest
+                            .add(new SearchRequest(index).source(new SearchSourceBuilder().query(new MatchAllQueryBuilder()).size(0)));
+                    }
+                    List<String> readableIndices = new ArrayList<>();
+                    readableIndices.add(ALL_AD_RESULTS_INDEX_PATTERN);
+
+                    context.restore();
+                    // Send multiple search to check which index uer has permission to read. If search all indices directly,
+                    // search request will throw exception if user has no permission to search any index.
```
ohltyler, line 140: I'm unsure why this is necessary. Why can't a single search over all indices work? And if there's some permissions error, then just propagate the error?

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -11,32 +11,163 @@
...
         this.searchHandler = searchHandler;
+        this.clusterService = clusterService;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
+        this.client = client;
     }
 
     @Override
     protected void doExecute(Task task, SearchRequest request, ActionListener<SearchResponse> listener) {
-        searchHandler.search(request, listener);
+        String[] indices = request.indices();
```
ohltyler, line 76: Seems all of this logic (lines 76 - 170) will need to be reused in a few places, such as multi-category filtering API and (partially) for the delete results API, right?

src/main/java/org/opensearch/ad/transport/handler/AnomalyIndexHandler.java
```diff
@@ -95,45 +93,50 @@ public AnomalyIndexHandler(
...
-                        }
-                    }));
+            if (customIndexName != null) {
+                // Only create custom AD result index when create detector, won’t recreate custom AD result index in realtime
+                // job and historical analysis later if it’s deleted. If user delete the custom AD result index, and AD plugin
+                // recreate it, that may bring confusion.
+                if (!anomalyDetectionIndices.doesIndexExist(customIndexName)) {
+                    throw new EndRunException(detectorId, CAN_NOT_FIND_RESULT_INDEX + customIndexName, true);
+                }
+                if (!anomalyDetectionIndices.isValidResultIndex(customIndexName)) {
```
ohltyler, line 118: What happens if an index pattern tries to be set here / some name with a wildcard? Tried to test on your test domain but seems to be broken.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
amitgalitz: I don't have this currently as part of the backend. We can discuss offline if it is practical to open another PR to add this validation check but not sure if there is time for 1.2

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
ylwu-amzn: User may have no write permission on custom result index. Talked with security plugin team, seems no easy way to verify if user has write permission. So just tried to write and delete a dummy anomaly result to verify.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
ylwu-amzn: Seems not so big change to add this to validation API, but I'm ok to leave it to 1.3 if no time.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
ylwu-amzn: In normal AD result, we always pass data start/end times. In custom result index, we need to write/delete a dummy AD result to verify if user has write permission to the custom result index. Just pass in null start/end time for this dummy anomaly result to make sure it won't be queried by  mistake.

src/main/java/org/opensearch/ad/rest/RestSearchAnomalyResultAction.java
ylwu-amzn: Discussed with manager, we planned to let user owns the result deletion by themselves. Actually, we rolled over default result index with some strategy like if default result index doc size reach some limit, we will roll over it and create a new index to store new anomaly result. We won't support such roll over for custom result index either. 
So user need to maintain the custom result index by themselves. We will show callout message to warn user and suggest to use ISM plugin to manage custom result index.

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
ylwu-amzn: Actually we don't need to catch exception here as exception already caught by `initCustomResultIndexAndExecute` and `createOrUpdateDetector`. Will remove this try catch to make code clean.

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: It's possible that user create domain with security disabled, then enable security. This is for BWC, for old detectors which created when security disabled, the user in AD task will be null.

src/main/java/org/opensearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Not a bug, this is to minimize the task slot usage. For example, we assign 10 task slots to HC detector, but actually the detector only has 3 top entities, then we should scale down task slots assigned to this HC from 10 to 3, so we can release 7 task slots for other detector's historical analysis.

src/main/java/org/opensearch/ad/transport/ADResultBulkTransportAction.java
ylwu-amzn: good catch, will fix

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
ylwu-amzn: Check the answer for comment of class `RestSearchAnomalyResultAction.java `

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
ylwu-amzn: I guess you mean we can send some search request like this `new SearchRequest(CUSTOM_RESULT_INDEX_PREFIX + "*")`. But this request will return security exception directly without telling us which index the user has no permission to read. From security plugin team, this is to avoid leaking security related information so user won't know which index they have permission and which they don't. 

Discussed with security plugin team, no easy way to solve this. So just use multiple search to check which index the user has no permission to read.

src/main/java/org/opensearch/ad/transport/handler/AnomalyIndexHandler.java
ylwu-amzn: Index name should not contain "*", OpenSearch will fail to create index with name including "*"

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
ylwu-amzn: Good catch, will fix

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
amitgalitz: If this is the only "blocker" check happening then I could definitely add this in after PR's are merged as code currently doesn't have these variables and etc.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
ohltyler: Yeah, just seems weird, in case it fails to delete for example, and the user sees some dummy result mixed in with their results. Have you considered security owning validation on write permissions? Seems more appropriate to be owned by them.

Obviously with the given timeline, this is fine with me (or even removing the write check altogether). Just thinking for future use cases.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
ohltyler: I see, can you add a comment in the code about this?

src/main/java/org/opensearch/ad/rest/RestSearchAnomalyResultAction.java
ohltyler: That's fair, agree the custom index should be owned by the user. May cause some user confusion if they try to run the delete results API on a detector with a custom result index. Might be worth calling that out in the API docs.

ylwu-amzn: Yes, will ask tech writer to help add these details to AD doc

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
ylwu-amzn: sure, will remove this debug log

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
ohltyler: Right, by "security owning validation on write permissions", I meant more about improved functionality on the security end to have such an interface to check index read/write permissions. In that case, it could just be an extra check during detector creation, where the result index permissions are checked, to see if such user has write permissions.

Since that's not available, I'm ok with the current way of injecting and removing dummy values. Agreed it is probably best to fail during creation phase than write anomaly result phase, where it would be more work for the user.

src/main/java/org/opensearch/ad/model/AnomalyResult.java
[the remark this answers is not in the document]
kaituo: Can you create an issue?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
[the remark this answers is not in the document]
kaituo: Real time detector does not send AnomalyDetector across the wire.  Is the code in the reader part still needed:

```
public AnomalyDetector(StreamInput input) throws IOException {
...
if (input.available() > 0) {
            resultIndex = input.readOptionalString();
        }
```

src/main/java/org/opensearch/ad/model/AnomalyDetectorJob.java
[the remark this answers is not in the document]
kaituo: Is the following code still needed?
```
public AnomalyDetectorJob(StreamInput input) throws IOException {
...
if (input.available() > 0) {
            resultIndex = input.readOptionalString();
        }
```

ohltyler (approved):
LGTM!

src/main/java/org/opensearch/ad/model/AnomalyResult.java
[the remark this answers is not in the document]
ylwu-amzn: https://github.com/opensearch-project/anomaly-detection/issues/282

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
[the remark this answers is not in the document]
ylwu-amzn: Thanks for confirming. Ideally we don't need this. How about we keep this now in case we missed any edge case, this could help not reading beyond the input stream.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -105,11 +110,15 @@
...
     private User user;
     private String detectorType;
+    private String resultIndex;
 
     // TODO: support backward compatibility, will remove in future
     @Deprecated
     private DetectionDateRange detectionDateRange;
 
+    public static final int MAX_RESULT_INDEX_NAME_SIZE = 255;
+    public static final String RESULT_INDEX_NAME_PATTERN = "^[a-z0-9_-]*$";
```
kaituo, line 120: Does this pattern mean a result index name should start with a-z, 0-9, or _ or -? If so, according to https://discuss.elastic.co/t/index-name-type-name-and-field-name-rules/133039, an index name cannot start with -, _,.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -192,6 +203,27 @@ public AnomalyDetector(
...
+            return null;
+        }
+        if (!resultIndex.startsWith(CUSTOM_RESULT_INDEX_PREFIX)) {
+            return INVALID_RESULT_INDEX_PREFIX;
+        }
+        if (resultIndex.length() > MAX_RESULT_INDEX_NAME_SIZE) {
+            return INVALID_RESULT_INDEX_NAME_SIZE;
+        }
+        if (!resultIndex.matches(RESULT_INDEX_NAME_PATTERN)) {
+            return INVALID_CHAR_IN_RESULT_INDEX_NAME;
```
kaituo, line 224: Can this branch ever be true when resultIndex.startsWith(CUSTOM_RESULT_INDEX_PREFIX) is true?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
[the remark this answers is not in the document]
kaituo: Can you add comments on why we need bwc in stream read but not in stream write? Otherwise, it may cause confusion for later maintenance. The same applies to AnomalyDetectorJob.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -286,25 +289,56 @@ public void startDetector(
...
     ) {
+        // upgrade index mapping of AD default indices
+        detectionIndices.update();
+
         getDetector(detectorId, (detector) -> {
             if (!detector.isPresent()) {
                 listener.onFailure(new OpenSearchStatusException(FAIL_TO_FIND_DETECTOR_MSG + detectorId, RestStatus.NOT_FOUND));
                 return;
             }
             if (validateDetector(detector.get(), listener)) { // validate if detector is ready to start
```
kaituo, line 303: If validateDetector returns false, would listener return?

src/main/java/org/opensearch/ad/model/AnomalyDetectorJob.java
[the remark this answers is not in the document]
ylwu-amzn: I think you mean remove this line only `if (input.available() > 0) {`. 
Yeah, we can remove this line as don't see any BWC issue.

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -11,32 +11,159 @@
...
+        // Set query indices as default result indices, will check custom result indices permission and add
+        // custom indices which user has search permission later.
+        request.indices(ALL_AD_RESULTS_INDEX_PATTERN);
```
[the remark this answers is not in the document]
ylwu-amzn: AD Dashboard and detector list will read all detectors' result, so we need to search both default result index and custom result index.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
ylwu-amzn: This is some old code, will return in `validateDetector` , will add more comment.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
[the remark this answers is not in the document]
ylwu-amzn: Did more checking, don't see any BWC issue for AnomalyDetector, will remove this checking `if (input.available() > 0) {`.

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -11,32 +11,159 @@
...
+
+        Set<String> customResultIndices = new HashSet<>();
+        if (indices != null && indices.length > 0) {
```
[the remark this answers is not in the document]
ylwu-amzn: Yes, good catch, not a bug, but should be good to remove it.

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -11,32 +11,159 @@
...
+                    MultiSearchRequest multiSearchRequest = new MultiSearchRequest();
+                    for (String index : targetIndices) {
+                        multiSearchRequest
+                            .add(new SearchRequest(index).source(new SearchSourceBuilder().query(new MatchAllQueryBuilder()).size(0)));
+                    }
+                    List<String> readableIndices = new ArrayList<>();
+                    readableIndices.add(ALL_AD_RESULTS_INDEX_PATTERN);
+
+                    context.restore();
+                    // Send multiple search to check which index uer has permission to read. If search all indices directly,
```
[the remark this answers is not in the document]
ylwu-amzn: sure, will fix

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -1115,17 +1118,84 @@ private void detectAnomaly(
...
+            storeAnomalyResultAndRunNextPiece(
+                adTask,
+                pieceEndTime,
+                dataStartTime,
+                dataEndTime,
+                interval,
+                internalListener,
+                anomalyResults,
+                resultIndex,
+                injectSecurity
```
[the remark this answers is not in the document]
ylwu-amzn: Sure, will add `ActionListener.runBefore`.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
ylwu-amzn: Actually I may forget to change the index mapping before pushing commit last night, I remember it should be "[a-z0-9_-]+"

Hard to find out exact pattern for OpenSearch index. We always add CUSTOM_RESULT_INDEX_PREFIX "opensearch-ad-plugin-result-" to custom index mapping.  This pattern here is to do some preliminary check, a complete index name check will be done by OpenSearch create index request if this name regex matches.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
ylwu-amzn: I think you mean this test case, check `AnomalyDetectorTests#testValidateResultIndex`
```
    errorMessage = AnomalyDetector.validateResultIndex(CUSTOM_RESULT_INDEX_PREFIX + "abc#");
    assertEquals(INVALID_CHAR_IN_RESULT_INDEX_NAME, errorMessage);
```

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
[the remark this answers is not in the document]
ylwu-amzn: If we still have time before 1.2 release, I will add one parameter to let client specify if only need to search custom result index

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -215,21 +217,42 @@ public AbstractAnomalyDetectorActionHandler(
...
+            createOrUpdateDetector();
+            return;
+        }
+
+        // user custom result index
+        anomalyDetectionIndices.initCustomResultIndexAndExecute(resultIndex, () -> createOrUpdateDetector(), listener);
+    }
+
+    private void createOrUpdateDetector() {
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
amitgalitz, line 240: What was the reason you moved stashContext() from IndexAnomalyDetectorTransportAction to here

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -281,10 +294,140 @@ public boolean doesAnomalyDetectorJobIndexExist() {
...
-    public boolean doesAnomalyResultIndexExist() {
+    public boolean doesDefaultAnomalyResultIndexExist() {
         return clusterService.state().metadata().hasAlias(CommonName.ANOMALY_RESULT_INDEX_ALIAS);
     }
 
+    public boolean doesIndexExist(String indexName) {
+        return clusterService.state().metadata().hasIndex(indexName);
+    }
+
+    public <T> void initCustomResultIndexAndExecute(String resultIndex, AnomalyDetectorFunction function, ActionListener<T> listener) {
```
amitgalitz, line 305: Is it correct to say all validation logic for user to create detector with custom result index is in inside the methods added in AnomalyDetectionIndices class or is there anything additional?

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
ylwu-amzn: If not move to here, all clients which needs to call `IndexAnomalyDetectorTransportAction` need to stash context first, and need to catch exception from all clients. More reasonable to move to here as it's always required to access detector index.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
ylwu-amzn: And another place is in AnomalyDetector, we will check if result index name match `RESULT_INDEX_NAME_PATTERN`. This is the first release, we don't have BWC issue like detector name.

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -215,21 +217,42 @@ public AbstractAnomalyDetectorActionHandler(
...
+    public void start() {
+        String resultIndex = anomalyDetector.getResultIndex();
+        // use default detector result index which is system index
+        if (resultIndex == null) {
+            createOrUpdateDetector();
+            return;
+        }
+
+        // user custom result index
+        anomalyDetectionIndices.initCustomResultIndexAndExecute(resultIndex, () -> createOrUpdateDetector(), listener);
```
amitgalitz, line 236: Before I only called on anomalyDetectionIndices.initAnomalyDetectorIndex if this.isDryRun was false which basically means the call was either Create/update API and not validate. Do we want to still call this even when just validating, basically it could create the index even before creation itself and just when validating.

ohltyler (approved):

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
ylwu-amzn: Oh, just saw your comments after merging. Currently we have no easy way to validate if user has permission to create index without trying to create index first. So I think it's ok for now as we have no other option. We can add this to AD public doc to let user know this. 
One workaround is to delete the custom result index after validating, but it's possible that another user running another detector on that index, so delete the result index is not safe.

ylwu-amzn: Another option is we can skip checking create permission for custom result index if we have concern about creating too many indices

PR #281, opened 2021-10-31
Support custom result indices in multi-category filtering API

### LINKED ISSUE #280: Support custom result indices searching in multi-category filtering API
The newly-added multi-category filtering API (#261) doesn't include support for searching over custom result indices, which is also being added as a new feature. Add support by adjusting the query used to fetch results, to search over additional custom result indices if necessary.

See [this comment](https://github.com/opensearch-project/anomaly-detection/pull/261/files/760657a3b72312dc3ab849804c255a559ffcb8aa#r737877644) for more details.

ohltyler:

### Description
Adds support for searching over detectors that have a specified custom result index (see #276). When retrieving detector info, checks for any set custom result index, and adds it to the list of indices to search over. 

Confirmed it works for custom & non-custom-result-index detectors, confirmed pagination works
 
### Issues Resolved
closes #280 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/SearchTopAnomalyResultTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/SearchTopAnomalyResultTransportAction.java
@@ -27,7 +27,6 @@ import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.util.Strings;
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
@@ -40,6 +39,7 @@ import org.opensearch.ad.model.AnomalyResult;
 import org.opensearch.ad.model.AnomalyResultBucket;
 import org.opensearch.ad.transport.handler.ADSearchHandler;
 import org.opensearch.client.Client;
+import org.opensearch.common.Strings;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.ExistsQueryBuilder;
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/ad/transport/SearchTopAnomalyResultTransportAction.java
```diff
@@ -308,6 +308,12 @@ protected void doExecute(Task task, SearchTopAnomalyResultRequest request, Actio
             // Generating the search request which will contain the generated query
             SearchRequest searchRequest = generateSearchRequest(request);
 
+            // Adding search over any custom result indices
+            String customResultIndex = Strings.trimToNull(getAdResponse.getDetector().getResultIndex());
+            if (customResultIndex != null) {
+                searchRequest.indices(defaultIndex, customResultIndex);
```
ohltyler, line 314: TODO: if an index pattern is set here, this may fail. May need to search and iterate over existing indices that match the detector, and add all of them into a concrete list of custom result indices to search over. See SearchAnomalyResultTransportAction in #276.

ohltyler: ^ this can probably be ignored. Since we don't need to add a result index param in the API (which could contain a wildcard / be an index pattern), then we can strictly retrieve the single concrete result index from the detector.

ylwu-amzn (approved):

ylwu-amzn:
BTW, do we need to change anything on frontend?

ohltyler:
> BTW, do we need to change anything on frontend?

Nothing needed on frontend. Verified w/ frontend that everything works as expected.

src/main/java/org/opensearch/ad/transport/SearchTopAnomalyResultTransportAction.java
amitgalitz: What is the reason for still needing to search all AD Result indices if we know the detector has a custom result index?

ohltyler: it's following the same strategy as custom result index implementation, which searches over both, as a performance tradeoff so that we don't miss any results, in queries where we want to fetch results from both places (e.g., detector list fetches results from detectors that have results in both system index & custom result indices). 

If in the future we let user change the destination result index, we would want to support both/multiple

ohltyler: ^ also, its a very minor performance hit here, since the result scope is narrowed to only one detector + results w/ anomaly grade > 0

amitgalitz: Got it thank you

amitgalitz (approved):

PR #285, opened 2021-11-03
Skipping checking create index permission for Validate API 

### LINKED ISSUE #284: Integrate new custom result index with Validation API
The newly added validation API isn't integrated with new user ability to add a custom result index during AD creation. This will likely involve wrapping exception from currently validation code for custom result index as ADValidationException and making sure they are displayed well as a callout during AD Creation process on the frontend.

amitgalitz:

### Description

 * Changed AnomalyDetector constructor exceptions types from `IllegalArgumentException` to `ADValidationException` for any issues with custom result index. ADValidationExceptions are set in a format that is easy to convert to a validationResponse. This happens when the validate API is called, ADValidationExceptions get converted to ADValidation Responses. This will mean any of these issues will be nicely displayed to the user when validate API runs on the `ReviewAndCreate` page on the frontend. The issue will be displayed as a callout. Examples of this can be seen in this [PR](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/106).
 * Made sure to skip checking for create index permission when validate API is called. The reason for this is because during the call to `initCustomResultIndexAndExecute` a custom result index will be created. However we don't want to already create the index when user is just validating as it can lead to indices being created that never get used.

Will add a few tests.  

### Issues Resolved
#284 
 
### Check List


kaituo:
Can you give me contexts of the PR?

First, why ADValidationException gives nicer display than IllegalArgumentException?
Second, why do we want to skip checking for create index permission when validate API is called? Where is the related change? From the 6 lines you added, nothing obvious to me.

amitgalitz:
> Can you give me contexts of the PR?
> 
> First, why ADValidationException gives nicer display than IllegalArgumentException? Second, why do we want to skip checking for create index permission when validate API is called? Where is the related change? From the 6 lines you added, nothing obvious to me.

I added more information to the PR description, let me know if this addresses your question or if you have any follow up questions based on new description.

kaituo:
For others who read the PRs,

1. about the exception conversion, related code is: https://github.com/opensearch-project/anomaly-detection/blob/44baa818f8cdc7bc44e98781c3716b98a83b2986/src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java#L129 where ADvalidationExceptions get caught in ValidateAnomalyDetectorTransportAction to get converted to validationResponse
2. dry run check is added in AbstractAnomalyDetectorActionHandler.start method to avoid going to the code path of initCustomResultIndexAndExecute

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -227,7 +227,7 @@ public AbstractAnomalyDetectorActionHandler(
     public void start() {
         String resultIndex = anomalyDetector.getResultIndex();
         // use default detector result index which is system index
-        if (resultIndex == null) {
+        if (resultIndex == null || this.isDryRun) {
```
kaituo, line 230: If using custom index, we won't validate result index security permission in validateCustomResultIndexAndExecute, right? Is that by design? If so, why?

ylwu-amzn: This is initially to avoid creating custom result index when user call validation API. This change also skip the permission check for an existing custom result index. I'm ok to this as we still have validation when user call create detector API. But like Kaituo said, it will be better if we can call `validateCustomResultIndexAndExecute` to validate permission for an existing custom result index.

amitgalitz: Yes I can call `validateCustomResultIndexAndExecute` if `resultIndex` is not null and `isDryRun` is true. What are your thoughts on wrapping these exception and returning them as ADValidationResponse or to simply return as exception/error message that will show as bottom of page notification. Might need to have a longer discussion offline for best way going forward with these exceptions in general.

kaituo: I like your idea of returning them as ADValidationResponse.  We can extract the original exception's error message and return ADValidationResponse instead.

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -231,11 +231,36 @@ public void start() {
             createOrUpdateDetector();
             return;
         }
+        // if only validate API called then check if any errors occur for custom result index
+        // and wrap them as ADValidationException so they are properly returned as ADValidationResponse
+        // also make sure not to create actual index when validating if it doesn't exist
+        if (this.isDryRun) {
```
ylwu-amzn, line 237: Only check permission when result index exists ? `If (dryRun && anomalyDetectionIndices.doesIndexExist(resultIndex))` ?

amitgalitz: At that line I would know that resultIndex exists cause the if check above it checks if its null and if its null it proceeds right away to `createOrUpdateDetector`

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -227,15 +227,40 @@ public AbstractAnomalyDetectorActionHandler(
...
+                                )
+                            );
+                        return;
+                    })
+                );
+            return;
+        }
 
-        // user custom result index
+        // user custom result index if not validate
```
kaituo, line 260: you meant to say: use custom result index if not validating/dry run?

amitgalitz: Yes, will fix this, thank you!

ylwu-amzn (approved):

kaituo (approved):

PR #292, opened 2021-11-09
support only searching results in custom result index

ylwu-amzn:

### Description
In current implementation, we always query AD results from default index. Actually detector result page doesn't need to query default index if detector is using custom result index. And for API users, they may need to query AD results from custom result index only as well.

This PR supports searching AD result from custom result index only if user specify "?only_query_custom_result_index=true" when call search result API.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
@@ -73,24 +73,39 @@ public class SearchAnomalyResultTransportAction extends HandledTransportAction<S
     @Override
     protected void doExecute(Task task, SearchRequest request, ActionListener<SearchResponse> listener) {
         String[] indices = request.indices();
+        if (indices == null || indices.length == 0) {
+            listener.onFailure(new IllegalArgumentException("No indices set in search request"));
+            return;
+        }
         // Set query indices as default result indices, will check custom result indices permission and add
         // custom indices which user has search permission later.
-        request.indices(ALL_AD_RESULTS_INDEX_PATTERN);
+
+        boolean onlyQueryCustomResultIndex = true;
+        for (String indexName : indices) {
+            // If only query custom result index, don't need to set ALL_AD_RESULTS_INDEX_PATTERN in search request
+            if (ALL_AD_RESULTS_INDEX_PATTERN.equals(indexName)) {
+                onlyQueryCustomResultIndex = false;
+            }
+        }
+
+        String[] concreteIndices = indexNameExpressionResolver
+            .concreteIndexNames(clusterService.state(), IndicesOptions.lenientExpandOpen(), indices);
+        if (concreteIndices == null || concreteIndices.length == 0) {
+            // No custom result indices found, just search default result index
+            listener.onFailure(new IllegalArgumentException("No indices found"));
+            return;
+        }
 
         Set<String> customResultIndices = new HashSet<>();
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -73,26 +73,41 @@ public SearchAnomalyResultTransportAction(
...
-            for (String index : concreteIndices) {
-                if (index.startsWith(CUSTOM_RESULT_INDEX_PREFIX)) {
-                    customResultIndices.add(index);
-                }
+        }
+
+        String[] concreteIndices = indexNameExpressionResolver
+            .concreteIndexNames(clusterService.state(), IndicesOptions.lenientExpandOpen(), indices);
+        if (concreteIndices == null || concreteIndices.length == 0) {
+            // No custom result indices found, just search default result index
```
ohltyler, line 94: Comment here seems confusing. Since `concreteIndices` can include the default result system index, won't this `if` block only be true if no indices in the entire cluster match the search request's `indices`?

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -105,6 +120,7 @@ protected void doExecute(Task task, SearchRequest request, ActionListener<Search
             try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
                 // Search result indices of all detectors. User may create index with same prefix of custom result index
                 // which not used for AD, so we should avoid searching extra indices which not used by anomaly detectors.
+                boolean finalOnlyQueryCustomResultIndex = onlyQueryCustomResultIndex;
```
ohltyler, line 123: can you explain why this is needed?

ylwu-amzn: sure, will add some comments
"Variable used in lambda expression should be final or effectively final"

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
ylwu-amzn: Good catch, will tune the comment, should be 
"// No result indices found, will throw exception"

src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -161,7 +183,9 @@ protected void doExecute(Task task, SearchRequest request, ActionListener<Search
                 listener.onFailure(e);
             }
         } else {
+            // onlyQueryCustomResultIndex is false in this branch
             // Search only default result index
+            request.indices(ALL_AD_RESULTS_INDEX_PATTERN);
```
amitgalitz, line 188: Since we are now skipping the permission checks in the `if` case will searchHandler handle search each index and seeing if user has permission to read it?

ylwu-amzn: Default result index is system index, so only super admin has permission to read it. So we don't need to do permission check, if user have search AD result permission, they should be able to query default AD result index.

ohltyler (approved):
LGTM!

amitgalitz (approved):

Issue #290, opened 2021-11-08
[BUG] Historical HC detector state showing as STOPPED during first run


ohltyler:
When starting a historical HC job for the first time, it seems to finish but reports the state as STOPPED instead of FINISHED. 

I've only experienced this happening under the following scenarios:
1. Detector job is somewhat long and running over lots of data (>1 week of dense HC data w/ 10 min. interval)
2. Is the first historical job ran for the detector, by selecting a historical date range in the create flow and clicking 'Create'

Any subsequent runs seem to report the same results but have a FINISHED state. Also, no stack traces / errors seen from backend.

ylwu-amzn:
This is a known issue for HC detector. When multiple entities send back task done message to coordinator node, coordinator node will update detector level task state&progress, if one entity is updating detector level task, all other entity task will be ignored.

PR #300, opened 2021-11-10
wait for some time to get semaphore when set HC detector task as done

ylwu-amzn:

### Description
https://github.com/opensearch-project/anomaly-detection/issues/290
> When starting a historical HC job for the first time, it seems to finish but reports the state as STOPPED instead of FINISHED.

This is a known issue for HC detector. When multiple entities send back task done message to coordinator node, coordinator node will update detector level task state&progress, if one entity is updating detector level task, all other entity task will be ignored. For example, HC detector has 3 entities, all of them finished almost the same time, so first entity can update detector level task as RUNNING, second and third entities task done messages will be ignored. So detector level task will remain RUNNING state. Then get detector API will reset task state as STOPPED.

This PR add wait time when get updating task state semaphore when HC detector historical analysis done. So we can make sure the last entity task done message can get semaphore and update HC detector task state as FINISHED correctly.

This PR also fix a bug of null `isHCDetector` of `AnomalyResultResponse` which result in HC detector stays on `Initializing` state forever. Check more details in issue https://github.com/opensearch-project/anomaly-detection/issues/301

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/290
https://github.com/opensearch-project/anomaly-detection/issues/301
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/task/ADTaskManager.java
+++ b/src/main/java/org/opensearch/ad/task/ADTaskManager.java
@@ -2316,52 +2316,73 @@ public class ADTaskManager {
             adTaskCacheManager.removeHistoricalTaskCache(detectorId);
         });
 
+        long timeoutInMillis = 2000;// wait for 2 seconds to acquire updating HC detector task semaphore
         if (state == ADTaskState.FINISHED) {
             this.countEntityTasksByState(detectorTaskId, ImmutableList.of(ADTaskState.FINISHED), ActionListener.wrap(r -> {
                 logger.info("number of finished entity tasks: {}, for detector {}", r, adTask.getDetectorId());
                 // Set task as FAILED if no finished entity task; otherwise set as FINISHED
                 ADTaskState hcDetectorTaskState = r == 0 ? ADTaskState.FAILED : ADTaskState.FINISHED;
-                updateADHCDetectorTask(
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -2316,52 +2316,73 @@ public void setHCDetectorTaskDone(ADTask adTask, ADTaskState state, ActionListen
             adTaskCacheManager.removeHistoricalTaskCache(detectorId);
         });
 
+        long timeoutInMillis = 2000;// wait for 2 seconds to acquire updating HC detector task semaphore
```
ohltyler, line 2319: From my understanding this is related to how long it takes for all of the entities to finish. Is it sufficient for a large number of entities, or do you expect this to be enough time even in a worst-case scenario?

ohltyler: Discussed offline, understood now this is a limit of how long it takes the last entity to finish updating the task, rather than a limit of the number of tasks. During testing, the avg time was < 40ms, so seems 2s is a sufficient timeout.

ohltyler (approved):

kaituo (approved):

PR #322, opened 2021-11-22
don't throw exception if no result indices found when search both default and custom result index

ylwu-amzn:

### Description
Detector list page will search both default and custom result indices. For a new cluster, if no default/custom result created yet, the detector list page will not show any detector as the backend will throw "No indices found" exception.

This PR removed that exception, just search all default result indices and will return empty SearchResponse to frontend. 

 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
@@ -90,19 +90,21 @@ public class SearchAnomalyResultTransportAction extends HandledTransportAction<S
 
         String[] concreteIndices = indexNameExpressionResolver
             .concreteIndexNames(clusterService.state(), IndicesOptions.lenientExpandOpen(), indices);
-        if (concreteIndices == null || concreteIndices.length == 0) {
-            // No result indices found, will throw exception
-            listener.onFailure(new IllegalArgumentException("No indices found"));
-            return;
-        }
+        // If concreteIndices is null or empty, don't throw exception. Detector list page will search both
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -90,19 +90,21 @@ protected void doExecute(Task task, SearchRequest request, ActionListener<Search
...
         String[] concreteIndices = indexNameExpressionResolver
             .concreteIndexNames(clusterService.state(), IndicesOptions.lenientExpandOpen(), indices);
-        if (concreteIndices == null || concreteIndices.length == 0) {
-            // No result indices found, will throw exception
-            listener.onFailure(new IllegalArgumentException("No indices found"));
-            return;
-        }
+        // If concreteIndices is null or empty, don't throw exception. Detector list page will search both
+        // default and custom result indices to get anomaly of last 24 hours. If throw exception, detector
+        // list page will throw error and won't show any detector.
```
ohltyler, line 95: Can you add a sentence about why this case is ok for the scenario of a cluster with no custom result indices + some new non-custom-result-detector that hasn't finished one interval (where no default result index exists)?

ylwu-amzn: Sure

kaituo (approved):

ohltyler:
@ylwu-amzn here's the issue you can link in the comments instead: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/154

ohltyler (approved):
Thanks for the quick fix!

kaituo (approved):

PR #325, opened 2021-11-23
Add tests for search top anomaly results API

ohltyler:
### Description
Add unit and integ tests for search top anomaly results API. Meets the jacoco coverage bar for all classes added as part of the API, removed from the exclusion list in `build.gradle`.
 

### Check List


amitgalitz:
Are you planning on adding any integration tests with security

src/test/java/org/opensearch/ad/model/AnomalyResultBucketTests.java
```diff
@@ -0,0 +1,25 @@
...
+package org.opensearch.ad.model;
+
+import java.io.IOException;
+
+import org.opensearch.ad.AbstractADTest;
+import org.opensearch.ad.TestHelpers;
+import org.opensearch.common.io.stream.BytesStreamOutput;
+import org.opensearch.common.io.stream.StreamInput;
+
+public class AnomalyResultBucketTests extends AbstractADTest {
```
amitgalitz, line 15: Could you add a test for the toXContent() method?

ohltyler:
> Are you planning on adding any integration tests with security

Not as part of this PR. Seems there is a lot of missing API tests for security so seems it would be beneficial to do all of these together as a separate effort.

src/test/java/org/opensearch/ad/TestHelpers.java
```diff
@@ -392,6 +387,61 @@ public static AnomalyDetector randomAnomalyDetectorUsingCategoryFields(String de
...
+            null,
+            randomInt(),
+            Instant.now(),
+            categoryFields,
+            randomUser(),
+            null
+        );
+    }
+
+    public static AnomalyDetector randomAnomalyDetectorUsingCategoryFields(
```
amitgalitz, line 417: nit: multiple `randomAnomalyDetectorUsingCategoryFields` is a little confusing. Maybe can add w/ result index to this one or with custom indices to the one just above since that differentiates the methods. Also nit picky but these could be refactored into only calling the full `new AnomalyDetector(...)` once by gradually passing more variables like we do in `randomAnomalyDetector` if we choose to keep same name.

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -1504,4 +1508,393 @@ public void testValidateAnomalyDetectorWithWrongCategoryField() throws Exception
...
+            true,
+            client()
+        );
+        Exception noCategoryFieldsException = expectThrows(
+            IOException.class,
+            () -> {
+                searchTopAnomalyResults(
+                    detectorWithNoCategoryFields.getDetectorId(),
+                    false,
+                    "{\"start_time_ms\":1, \"end_time_ms\":2}",
```
amitgalitz, line 1648: nit: maybe we should pass category field here just to make it clear the issue is that the detector itself has no category field and not that search request is missing it.

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -1504,4 +1508,393 @@ public void testValidateAnomalyDetectorWithWrongCategoryField() throws Exception
         );
 
     }
+
+    public void testSearchTopAnomalyResultsWithInvalidInputs() throws IOException {
```
amitgalitz, line 1512: This is a lot of edge cases! Impressive!

src/test/java/org/opensearch/ad/TestHelpers.java
```diff
@@ -1001,6 +1084,44 @@ public static void createIndex(RestClient client, String indexName, HttpEntity d
...
+        for (Map.Entry<String, String> entry : categoryFieldsAndTypes.entrySet()) {
+            indexMappings.append("\"" + entry.getKey() + "\":{\"type\":\"" + entry.getValue() + "\"},");
+        }
+        indexMappings.append("\"timestamp\":{\"type\":\"date\"}");
+        indexMappings.append("}}");
+        createEmptyIndex(client, indexName);
+        createIndexMapping(client, indexName, TestHelpers.toHttpEntity(indexMappings.toString()));
+    }
+
+    public static void createAnomalyResultIndex(RestClient client) throws IOException {
```
amitgalitz, line 1100: nit: name change to `createEmptyAnomalyResultIndex` or `createAnomalyResultIndexWithNoData`

src/test/java/org/opensearch/ad/transport/SearchTopAnomalyResultActionTests.java
```diff
@@ -0,0 +1,265 @@
...
+            false,
+            1,
+            Arrays.asList(categoryFields.get(0)),
+            "invalid-order",
+            startTime,
+            endTime
+        );
+        expectThrows(
+            IllegalArgumentException.class,
+            () -> client().execute(SearchTopAnomalyResultAction.INSTANCE, searchRequest).actionGet(10_000)
```
amitgalitz, line 102: smart to put timeout

src/test/java/org/opensearch/ad/transport/SearchTopAnomalyResultActionTests.java
```diff
@@ -0,0 +1,265 @@
...
+import org.opensearch.ad.HistoricalAnalysisIntegTestCase;
+import org.opensearch.ad.TestHelpers;
+import org.opensearch.ad.model.AnomalyDetector;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.test.OpenSearchIntegTestCase;
+
+import com.google.common.collect.ImmutableList;
+
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST)
+public class SearchTopAnomalyResultActionTests extends HistoricalAnalysisIntegTestCase {
```
amitgalitz, line 26: Add a happy case also maybe?

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
ohltyler: The category field doesn't have to be passed in the request - the API will default to using all category fields found for the detector. I didn't want to pass a dummy category field value in the request since it would guarantee failure due to invalid category field, rather than missing category field from the detector itself. I tried to clarify the test case on line 1630.

src/test/java/org/opensearch/ad/transport/SearchTopAnomalyResultResponseTests.java
```diff
@@ -0,0 +1,40 @@
...
+        );
+
+        BytesStreamOutput output = new BytesStreamOutput();
+        originalResponse.writeTo(output);
+        StreamInput input = output.bytes().streamInput();
+        SearchTopAnomalyResultResponse parsedResponse = new SearchTopAnomalyResultResponse(input);
+        assertEquals(originalResponse.getAnomalyResultBuckets(), parsedResponse.getAnomalyResultBuckets());
+    }
+
+    public void testEmptyResults() {
```
amitgalitz, line 31: nit: I assume this is fine as is since there would be an exception or compiler error if you couldn't initialize an empty response but maybe add an assertEquals with expected empty response. Same for `testPopulatedResults` below.

src/test/java/org/opensearch/ad/TestHelpers.java
ohltyler: Good catch, will change to cascading constructors.

src/test/java/org/opensearch/ad/TestHelpers.java
ohltyler: Sure, changed to first suggestion

src/test/java/org/opensearch/ad/model/AnomalyResultBucketTests.java
ohltyler: Sure, will add.

src/test/java/org/opensearch/ad/transport/SearchTopAnomalyResultActionTests.java
ohltyler: Can't add happy cases due to issues with the `lang-painless` module not being installed on test clusters brought up with `OpenSearchIntegTest`s, so I cover those at the rest-level. I actually had a comment about that in there originally, I will add it back to prevent confusion.

ohltyler: See https://github.com/elastic/elasticsearch/issues/41109

src/test/java/org/opensearch/ad/transport/SearchTopAnomalyResultResponseTests.java
ohltyler: Prefer to not add an assertEquals here since it's a bit redundant. The response constructor that takes in a list (used here and in `testPopulatedResults`) simply assigns it to the single `anomalyResultBuckets` field, so not much to check.

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
amitgalitz: This makes sense to me! Thank you!

src/test/java/org/opensearch/ad/transport/SearchTopAnomalyResultActionTests.java
amitgalitz: Okay I see, thanks for linking the issue. Comment could be nice to add as I just instinictaly thought of needing a happy case as I forgot about you mentioning this before.

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -1504,4 +1508,393 @@ public void testValidateAnomalyDetectorWithWrongCategoryField() throws Exception
...
+                ),
+            true,
+            client()
+        );
+
+        // Clear any existing result index, create an empty one
+        if (indexExists(CommonName.ANOMALY_RESULT_INDEX_ALIAS)) {
+            deleteIndex(CommonName.ANOMALY_RESULT_INDEX_ALIAS);
+        }
+        TestHelpers.createEmptyAnomalyResultIndex(client());
```
ylwu-amzn, line 1726: Seems this is not the normal process from user's perspective. We don't allow user to create default AD result index directly. How about change to run historical analysis to create AD result index, then search top anomaly result?

ohltyler: The purpose of this test was about having the index exist, but not return any results, so I just create the empty index so the API can query it, but still return empty results. Because this API doesn't interact with anything other than reading the results index, I still think this is ok. Same strategy is used in the below tests, as well as the existing search results tests.

I still prefer this than to run a historical analysis, so as to keep the test more lightweight, faster, and not increase the chances of it becoming flaky. Let me know if you still have concerns.

ylwu-amzn: Got it. Sounds good. 
Another way to generate an empty default result index is run historical analysis on a short period (so it can run quickly), then delete all results with delete result REST API. But like you said, these steps will make the test heavier and more flaky likely.

amitgalitz (approved):

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -1504,4 +1508,393 @@ public void testValidateAnomalyDetectorWithWrongCategoryField() throws Exception
         );
 
     }
+
+    public void testSearchTopAnomalyResultsWithInvalidInputs() throws IOException {
+        String indexName = randomAlphaOfLength(10).toLowerCase();
```
ylwu-amzn, line 1513: Index name must not contain some special chars like "#", "|", and must not start with "-", "+" or "_".
Not sure if the `randomAlphaOfLength` will always return valid index name. How about we use some concrete index name to avoid flaky test?

ylwu-amzn: And same concern for detector name as Amit added regex checking detector name. @amitgalitz can you help check if we have such potential flaky part?

ohltyler: Good catch. But, followed the stack of calls, and seems `randomAlphaOfLength` will only return `a-zA-Z`, since it's coming from this generator:
```
public class AsciiLettersGenerator extends CodepointSetGenerator {
  private final static char [] CHARS = 
      ("abcdefghijklmnopqrstuvwxyz" + 
       "ABCDEFGHIJKLMNOPQRSTUVWXYZ").toCharArray();

  public AsciiLettersGenerator() {
    super(CHARS);
  }
}
```

amitgalitz: +1 on what Tyler mentioned, doesn't create anything flaky.

ylwu-amzn: Cool, thanks for confirming that

ylwu-amzn (approved):

PR #334, opened 2021-12-03
Bump plugin to 1.3.0.0

ohltyler:

### Description
Bumps plugin version to 1.3.0.0.
 
### Check List


amitgalitz (dismissed):
Thanks for the change!

build.gradle
```diff
@@ -301,7 +301,7 @@ String bwcFilePath = "src/test/resources/org/opensearch/ad/bwc/"
     testClusters {
         "${baseName}$i" {
             testDistribution = "ARCHIVE"
-            versions = ["7.10.2","1.2.0-SNAPSHOT"]
+            versions = ["7.10.2","1.3.0-SNAPSHOT"]
```
saratvemulapalli, line 304: We should also test compatibility between minor versions.
Like 1.2.0 -> 1.3.0. 
cc: @VachaShah

saratvemulapalli (changes requested):
We should test the latest minor of last major version (7.10.2) and last minor of current major version (1.2) for backwards compatibility.

build.gradle
VachaShah: Definitely agree that we should now test with multiple upgrade paths. One of the ways it is done in the OpenSearch repo is there is a list of bwcVersions that are iterated over and their upgrade paths are tested with respect to the current version.

ohltyler: @VachaShah if I'm understanding, you are saying that we should support BWC for the following scenarios, right?

7.10 -> 1.3
1.0 -> 1.3
1.1 -> 1.3
1.2 -> 1.3

I briefly looked at OpenSearch's `build.gradle`, seems it's quite different. Do you know of any plugin that's done this already? I guess it would mostly involve adding more loops to iterate over the older versions. Not sure if there's a cleaner solution.

Overall, agreed that more BWC testing could be supported. Rather than blocking this PR how about I open a tracking issue?

saratvemulapalli: @ohltyler absolutely agreed. As long as we take care of this before 1.3 releases we should be good.
Sooner the better :)

saratvemulapalli (approved):

build.gradle
VachaShah: @ohltyler Yeah thats correct about the scenarios. I will see if any plugin has done this already.

VachaShah (approved):

ylwu-amzn:
> We should test the latest minor of last major version (7.10.2) and last minor of current major version (1.2) for backwards compatibility.

@saratvemulapalli , Do you know what's the suggested upgrade policy for OpenSearch? Just upgrade to next newer version, rather than the latest version? For example, if user running cluster on 1.0, do we suggest them to upgrade to 1.3 directly? If yes, I think we should run BWC test for  `1.0 -> 1.3`.

saratvemulapalli:
> > We should test the latest minor of last major version (7.10.2) and last minor of current major version (1.2) for backwards compatibility.
> 
> @saratvemulapalli , Do you know what's the suggested upgrade policy for OpenSearch? Just upgrade to next newer version, rather than the latest version? For example, if user running cluster on 1.0, do we suggest them to upgrade to 1.3 directly? If yes, I think we should run BWC test for `1.0 -> 1.3`.

Correct. @VachaShah and @ohltyler are correct, I missed a couple of them. We should verify:
```
7.10 -> 1.3
1.0 -> 1.3
1.1 -> 1.3
1.2 -> 1.3
```
We do support upgrading to latest minor directly. (e.g 1.1 -> 1.3 is supported).

ylwu-amzn (approved):

ohltyler:
Created #336 to track support of multiple upgrade paths.

PR #346, opened 2021-12-31
Incremented version to 1.2.4.

dblock:
Coming from https://github.com/opensearch-project/opensearch-build/issues/1417.


dblock:
```
[ant:jacocoReport] Rule violated for class org.opensearch.ad.task.ADTaskManager: lines covered ratio is 0.74, but expected minimum is 0.75
```

ylwu-amzn (approved):
LGTM. Thanks for the change!

amitgalitz (approved):
Thank you for the change!

PR #348, opened 2021-12-31
Do not git ignore job scheduler dir, used.

dblock:
Signed-off-by: dblock <dblock@dblock.org>

### Description

When updating to 1.2.4 a .zip file was not git add-ed, had to do it manually.
 
### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ out/
 .classpath
 .vscode
 bin/
-src/test/resources/job-scheduler


ylwu-amzn (dismissed):
LGTM. The CI workflow failure fixed in PR https://github.com/opensearch-project/anomaly-detection/pull/350
You can rebase your branch on `main` to pass CI workflow. But as this PR just change `.gitignore` file, I think it's safe to skip the CI workflow. So it's up to you.

amitgalitz:
Are there any downsides to having it not be a part of .gitignore and potentially committing it when developing with an older job-scheduler build. I suppose that wont happen often but just wondering if you had any other thoughts on this as it was explicitly added to .gitignore in the past

ylwu-amzn (approved):

amitgalitz (approved):

dblock:
> Are there any downsides to having it not be a part of .gitignore and potentially committing it when developing with an older job-scheduler build. I suppose that wont happen often but just wondering if you had any other thoughts on this as it was explicitly added to .gitignore in the past

AFAIK bwc CI won't work without these

The fact that any dependency JARs are checked into the project are a problem. I opened https://github.com/opensearch-project/anomaly-detection/issues/352.

PR #354, opened 2022-01-11
upgrade gson to 2.8.9

ylwu-amzn:

### Description
Upgrade Gson to 2.8.9
 
 
### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -112,7 +112,7 @@ configurations.all {
         force "org.objenesis:objenesis:3.0.1"
         force "net.bytebuddy:byte-buddy:1.9.15"
         force "net.bytebuddy:byte-buddy-agent:1.9.15"
-        force "com.google.code.gson:gson:2.8.6"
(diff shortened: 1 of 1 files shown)


amitgalitz (approved):
LGTM if CI passes, thanks for making this change. Read over https://github.com/google/gson/blob/master/CHANGELOG.md and no breaking change seems to be introduced.  Are we backporting this to 1.2 then?

ylwu-amzn:
> LGTM if CI passes, thanks for making this change. Read over https://github.com/google/gson/blob/master/CHANGELOG.md and no breaking change seems to be introduced. Are we backporting this to 1.2 then?

We can backport to all old versions. The vulnerability is from `writeReplace` method which not used in AD. So should be ok if we don't backfill. 
We met some issue before as AD and other components used different versions of Gson. We should test on every version before backporting.

ohltyler (approved):

PR #384, opened 2022-02-28
Adding Model Type Validation to Validate API ("non-blocker")

amitgalitz:
### Description
This PR adds two new changes.

1.  Adds model (“non-blocker”) type validation to the current validation API. The model type validation is non-blocking in nature meaning that any warning that arise are simply suggestions or added information to the user about which configuration could be changed in order to increase the likelihood of model training completing successfully. These validation checks could also potentially recommend a specific detector interval or window delay value that the user can try to use.
    * Every-time model type validation is called, it initially also calls the blocker validation since there its not possible to run non-blocker checks if there is a blocking issue initially. 
2. Adds a check that the given timefield exists in the index mapping and also that it is of type date to AbstractAnomalyDetectorActionHandler which means this check is now executed for model

### Non-Blocker validation steps:

1. Get latest time stamp (if no latest time stamp found then user is told there is not enough historical data)
2. Check if detector is of type multi-entity and if it is then only look at the top entity for any proceeding validation
3. Check if general density of data is good enough for current configurations to complete model training with all the configurations applied.
    * This check will be run multiple times with different interval configurations until an interval is either recommended, same as the given one or no interval can be found
4. If no interval can be found with all configurations applied that will lead to data that is dense enough then proceed to run general density check by sequentially adding configurations one at a time
    * For example, first only the raw data is looked at in terms of how dense the data is →  if raw data on its own is dense enough then next the data filter (filter_query) will be added and general density will be checked for again
5. If no other issues exist in the data then check if any new window delay recommendation is possible (latest time stamp is used)

### Additional Info (Important):

* All validation for general density involves a bucket aggregation that finds if enough buckets in the last x intervals have at least 1 document present.
* I will be adding more new tests in separate commit or PR
* Response wording haven't finished tech writer review yet
### API Request:
* Path: `_plugins/_anomaly_detection/detectors/_validate/<aspect>`
* Parameter(aspect)
    * `model`
* Payload: Detector Configs
 
#### Example 1 (data ingested only every 5 mins, recommends 5 minutes interval of 1 minute) :
 ```
{
    "name": "test-hc-detector",
    "description": "Test detector",
    "time_field": "@timestamp",
    "indices": [
        "host-cloudwatch-two"
    ],
    "feature_attributes": [
        {
            "feature_name": "test",
            "feature_enabled": true,
            "aggregation_query": {
                "test": {
                    "avg": {
                        "field": "cpu"
                    }
                }
            }
        }
    ],
    "filter_query": {
        "bool": {
            "filter": [
                {
                    "exists": {
                        "field": "host"
                    }
                }
            ],
            "adjust_pure_negative": true,
            "boost": 1
        }
    },
    "detection_interval": {
        "period": {
            "interval": 1,
            "unit": "Minutes"
        }
    },
    "window_delay": {
        "period": {
            "interval": 1,
            "unit": "Minutes"
        }
    },
    "category_field": [
        "host"
    ]
}
```

#### Response 1:

```
{
    "model": {
        "detection_interval": {
            "message": "We suggest using a detector interval of: 5",
            "suggested_value": {
                "period": {
                    "interval": 5,
                    "unit": "Minutes"
                }
            }
        }
    }
}
```
#### Example 2 request (filter query leads to not enough dense data)
* CPU values in this data set range between 5-15 minutes
```
{
    "name": "test-hc-detector",
  ...
   "filter_query": {
        "bool": {
            "filter": [
                {
                    "range": {
                        "cpu": {
                            "gt": 10000
                        }
                    }
                }
            ],
            "adjust_pure_negative": true,
            "boost": 1
        }
    }
    "detection_interval": {
        "period": {
            "interval": 5,
            "unit": "Minutes"
        }
    },
    ...
}
```
#### Response 2: 
```
{
    "model": {
        "filter_query": {
            "message": "Data is too sparse after data filter is applied."
        }
    }
}
```

### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
+++ b/src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
@@ -13,11 +13,13 @@ package org.opensearch.ad.common.exception;
 
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.DetectorValidationIssueType;
+import org.opensearch.ad.model.IntervalTimeConfiguration;
 import org.opensearch.ad.model.ValidationAspect;
 
 public class ADValidationException extends AnomalyDetectionException {
     private final DetectorValidationIssueType type;
     private final ValidationAspect aspect;
+    private final IntervalTimeConfiguration intervalSuggestion;
 
     public DetectorValidationIssueType getType() {
         return type;
@@ -27,14 +29,34 @@ public class ADValidationException extends AnomalyDetectionException {
         return aspect;
     }
 
-    public ADValidationException(String message, DetectorValidationIssueType type, ValidationAspect aspect) {
-        this(message, null, type, aspect);
+    public IntervalTimeConfiguration getIntervalSuggestion() {
+        return intervalSuggestion;
     }
 
-    public ADValidationException(String message, Throwable cause, DetectorValidationIssueType type, ValidationAspect aspect) {
+    public ADValidationException(String message, DetectorValidationIssueType type, ValidationAspect aspect) {
+        this(message, null, type, aspect, null);
+    }
+
+    public ADValidationException(
+        String message,
+        DetectorValidationIssueType type,
+        ValidationAspect aspect,
+        IntervalTimeConfiguration intervalSuggestion
+    ) {
+        this(message, null, type, aspect, intervalSuggestion);
+    }
+
+    public ADValidationException(
+        String message,
+        Throwable cause,
+        DetectorValidationIssueType type,
+        ValidationAspect aspect,
+        IntervalTimeConfiguration intervalSuggestion
+    ) {
         super(AnomalyDetector.NO_ID, message, cause);
         this.type = type;
         this.aspect = aspect;
+        this.intervalSuggestion = intervalSuggestion;
     }
 
     @Override
@@ -49,6 +71,12 @@ public class ADValidationException extends AnomalyDetectionException {
         if (aspect != null) {
             sb.append(" aspect: ");
             sb.append(aspect.getName());
+        }
+
+        if (intervalSuggestion != null) {
+            sb.append("interval Suggestion: ");
+            sb.append(intervalSuggestion.getInterval());
+            sb.append(intervalSuggestion.getUnit());
         }
 
         return sb.toString();
--- a/src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
+++ b/src/main/java/org/opensearch/ad/common/exception/ADValidationException.java
@@ -74,7 +74,7 @@ public class ADValidationException extends AnomalyDetectionException {
         }
 
         if (intervalSuggestion != null) {
-            sb.append("interval Suggestion: ");
+            sb.append(" interval suggestion: ");
             sb.append(intervalSuggestion.getInterval());
             sb.append(intervalSuggestion.getUnit());
         }
--- a/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
@@ -115,7 +115,8 @@ public class CommonErrorMessages {
     public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied. Consider changing the data filter";
     public static String CATEGORY_FIELD_TOO_SPARSE = "Data is most likely too sparse with the given category fields.";
     public static String CATEGORY_FIELD_NO_DATA = "No entity was found with the given categorical fields.";
-    public static String FEATURE_QUERY_TOO_SPARSE = "Given data is most likely too sparse when given feature query is applied: ";
+    public static String FEATURE_QUERY_TOO_SPARSE =
+        "Given data is most likely too sparse when given feature queries are applied. Consider revising feature queries.";
 
     // Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction
     public static final String FEATURE_INVALID_MSG_PREFIX = "Feature has an invalid query";
--- a/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
@@ -107,12 +107,13 @@ public class CommonErrorMessages {
         + " characters.";
 
     public static String WINDOW_DELAY_REC = "We suggest using a window delay value of at least: ";
-    public static String NOT_ENOUGH_HISTORICAL_DATA = "There isn't enough historical data found with current configurations";
-    public static String DETECTOR_INTERVAL_REC = "We suggest using a detector interval of : ";
+    public static String NOT_ENOUGH_HISTORICAL_DATA = "There isn't enough historical data found with current configurations.";
+    public static String DETECTOR_INTERVAL_REC = "We suggest using a detector interval of: ";
     public static String RAW_DATA_TOO_SPARSE = "Given index data is potentially too sparse for model training.";
     public static String MODEL_VALIDATION_FAILED_UNEXPECTEDLY = "Model validation experienced issues completing.";
     public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied.";
-    public static String CATEGORY_FIELD_TOO_SPARSE = "Data is most likely too sparse with the given category fields";
+    public static String CATEGORY_FIELD_TOO_SPARSE = "Data is most likely too sparse with the given category fields.";
+    public static String CATEGORY_FIELD_NO_DATA = "No entity was found with the given categorical fields.";
     public static String FEATURE_QUERY_TOO_SPARSE = "Given data is most likely too sparse when given feature query is applied: ";
 
     // Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction
--- a/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
@@ -111,9 +111,9 @@ public class CommonErrorMessages {
     public static String TIME_FIELD_NOT_ENOUGH_HISTORICAL_DATA =
         "There isn't enough historical data found with current timefield selected.";
     public static String DETECTOR_INTERVAL_REC =
-        "The selected detector interval might collect sparse data. Consider increasing interval range too: ";
+        "The selected detector interval might collect sparse data. Consider changing interval length too: ";
     public static String RAW_DATA_TOO_SPARSE =
-        "Given index data is potentially too sparse for model training. Consider changing interval length or ingesting more data";
+        "Source index data is potentially too sparse for model training. Consider changing interval length or ingesting more data";
     public static String MODEL_VALIDATION_FAILED_UNEXPECTEDLY = "Model validation experienced issues completing.";
     public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied. Consider changing the data filter";
     public static String CATEGORY_FIELD_TOO_SPARSE =
@@ -121,7 +121,7 @@ public class CommonErrorMessages {
     public static String CATEGORY_FIELD_NO_DATA =
         "No entity was found with the given categorical fields. Consider revising category field/s or ingesting more data";
     public static String FEATURE_QUERY_TOO_SPARSE =
-        "Given data is most likely too sparse when given feature queries are applied. Consider revising feature queries.";
+        "Data is most likely too sparse when given feature queries are applied. Consider revising feature queries.";
     public static String TIMEOUT_ON_INTERVAL_REC = "Timed out getting interval recommendation";
 
     // Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction
--- a/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
@@ -80,7 +80,7 @@ public class CommonErrorMessages {
         "Valid characters for detector name are a-z, A-Z, 0-9, -(hyphen), _(underscore) and .(period)";
     public static String DUPLICATE_FEATURE_AGGREGATION_NAMES = "Detector has duplicate feature aggregation query names: ";
     public static String INVALID_TIMESTAMP = "Timestamp field: (%s) must be of type date";
-    public static String NON_EXISTENT_TIMESTAMP = "Timestamp field: (%s) is not found in index mapping:";
+    public static String NON_EXISTENT_TIMESTAMP = "Timestamp field: (%s) is not found in index mapping";
 
     public static String FAIL_TO_GET_DETECTOR = "Fail to get detector";
     public static String FAIL_TO_GET_DETECTOR_INFO = "Fail to get detector info";
@@ -106,17 +106,23 @@ public class CommonErrorMessages {
         + MAX_DETECTOR_NAME_SIZE
         + " characters.";
 
-    public static String WINDOW_DELAY_REC = "We suggest using a window delay value of at least: ";
+    public static String WINDOW_DELAY_REC =
+        "Latest seen data point is at least %d minutes ago, consider changing window delay to at least %d minutes.";
     public static String TIME_FIELD_NOT_ENOUGH_HISTORICAL_DATA =
         "There isn't enough historical data found with current timefield selected.";
-    public static String DETECTOR_INTERVAL_REC = "Suggested detector interval: ";
-    public static String RAW_DATA_TOO_SPARSE = "Given index data is potentially too sparse for model training.";
+    public static String DETECTOR_INTERVAL_REC =
+        "The selected detector interval might collect sparse data. Consider increasing interval range too: ";
+    public static String RAW_DATA_TOO_SPARSE =
+        "Given index data is potentially too sparse for model training. Consider changing interval length or ingesting more data";
     public static String MODEL_VALIDATION_FAILED_UNEXPECTEDLY = "Model validation experienced issues completing.";
     public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied. Consider changing the data filter";
-    public static String CATEGORY_FIELD_TOO_SPARSE = "Data is most likely too sparse with the given category fields.";
-    public static String CATEGORY_FIELD_NO_DATA = "No entity was found with the given categorical fields.";
+    public static String CATEGORY_FIELD_TOO_SPARSE =
+        "Data is most likely too sparse with the given category fields. Consider revising category field/s or ingesting more data ";
+    public static String CATEGORY_FIELD_NO_DATA =
+        "No entity was found with the given categorical fields. Consider revising category field/s or ingesting more data";
     public static String FEATURE_QUERY_TOO_SPARSE =
         "Given data is most likely too sparse when given feature queries are applied. Consider revising feature queries.";
+    public static String TIMEOUT_ON_INTERVAL_REC = "Timed out getting interval recommendation";
 
     // Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction
     public static final String FEATURE_INVALID_MSG_PREFIX = "Feature has an invalid query";
--- a/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
@@ -79,8 +79,8 @@ public class CommonErrorMessages {
     public static String INVALID_DETECTOR_NAME =
         "Valid characters for detector name are a-z, A-Z, 0-9, -(hyphen), _(underscore) and .(period)";
     public static String DUPLICATE_FEATURE_AGGREGATION_NAMES = "Detector has duplicate feature aggregation query names: ";
-    public static String INVALID_TIMESTAMP = "Timestamp must be of type date";
-    public static String NON_EXISTENT_TIMESTAMP = "Timestamp not found in index mapping";
+    public static String INVALID_TIMESTAMP = "Timestamp field: (%s) must be of type date";
+    public static String NON_EXISTENT_TIMESTAMP = "Timestamp field: (%s) is not found in index mapping:";
 
     public static String FAIL_TO_GET_DETECTOR = "Fail to get detector";
     public static String FAIL_TO_GET_DETECTOR_INFO = "Fail to get detector info";
@@ -107,11 +107,12 @@ public class CommonErrorMessages {
         + " characters.";
 
     public static String WINDOW_DELAY_REC = "We suggest using a window delay value of at least: ";
-    public static String NOT_ENOUGH_HISTORICAL_DATA = "There isn't enough historical data found with current configurations.";
-    public static String DETECTOR_INTERVAL_REC = "We suggest using a detector interval of: ";
+    public static String TIME_FIELD_NOT_ENOUGH_HISTORICAL_DATA =
+        "There isn't enough historical data found with current timefield selected.";
+    public static String DETECTOR_INTERVAL_REC = "Suggested detector interval: ";
     public static String RAW_DATA_TOO_SPARSE = "Given index data is potentially too sparse for model training.";
     public static String MODEL_VALIDATION_FAILED_UNEXPECTEDLY = "Model validation experienced issues completing.";
-    public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied.";
+    public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied. Consider changing the data filter";
     public static String CATEGORY_FIELD_TOO_SPARSE = "Data is most likely too sparse with the given category fields.";
     public static String CATEGORY_FIELD_NO_DATA = "No entity was found with the given categorical fields.";
     public static String FEATURE_QUERY_TOO_SPARSE = "Given data is most likely too sparse when given feature query is applied: ";
--- a/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
@@ -79,6 +79,8 @@ public class CommonErrorMessages {
     public static String INVALID_DETECTOR_NAME =
         "Valid characters for detector name are a-z, A-Z, 0-9, -(hyphen), _(underscore) and .(period)";
     public static String DUPLICATE_FEATURE_AGGREGATION_NAMES = "Detector has duplicate feature aggregation query names: ";
+    public static String INVALID_TIMESTAMP = "Timestamp must be of type date";
+    public static String NON_EXISTENT_TIMESTAMP = "Timestamp not found in index mapping";
 
     public static String FAIL_TO_GET_DETECTOR = "Fail to get detector";
     public static String FAIL_TO_GET_DETECTOR_INFO = "Fail to get detector info";
@@ -103,4 +105,22 @@ public class CommonErrorMessages {
     public static String INVALID_DETECTOR_NAME_SIZE = "Name should be shortened. The maximum limit is "
         + MAX_DETECTOR_NAME_SIZE
         + " characters.";
+
+    public static String WINDOW_DELAY_REC = "We suggest using a window delay value of at least: ";
+    public static String NOT_ENOUGH_HISTORICAL_DATA = "There isn't enough historical data found with current configurations";
+    public static String DETECTOR_INTERVAL_REC = "We suggest using a detector interval of : ";
+    public static String RAW_DATA_TOO_SPARSE = "Given index data is potentially too sparse for model training.";
+    public static String MODEL_VALIDATION_FAILED_UNEXPECTEDLY = "Model validation experienced issues completing.";
+    public static String FILTER_QUERY_TOO_SPARSE = "Data is too sparse after data filter is applied.";
+    public static String CATEGORY_FIELD_TOO_SPARSE = "Data is most likely too sparse with the given category fields";
+    public static String FEATURE_QUERY_TOO_SPARSE = "Given data is most likely too sparse when given feature query is applied: ";
+
+    // Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction
+    public static final String FEATURE_INVALID_MSG_PREFIX = "Feature has an invalid query";
+    public static final String FEATURE_WITH_EMPTY_DATA_MSG = FEATURE_INVALID_MSG_PREFIX + " returning empty aggregated data: ";
+    public static final String FEATURE_WITH_INVALID_QUERY_MSG = FEATURE_INVALID_MSG_PREFIX + " causing a runtime exception: ";
+    public static final String UNKNOWN_SEARCH_QUERY_EXCEPTION_MSG =
+        "Feature has an unknown exception caught while executing the feature query: ";
+    public static final String VALIDATION_FEATURE_FAILURE = "Validation failed for feature(s) of detector %s";
+
 }
--- a/src/main/java/org/opensearch/ad/constant/CommonName.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonName.java
@@ -89,6 +89,7 @@ public class CommonName {
     public static final String TYPE = "type";
     public static final String KEYWORD_TYPE = "keyword";
     public static final String IP_TYPE = "ip";
+    public static final String DATE = "date";
 
     // used for updating mapping
     public static final String SCHEMA_VERSION_FIELD = "schema_version";
@@ -134,7 +135,8 @@ public class CommonName {
     // Validation
     // ======================================
     // detector validation aspect
-    public static final String DETECTOR = "detector";
+    public static final String DETECTOR_ASPECT = "detector";
+    public static final String MODEL_ASPECT = "model";
 
     // ======================================
     // Used for custom AD result index
--- a/src/main/java/org/opensearch/ad/constant/CommonName.java
+++ b/src/main/java/org/opensearch/ad/constant/CommonName.java
@@ -89,7 +89,7 @@ public class CommonName {
     public static final String TYPE = "type";
     public static final String KEYWORD_TYPE = "keyword";
     public static final String IP_TYPE = "ip";
-    public static final String DATE = "date";
+    public static final String DATE_TYPE = "date";
 
     // used for updating mapping
     public static final String SCHEMA_VERSION_FIELD = "schema_version";
--- a/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
@@ -496,24 +496,6 @@ public class SearchFeatureDao extends AbstractRetriever {
             );
     }
 
-    /**
-     * Get the entity's earliest timestamps
-     * @param detector detector config
-     * @param listener listener to return back the requested timestamps
-     */
-    public void getMinDataTime(AnomalyDetector detector, ActionListener<Optional<Long>> listener) {
-        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
-            .aggregation(AggregationBuilders.min(AGG_NAME_MIN).field(detector.getTimeField()))
-            .trackTotalHits(false)
-            .size(0);
-        SearchRequest searchRequest = new SearchRequest().indices(detector.getIndices().toArray(new String[0])).source(searchSourceBuilder);
-        client
-            .search(
-                searchRequest,
-                ActionListener.wrap(response -> { listener.onResponse(parseMinDataTime(response)); }, listener::onFailure)
-            );
-    }
-
     private Optional<Long> parseMinDataTime(SearchResponse searchResponse) {
         Optional<Map<String, Aggregation>> mapOptional = Optional
             .ofNullable(searchResponse)
--- a/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
@@ -496,6 +496,24 @@ public class SearchFeatureDao extends AbstractRetriever {
             );
     }
 
+    /**
+     * Get the entity's earliest timestamps
+     * @param detector detector config
+     * @param listener listener to return back the requested timestamps
+     */
+    public void getMinDataTime(AnomalyDetector detector, ActionListener<Optional<Long>> listener) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
+            .aggregation(AggregationBuilders.min(AGG_NAME_MIN).field(detector.getTimeField()))
+            .trackTotalHits(false)
+            .size(0);
+        SearchRequest searchRequest = new SearchRequest().indices(detector.getIndices().toArray(new String[0])).source(searchSourceBuilder);
+        client
+            .search(
+                searchRequest,
+                ActionListener.wrap(response -> { listener.onResponse(parseMinDataTime(response)); }, listener::onFailure)
+            );
+    }
+
     private Optional<Long> parseMinDataTime(SearchResponse searchResponse) {
         Optional<Map<String, Aggregation>> mapOptional = Optional
             .ofNullable(searchResponse)
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -94,7 +94,8 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     public static final String USER_FIELD = "user";
     public static final String DETECTOR_TYPE_FIELD = "detector_type";
     public static final String RESULT_INDEX_FIELD = "result_index";
-    public static final String MODEL_VALIDATION_ISSUE = "aggregation_issue";
+    public static final String AGGREGATION = "aggregation_issue";
+    public static final String TIMEOUT = "timeout";
     @Deprecated
     public static final String DETECTION_DATE_RANGE_FIELD = "detection_date_range";
 
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -95,6 +95,8 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     public static final String USER_FIELD = "user";
     public static final String DETECTOR_TYPE_FIELD = "detector_type";
     public static final String RESULT_INDEX_FIELD = "result_index";
+    public static final String GENERAL_DATA = "general_data";
+    public static final String MODEL_VALIDATION_ISSUE = "aggregation_issue";
     @Deprecated
     public static final String DETECTION_DATE_RANGE_FIELD = "detection_date_range";
 
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -77,7 +77,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     public static final String TYPE = "_doc";
     public static final String QUERY_PARAM_PERIOD_START = "period_start";
     public static final String QUERY_PARAM_PERIOD_END = "period_end";
-    public static final String PARSING_ISSUE = "query_parsing";
     public static final String GENERAL_SETTINGS = "general_settings";
 
     public static final String NAME_FIELD = "name";
@@ -95,7 +94,6 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
     public static final String USER_FIELD = "user";
     public static final String DETECTOR_TYPE_FIELD = "detector_type";
     public static final String RESULT_INDEX_FIELD = "result_index";
-    public static final String GENERAL_DATA = "general_data";
     public static final String MODEL_VALIDATION_ISSUE = "aggregation_issue";
     @Deprecated
     public static final String DETECTION_DATE_RANGE_FIELD = "detection_date_range";
--- a/src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
+++ b/src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
@@ -24,7 +24,10 @@ public enum DetectorValidationIssueType implements Name {
     FILTER_QUERY(AnomalyDetector.FILTER_QUERY_FIELD),
     WINDOW_DELAY(AnomalyDetector.WINDOW_DELAY_FIELD),
     GENERAL_SETTINGS(AnomalyDetector.GENERAL_SETTINGS),
-    RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD);
+    RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD),
+    GENERAL_DATA(AnomalyDetector.GENERAL_DATA),
+    MODEL_VALIDATION_ISSUE(AnomalyDetector.MODEL_VALIDATION_ISSUE); // TODO: this is a unique case where aggregation failed but not
+                                                                    // exception
 
     private String name;
 
--- a/src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
+++ b/src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
@@ -25,8 +25,9 @@ public enum DetectorValidationIssueType implements Name {
     WINDOW_DELAY(AnomalyDetector.WINDOW_DELAY_FIELD),
     GENERAL_SETTINGS(AnomalyDetector.GENERAL_SETTINGS),
     RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD),
-    MODEL_VALIDATION_ISSUE(AnomalyDetector.MODEL_VALIDATION_ISSUE); // this is a unique case where aggregation failed but
-                                                                    // don't want to throw exception
+    TIMEOUT(AnomalyDetector.TIMEOUT),
+    AGGREGATION(AnomalyDetector.AGGREGATION); // this is a unique case where aggregation failed but
+                                              // don't want to throw exception
 
     private String name;
 
--- a/src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
+++ b/src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
@@ -26,7 +26,7 @@ public enum DetectorValidationIssueType implements Name {
     GENERAL_SETTINGS(AnomalyDetector.GENERAL_SETTINGS),
     RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD),
     TIMEOUT(AnomalyDetector.TIMEOUT),
-    AGGREGATION(AnomalyDetector.AGGREGATION); // this is a unique case where aggregation failed but
+    AGGREGATION(AnomalyDetector.AGGREGATION); // this is a unique case where aggregation failed due to an issue in core but
                                               // don't want to throw exception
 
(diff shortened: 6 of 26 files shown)

### CODE REVIEW

**Comment by ohltyler** [comment]
Have the responses been reviewed by the tech writer? The 2 examples you have seem a little confusing to me:
> "message": "We suggest using a detector interval of: 5",

Who is "we"? Can this be written as more of a direct suggestion, like "Suggested interval: 5 minutes" or something?

> "message": "Data is too sparse after data filter is applied."

This seems unclear on next steps for the user. Can there be some added text about removing or revising the data filter?

File: src/main/java/org/opensearch/ad/common/exception/ADValidationException.java [hunk 1]
```diff
@@ -51,6 +73,12 @@ public String toString() {
             sb.append(aspect.getName());
         }
 
+        if (intervalSuggestion != null) {
+            sb.append("interval Suggestion: ");
```
ohltyler, line 77: nit: if matching the above, can change to `" interval suggestion: "`

src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
```diff
@@ -79,6 +79,8 @@ public static String getTooManyCategoricalFieldErr(int limit) {
     public static String INVALID_DETECTOR_NAME =
         "Valid characters for detector name are a-z, A-Z, 0-9, -(hyphen), _(underscore) and .(period)";
     public static String DUPLICATE_FEATURE_AGGREGATION_NAMES = "Detector has duplicate feature aggregation query names: ";
+    public static String INVALID_TIMESTAMP = "Timestamp must be of type date";
+    public static String NON_EXISTENT_TIMESTAMP = "Timestamp not found in index mapping";
```
ohltyler, line 83: nit: maybe `"Timestamp field not found in index mapping"`?

amitgalitz:
> Have the responses been reviewed by the tech writer? The 2 examples you have seem a little confusing to me:
> 
> > "message": "We suggest using a detector interval of: 5",
> 
> Who is "we"? Can this be written as more of a direct suggestion, like "Suggested interval: 5 minutes" or something?
> 
> > "message": "Data is too sparse after data filter is applied."
> 
> This seems unclear on next steps for the user. Can there be some added text about removing or revising the data filter?

These responses have yet to be finalized by tech writers, will add that to top description. Currently planned on adding a little more text on the frontend such as "We suggest going back and changing the data filter" (this wording will need to be improved). Do you suggest I add this sort of direction to the API response itself

ohltyler:
> > Have the responses been reviewed by the tech writer? The 2 examples you have seem a little confusing to me:
> > > "message": "We suggest using a detector interval of: 5",
> > 
> > 
> > Who is "we"? Can this be written as more of a direct suggestion, like "Suggested interval: 5 minutes" or something?
> > > "message": "Data is too sparse after data filter is applied."
> > 
> > 
> > This seems unclear on next steps for the user. Can there be some added text about removing or revising the data filter?
> 
> These responses have yet to be finalized by tech writers, will add that to top description. Currently planned on adding a little more text on the frontend such as "We suggest going back and changing the data filter" (this wording will need to be improved). Do you suggest I add this sort of direction to the API response itself

I see, I'd discuss with UX on that. To me it seems that you could add all of the text as part of the API and just propagate it to the frontend for simplicity. Don't need to add UI-specific context like "go back to this page" but more like "change the data filter"

src/main/java/org/opensearch/ad/constant/CommonName.java
```diff
@@ -89,6 +89,7 @@
     public static final String TYPE = "type";
     public static final String KEYWORD_TYPE = "keyword";
     public static final String IP_TYPE = "ip";
+    public static final String DATE = "date";
```
ohltyler, line 92: nit: suggest to change to `DATE_TYPE` to match the other vars?

src/main/java/org/opensearch/ad/rest/AbstractAnomalyDetectorAction.java
```diff
@@ -39,6 +34,7 @@ public AbstractAnomalyDetectorAction(Settings settings, ClusterService clusterSe
         this.maxSingleEntityDetectors = MAX_SINGLE_ENTITY_ANOMALY_DETECTORS.get(settings);
         this.maxMultiEntityDetectors = MAX_MULTI_ENTITY_ANOMALY_DETECTORS.get(settings);
         this.maxAnomalyFeatures = MAX_ANOMALY_FEATURES.get(settings);
+        // this.trainSampleTimeRangeInHours = TRAIN_SAMPLE_TIME_RANGE_IN_HOURS.get(settings);
```
ohltyler, line 37: leftover line - should it be removed?

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -310,7 +322,71 @@ protected void validateDetectorName(boolean indexingDryRun) {
...
+                            }
+                        }
+                    }
+                }
+            }
+            if (!foundField) {
+                listener
+                    .onFailure(
+                        new ADValidationException(
+                            CommonErrorMessages.NON_EXISTENT_TIMESTAMP,
```
ohltyler, line 376: Can the exception include the detector's timestamp field name for more detail?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                        .getCategoryField()
+                        .stream()
+                        .map(f -> new TermsValuesSourceBuilder(f).field(f))
+                        .collect(Collectors.toList())
+                )
+                .size(1000)
+                .subAggregation(
+                    PipelineAggregatorBuilders
+                        .bucketSort("bucketSort", Collections.singletonList(new FieldSortBuilder("_count").order(SortOrder.DESC)))
+                        .size(1)
```
ohltyler, line 178: It's understandable to only get the top entity combination to use when doing the sparsity checks. But just thinking, what if the data has one or two dominant/frequent entity combos, and many more that are less common, but still useful? For example, maybe an interval of 5 mins is above the threshold for the top 1 entity combo, but an interval of 10 mins is above the threshold for the top 20 entity combos, in which case a suggestion of 10 mins interval would be much more useful from the customer perspective. Is it easy enough to repeat this for the top x entity combos and try to derive a more general working interval suggestion? (this applies to single-category HCAD as well, just using multi-category as the example).

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+            listener
+                .onFailure(
+                    new ADValidationException(
+                        CommonErrorMessages.NOT_ENOUGH_HISTORICAL_DATA,
+                        DetectorValidationIssueType.GENERAL_DATA,
+                        ValidationAspect.MODEL
+                    )
+                );
+            return;
+        }
```
ohltyler, line 257: Seems these 2 can be consolidated in one block. Also, when would the time be <= 0?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+            }
+            for (Map.Entry<String, Object> entry : topEntity.entrySet()) {
+                query.filter(QueryBuilders.termQuery(entry.getKey(), entry.getValue()));
+            }
+        }
+        if (anomalyDetector.isMultiCategoryDetector()) {
+            for (String category : anomalyDetector.getCategoryField()) {
+                query.filter(QueryBuilders.existsQuery(category));
+            }
+        }
```
ohltyler, line 294: What is the purpose of this filter? Wouldn't the keys in the previous if block encapsulate this already?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+        for (String featureField : featureFields) {
+            query.filter(QueryBuilders.existsQuery(featureField));
+        }
```
ohltyler, line 297: Same question for this filter

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+        Long detectorInterval = timeConfigToMilliSec(detectorIntervalInMinutes);
+        Long startMillis = endMillis - ((long) getNumberOfSamples() * detectorInterval);
+        return new LongBounds(startMillis, endMillis);
+    }
+
+    private int getNumberOfSamples() {
+        long interval = anomalyDetector.getDetectorIntervalInMilliseconds();
+        return Math
+            .max(
+                (int) (Duration.ofHours(AnomalyDetectorSettings.TRAIN_SAMPLE_TIME_RANGE_IN_HOURS).toMillis() / interval),
```
ohltyler, line 710: May need to handle divide by 0 error if the request is coming via API. Frontend prevents a value <=0 to be set here, but a direct request could cause a problem.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                double fullBucketRate = processBucketAggregationResults(aggregate);
+                long newInterval = (long) Math.ceil(convertIntervalToMinutes(detectorInterval) + 1);
+                // If rate is below success minimum then call search again.
+                if (fullBucketRate > INTERVAL_BUCKET_MINIMUM_SUCCESS_RATE) {
+                    intervalListener.onResponse(this.detectorInterval);
+                } else if (expirationEpochMs < clock.millis()) {
+                    listener
+                        .onFailure(
+                            new AnomalyDetectionException(
+                                "Timed out getting interval recommendation. Please continue with detector creation."
```
ohltyler, line 393: Second sentence assumes it is ok. Suggest to remove it or change to something like "Please try again"

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+        @Override
+        public void onResponse(SearchResponse response) {
+            try {
+                Histogram aggregate = checkBucketResultErrors(response);
+                if (aggregate == null) {
+                    return;
+                }
+
+                double fullBucketRate = processBucketAggregationResults(aggregate);
+                long newInterval = (long) Math.ceil(convertIntervalToMinutes(detectorInterval) + 1);
```
ohltyler, line 385: Can you add some detailed comments on the logic behind this calculation?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: Added this check on the backend as well: https://github.com/opensearch-project/anomaly-detection/blob/30d0636ccacfbfc5aa1844b66b35da2ba44f0d79/src/main/java/org/opensearch/ad/model/AnomalyDetector.java#L213

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                // If rate is below success minimum then call search again.
```
ohltyler, line 386: Misleading comment

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                }
+
+            } catch (Exception e) {
+                onFailure(e);
+            }
+        }
+
+        @Override
+        public void onFailure(Exception e) {
+            logger.error("Failed to paginate top anomaly results", e);
```
ohltyler, line 424: nit: copied over message that can be tuned

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+            listener.onFailure(e);
+        }
+    }
+
+    private void processIntervalRecommendation(IntervalTimeConfiguration interval, long latestTime) {
+        if (interval == null) {
+            checkRawDataSparsity(latestTime);
+        } else {
```
ohltyler, line 432: Can you add a clarifying comment about this block? From my understanding this means that no interval could be found with the filters applied (category field, feature fields, etc), and so now you are checking the raw data, to determine if the problem is from the raw data, or the filters

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+
+    private void checkRawDataSparsity(long latestTime) {
+        AggregationBuilder aggregation = getBucketAggregation(latestTime);
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder().aggregation(aggregation).size(0).timeout(requestTimeout);
+        try {
+            AggregatorFactories.Builder internalAgg = parseAggregators(
+                anomalyDetector.getFeatureAttributes().get(0).getAggregation().toString(),
+                xContentRegistry,
+                anomalyDetector.getFeatureAttributes().get(0).getId()
+            );
```
ohltyler, line 473: Shouldn't this include all of the features and not just the `0`th one?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                .onFailure(
+                    new ADValidationException(
+                        CommonErrorMessages.MODEL_VALIDATION_FAILED_UNEXPECTEDLY,
+                        DetectorValidationIssueType.MODEL_VALIDATION_ISSUE,
+                        ValidationAspect.MODEL
+                    )
+                );
+            return null;
+        }
+        Histogram aggregate = aggs.get("agg");
```
ohltyler, line 500: Suggest to throw all of the `"agg"` references into a constant

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                );
+        } else if (anomalyDetector.isMultientityDetector()) {
+            getTopEntityForCategoryField(latestTime);
+        } else {
+            try {
+                checkFeatureQuery(latestTime);
+            } catch (Exception ex) {
+                logger.error(ex);
+                listener.onFailure(ex);
+            }
```
ohltyler, line 559: Can you add a comment about these `else` blocks? From my understanding it means that the filter query is ok, so now you are checking the features / category fields, in case the sparsity problem likes there

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+            listener
+                .onFailure(
+                    new ADValidationException(
+                        CommonErrorMessages.CATEGORY_FIELD_TOO_SPARSE,
+                        DetectorValidationIssueType.CATEGORY,
+                        ValidationAspect.MODEL
+                    )
+                );
+            return;
+        }
```
ohltyler, line 584: Isn't this already caught on line 275?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+
+    private void checkFeatureQuery(long latestTime) throws IOException {
+        ActionListener<MergeableList<double[]>> validateFeatureQueriesListener = ActionListener
+            .wrap(response -> { windowDelayRecommendation(latestTime); }, exception -> {
+                listener
+                    .onFailure(
+                        new ADValidationException(
+                            exception.getMessage(),
+                            DetectorValidationIssueType.FEATURE_ATTRIBUTES,
+                            ValidationAspect.DETECTOR
```
ohltyler, line 628: Should be `ValidationAspect.MODEL` right?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+                    multiFeatureQueriesResponseListener
+                        .onResponse(new MergeableList<>(new ArrayList<>(Collections.singletonList(new double[] { fullBucketRate }))));
+                }
+            }, e -> {
+                String errorMessage;
+                if (isExceptionCausedByInvalidQuery(e)) {
+                    errorMessage = CommonErrorMessages.FEATURE_WITH_INVALID_QUERY_MSG + feature.getName();
+                } else {
+                    errorMessage = CommonErrorMessages.UNKNOWN_SEARCH_QUERY_EXCEPTION_MSG + feature.getName();
+                }
```
ohltyler, line 675: Wouldn't these checks already be caught in the blocker validation logic?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,718 @@
...
+            });
+        MultiResponsesDelegateActionListener<MergeableList<double[]>> multiFeatureQueriesResponseListener =
+            new MultiResponsesDelegateActionListener<>(
+                validateFeatureQueriesListener,
+                anomalyDetector.getFeatureAttributes().size(),
+                String.format(Locale.ROOT, CommonErrorMessages.VALIDATION_FEATURE_FAILURE, anomalyDetector.getName()),
+                false
+            );
+
+        for (Feature feature : anomalyDetector.getFeatureAttributes()) {
```
ohltyler, line 640: This only checks each individual feature for causing the sparsity. What about the case of some combination of features is what causes the sparsity? For example, say feature `A` by itself passes the minimum success rate, and feature `B` by itself passes the minimum success rate, but when combining the queries of feature `A` and feature `B`, drops it below the minimum success rate. In that case, the API would return the features as being ok when it's actually the cause of data sparsity.

Suggest to combine all of these features into one query to holistically determine if the combination of features is the culprit of sparsity or not

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: I left this from when I was just making sure category fields exists and not getting the top entity. Removing this.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: This filter makes sure that the fields given in the feature queries are present in the documents that I am looking for in the bucket aggregation

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: I could potentially remove this, right now we aren't targeting having a manual validation process for them to try again but more of just a message that they can continue. I might change this to a generic message that shows `"Model validation experienced issues completing."`

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: Consildating these into 1. There are cases when last timestamp is negative when date field is in index mapping but its not ever filled (these happens on when timefield is set to `customer_birth_date` for the opensearch dashboard sample data). It would return a last timestamp of `-9223372036854775808`

src/main/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportAction.java
```diff
@@ -186,17 +191,24 @@ protected DetectorValidationIssue parseADValidationException(ADValidationExcepti
...
+                if (exception.getAspect().equals(ValidationAspect.MODEL)) {
+                    intervalSuggestion = exception.getIntervalSuggestion();
+                }
             case FILTER_QUERY:
             case TIMEFIELD_FIELD:
             case SHINGLE_SIZE_FIELD:
             case WINDOW_DELAY:
+                if (exception.getAspect().equals(ValidationAspect.MODEL)) {
+                    intervalSuggestion = exception.getIntervalSuggestion();
+                }
```
ohltyler, line 203: Seems the added logic could be added before or after the switch statements, and doesn't have to do with the exception type. Also, this is duplicate logic with no breaks, so in the case of the type being `DETECTION_INTERVAL` this would get executed twice

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
```diff
@@ -119,6 +126,7 @@ public void setUp() throws Exception {
     }
 
     @SuppressWarnings("unchecked")
+    @Ignore
```
ohltyler, line 129: can remove this ignore when fixed I'm assuming?

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -304,9 +310,13 @@ public void testUpdateAnomalyDetector() throws Exception {
...
     public void testUpdateAnomalyDetectorNameToExisting() throws Exception {
-        AnomalyDetector detector1 = createRandomAnomalyDetector(true, true, client());
+        AnomalyDetector detector1 = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
+        String indexName1 = detector1.getIndices().get(0);
+        TestHelpers.createIndex(client(), indexName1, TestHelpers.toHttpEntity("{\"name\": \"test\"}"));
 
-        AnomalyDetector detector2 = createRandomAnomalyDetector(true, true, client());
+        AnomalyDetector detector2 = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
+        String indexName2 = detector2.getIndices().get(0);
+        TestHelpers.createIndex(client(), indexName2, TestHelpers.toHttpEntity("{\"name\": \"test\"}"));
```
ohltyler, line 319: Seems a lot of duplicate code in this test suite - suggest to create a helper method that creates a detector and ingests sample data to the generated source index

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
```diff
@@ -24,7 +24,10 @@
     FILTER_QUERY(AnomalyDetector.FILTER_QUERY_FIELD),
     WINDOW_DELAY(AnomalyDetector.WINDOW_DELAY_FIELD),
     GENERAL_SETTINGS(AnomalyDetector.GENERAL_SETTINGS),
-    RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD);
+    RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD),
+    GENERAL_DATA(AnomalyDetector.GENERAL_DATA),
```
ohltyler, line 28: Can you explain what this issue type means?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
ohltyler: Ah, I see this isn't in the HCAD block. Still seems unnecessary in the HCAD scenario, can it be added as an `else` block so it's only added in the non-HCAD scenario?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: This is a fair point, original idea behind doing each feature at a time was to find if one of the feature is causing the issue. But as you say that might lead to feature query being skipped as root issue. I'll change to running them all at once with the only downside being that the user won't know which feature contributed more to the lack of sparsity unless I do further analysis which I wont be implementing at this point. Overall it does make sense to find the feature queries combined being the issue rather than saying there is no issue.

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
amitgalitz: I'll actually change this since it occurs when `lastTimestamp` produces a negative value which means there isn't enough historical data with the timestamp used to get the time range. But this is just the effect and not the cause which is a timefield being chosen with no latest document found for it. Will change message to convey that and remove general_data issue type for now.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
ohltyler: Yeah, tradeoffs for either soln. Perfect soln would be trying to find a subset of features responsible for the sparsity, but even then, there could be multiple combinations, and could lead to a confusing response.

I don't have too strong opinion either way, but I slightly lean towards checking them altogether. That way it guarantees a valid response, even if it's more vague.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: I still think its necessary in the HCAD scenario as the fields given in the feature query don't necessarily have to relate to the category fields given. For example the feature query could be aggregating `memory` field but up category field is `ip`. I still want to only look and see if we have enough documents with the `memory` field present in them when aggregating.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: I could implement this within the first check when all configurations are applied. It would mean running multiple intervals for each one of the top 10 entities and keeping track of the highest interval. The best case is when I find an interval for all top entities but I guess if there is only an interval found for 5 of them, maybe it means the category fields should be relooked at or it could be good enough? On the check where I am adding the category field to the configurations I could repeat this as well but first goal could be only to initially do it for the first run with all configurations.

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
amitgalitz: Removed this field and changed the use of it to be referring to the timefield since the issue leads to no historical data but it cause no last timestamp can be found and this is after the check that there is at least 1 document in index

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: This shouldn't actually be even added to the raw data aggregation and I removed it.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: Removing timeout clock completely as it doesn't seem to ever get even close to it

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
ohltyler: Yeah it's impossible to know for sure what would be the ideal number of running entities. Roughly, my idea would be
1. get top x entities (if < y then just optimize for single entity)
2. find a baseline interval that works for single top entity (what you're already doing)
3. check if that works for top z entities - if not, increase it a few more times to see if it can go over the threshold for top z entities - if not, then stick with baseline found in step 2

where x = num top entities constant, y = num top entities threshold, z = num top entities working given the interval threshold

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: Changed to generic model validation exception for now

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
ohltyler: ^ Id consider this not a hard reqt for this release, just an idea.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,707 @@
...
+        }
+        windowDelayRecommendation(latestTime);
+    }
+
+    private void checkFeatureQuery(long latestTime) throws IOException {
+        List<String> featureFields = ParseUtils.getFeatureFieldNames(anomalyDetector, xContentRegistry);
+        AggregationBuilder aggregation = getBucketAggregation(latestTime);
+        BoolQueryBuilder query = QueryBuilders.boolQuery().filter(anomalyDetector.getFilterQuery());
+        for (String featureField : featureFields) {
+            query.filter(QueryBuilders.existsQuery(featureField));
```
kaituo, line 664: What if this is a runtime field that does not actually exist?

kaituo (commented):
Finished 3/19 files

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -789,4 +789,18 @@ private AnomalyDetectorSettings() {}
...
+
+    // ======================================
+    // Validate Detector API setting
+    // ======================================
+    public static final long TOP_VALIDATE_TIMEOUT_IN_MILLIS = 60_000;
+    public static final long MAX_INTERVAL_REC_LENGTH_IN_MINUTES = 120L;
+    public static final double INTERVAL_RECOMMENDATION_INCREASING_MULTIPLIER = 1.2;
+    public static final double INTERVAL_RECOMMENDATION_DECREASING_MULTIPLIER = 0.8;
+    public static final double INTERVAL_BUCKET_MINIMUM_SUCCESS_RATE = 0.75;
+    public static final double CONFIG_BUCKET_MINIMUM_SUCCESS_RATE = 0.25;
```
kaituo, line 801: Why do we have different sparsity requirement for interval and other configs?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -95,6 +94,7 @@
     public static final String USER_FIELD = "user";
     public static final String DETECTOR_TYPE_FIELD = "detector_type";
     public static final String RESULT_INDEX_FIELD = "result_index";
+    public static final String MODEL_VALIDATION_ISSUE = "aggregation_issue";
```
kaituo, line 97: Does model validation issue equal aggregation issue?

amitgalitz: At first it only really happened when aggregation failed for some reason (could be a change in core) but now I also use this for timeout exception. However based on a different comment you made, I will make a different exception for timeout issue and pass that specifically to frontend.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
amitgalitz: The interval recommendation is done with multiple different intervals while the A/B config testing is done only with the given interval. We previously discussed to lower this to 0.5 but I tried 0.25 and it seemed to make more sense as A/B testing was more about finding if config would significantly lower success rate. This is to make knowing which config cause the issue slightly more likely to be accurate but I can change this back to 0.5. Future implementation for 1.4 release could include changing implementation to check for multiple intervals for each configuration.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,707 @@
...
+                } else if (expirationEpochMs < clock.millis()) {
+                    listener
+                        .onFailure(
+                            new ADValidationException(
+                                CommonErrorMessages.MODEL_VALIDATION_FAILED_UNEXPECTEDLY,
+                                DetectorValidationIssueType.MODEL_VALIDATION_ISSUE,
+                                ValidationAspect.MODEL
+                            )
+                        );
+                    logger.info("Timed out getting interval recommendation");
```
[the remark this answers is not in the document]
amitgalitz: Yes I can add it to the exception, I preferred keeping it simple with ADValidationException type so I'll create this type of exception for a timeout issue and pass along the timeout message in the exception.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: I can add a check to see if `runtime_mapping` is a field in the feature query but since we still don't allow these types of queries I will prioritize other bugs and frontend first.

kaituo: sounds good

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: I don't quite follow. So you start with interval search and find none can give you >= 75%. Then you go to search other config and try to find any one config results in <= 25%. That config/raw data is the potential issue.  If all of the following checks return a value < 75% > 25%, you would give no validation error/recommendation. Is that correct? If yes, in this case I would prefer you tell me my raw data is too sparse instead of no error.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
kaituo: If it is only about aggregation, maybe change the variable name to be more specific?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
amitgalitz: Yes there is the case where everything else is above 25% but below the 75% with everything.  I'll change it to suggest raw data is too sparse

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,707 @@
...
+        private AggregationBuilder getNewAggregationBuilder(long newInterval) {
+            return AggregationBuilders
+                .dateHistogram(AGGREGATION)
+                .field(anomalyDetector.getTimeField())
+                .minDocCount(0)
+                .hardBounds(getTimeRangeBounds(latestTime, new IntervalTimeConfiguration(newInterval, ChronoUnit.MINUTES)))
+                .fixedInterval(DateHistogramInterval.minutes((int) newInterval));
+        }
+
+        private long convertIntervalToMinutes(IntervalTimeConfiguration interval) {
```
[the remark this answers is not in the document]
amitgalitz: Yes that is correct, I have a method that converts it to millisecond when I recommend window delay or get the time range bounds but that isn't relevant to this. I'll change this to `convertIntervalToSeconds` and move it over to be a method in `IntervalTimeConfiguration` class

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -793,8 +793,8 @@ private AnomalyDetectorSettings() {}
     // ======================================
     // Validate Detector API setting
     // ======================================
-    public static final long TOP_VALIDATE_TIMEOUT_IN_MILLIS = 60_000;
-    public static final long MAX_INTERVAL_REC_LENGTH_IN_MINUTES = 120L;
+    public static final long TOP_VALIDATE_TIMEOUT_IN_MILLIS = 10_000;
```
ohltyler, line 796: Why decrease so much? Most of the API-level timeouts are 60s. Given a cluster under heavy load with a lot of data, couldn't this end up timing out a lot? 

I'm not sure on the details of your performance testing, but given all of the potential queries you're making, I can imagine this can easily take 2-3s given just a normal amount of data.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -496,6 +496,24 @@ public void getEntityMinDataTime(AnomalyDetector detector, Entity entity, Action
             );
     }
 
+    /**
+     * Get the entity's earliest timestamps
+     * @param detector detector config
+     * @param listener listener to return back the requested timestamps
+     */
+    public void getMinDataTime(AnomalyDetector detector, ActionListener<Optional<Long>> listener) {
```
ohltyler, line 504: Shouldn't the minimum time be related to what the specified interval is? Why not just default to using a min time that's equal to the min time that would be used during cold start?

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
```diff
@@ -25,8 +25,9 @@
     WINDOW_DELAY(AnomalyDetector.WINDOW_DELAY_FIELD),
     GENERAL_SETTINGS(AnomalyDetector.GENERAL_SETTINGS),
     RESULT_INDEX(AnomalyDetector.RESULT_INDEX_FIELD),
-    MODEL_VALIDATION_ISSUE(AnomalyDetector.MODEL_VALIDATION_ISSUE); // this is a unique case where aggregation failed but
-                                                                    // don't want to throw exception
+    TIMEOUT(AnomalyDetector.TIMEOUT),
+    AGGREGATION(AnomalyDetector.AGGREGATION); // this is a unique case where aggregation failed but
```
ohltyler, line 29: Confused on what this issue type means exactly. What aggregation are you saying is failing here? Can you make the comment more clear? Why would we not want to throw an exception?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -384,29 +400,36 @@ public void onResponse(SearchResponse response) {
...
+                    searchWithDifferentInterval(newIntervalMinute);
                     // The below block is executed only the first time when new interval is above max and
                     // we aren't decreasing yet, at this point
-                } else if (newInterval >= MAX_INTERVAL_REC_LENGTH_IN_MINUTES && !decreasingInterval) {
+                } else if (newIntervalMinute >= MAX_INTERVAL_REC_LENGTH_IN_MINUTES && !decreasingInterval) {
                     IntervalTimeConfiguration givenInterval = (IntervalTimeConfiguration) anomalyDetector.getDetectionInterval();
                     this.detectorInterval = new IntervalTimeConfiguration(
-                        (long) Math.ceil(convertIntervalToMinutes(givenInterval) * INTERVAL_RECOMMENDATION_DECREASING_MULTIPLIER),
+                        (long) Math
+                            .floor(
```
ohltyler, line 418: If we're still increasing, shouldn't this still be `Math.ceil()`?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -252,7 +278,10 @@ private void getBucketAggregates(
         Map<String, Object> topEntity
     ) throws IOException {
         List<String> featureFields = ParseUtils.getFeatureFieldNames(anomalyDetector, xContentRegistry);
-        AggregationBuilder aggregation = getBucketAggregation(latestTime);
+        AggregationBuilder aggregation = getBucketAggregation(
+            latestTime,
+            (IntervalTimeConfiguration) anomalyDetector.getDetectionInterval()
```
ohltyler, line 283: Suggest to remove all these downcasts to `IntervalTimeConfiguration` and just change `AnomalyDetector`s interval field from `TimeConfiguration` -> `IntervalTimeConfiguration`

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -675,15 +710,24 @@ private void windowDelayRecommendation(long latestTime) {
...
                         DetectorValidationIssueType.WINDOW_DELAY,
                         ValidationAspect.MODEL,
                         new IntervalTimeConfiguration(minutesSinceLastStamp, ChronoUnit.MINUTES)
                     )
                 );
             return;
         }
-        listener.onResponse(null);
+        // This case has been reached if no interval or window delay recommendation was found and no configuration took down bucket
+        // rate bellow the config bucket rate minimum.
```
ohltyler, line 722: Don't you mean no config was found to take the bucket rate **above** the minimum?

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: In this block it will be the first time we start decreasing. I'll change comment wording to signify that I meant this block will be executed when we haven't decreased yet but we will start to decrease within execute of the code inside this if case

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
amitgalitz: This would happen when aggregations returns null indicating that there might be a bug or some opensearch core changes. We do this when aggregating for preview api and one or two others I believe. This comment explains why we just return empty response instead of exception in `SearchFeatureDao`(https://github.com/opensearch-project/anomaly-detection/blob/main/src%2Fmain%2Fjava%2Forg%2Fopensearch%2Fad%2Ffeature%2FSearchFeatureDao.java#L405-L407) and since an empty response means no issue in validation API instead I return that there was an aggregation issue. I will make this more clear in comments.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: Config validation has a lower minimum threshold (0.25) since we are looking to see if any configs drastically decreases bucket success rate below 25%. There are cases when with all configs applied that no interval rec was found to increase success rate above 75% but no single configurations when added reduced bucket rate below 25%. This means that the suggestion is too simply ingest more data if possible but we don't have anymore insight on the root cause. In the future this can be improved to also give interval recommendation that gave the best rate that was found anyways even if below .75 and we can also do more testing to change the .25 rate for configs checks and etc. I was planning on trying to make those changes only after this is merged. Probably only in time for 1.4 and not for 1.3

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
amitgalitz: I chose to use minimum data time to get the the top entity so it encapsulates all data that could be found with the different interval recommendation. To decrease amount of queries being made which are already quite high I didn't want to get the top entity based on each interval I try again and again. I could however change this to get the minimum time based on the longest possible interval that will be tried in the recommendation (which is currently 60 minutes). I think I will change to that if that makes sense to you

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
amitgalitz: Will do this in separate PR as it creates dozens of failing test that I don't want to prioritize right now along with making sure there was no specific reason why we had `TimeConfiguration` used and not `IntervalTimeConfiguration` in the first place (seems to be parsing related probably)

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
ohltyler: > I could however change this to get the minimum time based on the longest possible interval that will be tried in the recommendation (which is currently 60 minutes)

I think this is better than the current soln, yes. Even so, why not just default to using whatever the min timestamp would be as a lower bounds, instead of needing to make a separate query to see what the actual min is in the index?

src/main/java/org/opensearch/ad/model/DetectorValidationIssueType.java
ohltyler: Ok, got it. If this is strictly related to a problem with core instead of the detector config causing invalid aggregations in the queries then I'm fine with this.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
ohltyler: Sure, sounds good

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
ohltyler: Ok, that makes sense. Can you add more detail in the comment like you've mentioned here? I think mentioning the 2 different thresholds, and how neither was crossed, can help others understand when this scenario occurs.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -718,8 +707,11 @@ private void windowDelayRecommendation(long latestTime) {
                 );
             return;
         }
-        // This case has been reached if no interval or window delay recommendation was found and no configuration took down bucket
-        // rate bellow the config bucket rate minimum.
+        // This case has been reached if no interval recommendation was found that leads to a bucket success rate of >= 0.75
+        // but no single configuration during the following checks reduced the bucket success rate below 0.25
+        // This means the rate with all configs applied was below 0.75 but the rate when checking each configuration at time
+        // was always above 0.25 meaning the best suggestion is to simply ingest more data since we have no more insight
+        // regarding the root cause of the lower density.
```
ohltyler, line 714: Does this comment need to be tuned after recent changes to the thresholds?

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -304,9 +307,12 @@ public void testUpdateAnomalyDetector() throws Exception {
...
     public void testUpdateAnomalyDetectorNameToExisting() throws Exception {
-        AnomalyDetector detector1 = createRandomAnomalyDetector(true, true, client());
-
-        AnomalyDetector detector2 = createRandomAnomalyDetector(true, true, client());
+        AnomalyDetector detector1 = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
+        String indexName1 = detector1.getIndices().get(0);
+        TestHelpers.createIndex(client(), indexName1, TestHelpers.toHttpEntity("{\"name\": \"test\"}"));
+        AnomalyDetector detector2 = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
+        String indexName2 = detector2.getIndices().get(0);
+        TestHelpers.createIndex(client(), indexName2, TestHelpers.toHttpEntity("{\"name\": \"test\"}"));
```
ohltyler, line 315: can these use the `createIndexAndGetAnomalyDetector` helper method as well?

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -1328,9 +1331,9 @@ public void testValidateAnomalyDetectorWithIncorrectShingleSize() throws Excepti
     }
 
     public void testValidateAnomalyDetectorWithNoIssue() throws Exception {
-        AnomalyDetector detector = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null);
-        String indexName = detector.getIndices().get(0);
-        TestHelpers.createIndex(client(), indexName, TestHelpers.toHttpEntity("{\"timestamp\": " + Instant.now().toEpochMilli() + "}"));
+        String indexName = "testindex";
+        TestHelpers.createIndexWithTimeField(client(), "testindex", TIME_FIELD);
+        AnomalyDetector detector = TestHelpers.randomAnomalyDetector(TIME_FIELD, indexName);
```
ohltyler, line 1336: here as well

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -1441,17 +1446,16 @@ public void testValidateAnomalyDetectorWithFeatureQueryReturningNoData() throws
...
     }
 
     public void testValidateAnomalyDetectorWithFeatureQueryRuntimeException() throws Exception {
         String nonNumericField = "_type";
         Feature nonNumericFeature = TestHelpers.randomFeature("non-numeric-feature", nonNumericField, "avg", true);
-        AnomalyDetector detector = TestHelpers.randomAnomalyDetector(ImmutableList.of(nonNumericFeature));
-        String indexName = detector.getIndices().get(0);
-        TestHelpers.createIndex(client(), indexName, TestHelpers.toHttpEntity("{\"timestamp\": " + Instant.now().toEpochMilli() + "}"));
+        AnomalyDetector detector = TestHelpers.randomAnomalyDetector(TIME_FIELD, "index-test", ImmutableList.of(nonNumericFeature));
+        TestHelpers.createIndexWithTimeField(client(), "index-test", TIME_FIELD);
```
ohltyler, line 1458: I see several here that could use the helper method too, or at least a similar helper method

src/test/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportActionTests.java
```diff
@@ -435,4 +423,47 @@ public void testValidateAnomalyDetectorWithDetectorNameTooLong() throws IOExcept
         assertEquals(ValidationAspect.DETECTOR, response.getIssue().getAspect());
         assertTrue(response.getIssue().getMessage().contains("Name should be shortened. The maximum limit is"));
     }
+
+    @Test
+    public void testValidateAnomalyDetectorWithNonExistentTimefield() throws IOException {
```
ohltyler, line 428: Are you working on any tests to run the validation API with sparse data to test out all of the different scenarios? I think it's ok to have as a TODO, but maybe open an issue to track and add for following release.

amitgalitz: Yes I am currently continuing to do it in terms of integration tests like what I added here: `src/test/java/org/opensearch/ad/e2e/DetectionResultEvalutationIT.java`. Will cut an issue and work on adding more in separate PR.

ohltyler (approved):
LGTM! Just a few small things in recent comments.

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -692,26 +697,33 @@ private void checkFeatureQuery(long latestTime) throws IOException {
         client.search(searchRequest, ActionListener.wrap(response -> processFeatureQuery(response, latestTime), listener::onFailure));
     }
 
+    private void sendWindowDelayRec(long latestTime) {
```
ohltyler, line 700: nit: `latestTimeInMillis` may be more informative

src/main/java/org/opensearch/ad/rest/handler/ModelValidationActionHandler.java
```diff
@@ -0,0 +1,755 @@
...
+        for (String featureField : featureFields) {
+            query.filter(QueryBuilders.existsQuery(featureField));
+        }
+        SearchSourceBuilder searchSourceBuilder = getSearchSourceBuilder(query, aggregation);
+        SearchRequest searchRequest = new SearchRequest(anomalyDetector.getIndices().toArray(new String[0])).source(searchSourceBuilder);
+        client.search(searchRequest, ActionListener.wrap(response -> processFeatureQuery(response, latestTime), listener::onFailure));
+    }
+
+    private void sendWindowDelayRec(long latestTime) {
+        long minutesSinceLastStamp = TimeUnit.MILLISECONDS.toMinutes(Instant.now().toEpochMilli() - latestTime);
```
ohltyler, line 701: How does rounding work in this case? For example, if the difference is 1400 ms -> 1.4 minutes, will it return 1 or 2?. In this case, 1 minute isn't enough delay, and 2 minutes may be a better suggestion to guarantee incoming data.

ohltyler (approved):
LGTM!

kaituo (approved):

PR #395, opened 2022-03-02
Add JDK 11 to CI test matrix

### LINKED ISSUE #320: Default CI Java Version to Java 11, run tests on 8, 14 and 17
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/64, default CI Java version to 11 (LTS), and run tests on 8, 14 and 17.

- [ ] Change the default Java version for CI in GitHub actions from 14/15 to 11 (LTS).
- [ ] Add Java 8, 14 (used before), and 17 (next LTS) to the CI matrix.
- [ ] Document Java versions in DEVELOPER_GUIDE.md.

ohltyler:

### Description
One of the campaigns for upcoming 1.3 release is testing on JDK 11, with optional versions of 8, 14, 17. This updates the workflow to add JDK 11. There are other problems with 8/17 so ignoring those for now.

Also removes CD workflow since that hasn't been used since pre-OpenSearch days.
 
### Issues Resolved
Resolves #320 
 
### Check List


amitgalitz (approved):
LGTM, just make sure to wait until CI passes especially as this involves CI workflow changes. Also JDK8 might not be working because of protostuff dependency compiled with java 11, before that change was pushed it in, it was okay to use runtime jdk8. However the exception itself mentions something was compiled with java version 54 which is equivalent to java 10 so not 100% sure. Can be addressed in future PRs

.github/workflows/CI.yml
```diff
@@ -8,11 +8,10 @@ on:
       - "*"
 
 jobs:
-
   Build-ad:
     strategy:
       matrix:
-        java: [14]
+        java: [11, 14]
```
ylwu-amzn, line 14: All plugins and OpenSearch core should make sure Java 11 works for 1.3 release?

ylwu-amzn (approved):

.github/workflows/CI.yml
ohltyler: Yes

ohltyler: 11 is the default used by bundle build / integ tests

PR #406, opened 2022-03-09
Adding Locale.ROOT param on toLowerCase() calls

amitgalitz:

### Description
Currently we use .toLowerCase() in a lot of our tests to guarantee whatever string is being used for index name is lower case. Currently we only support index names that fit this regex: `a-z0-9_-` which doesn't include any non-english characters. However using .toLowerCase() can convert a capital i to a dotless i, which is used in Turkish and not part of the english characters we allow for index names ([javadocs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#toLowerCase())). 
 
### Issues Resolved
No specific issue but caused a flaky test failure on this run: https://github.com/opensearch-project/anomaly-detection/runs/5474162487?check_suite_focus=true
 
### Check List


ohltyler (approved):
Thanks for the quick fix!

ylwu-amzn (approved):

PR #438, opened 2022-03-15
Changed default description to empty string instead of null

### LINKED ISSUE #437: Handle missing description fields in create detector API
As pointed out in [this issue](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/171), the create API does not reject the missing `description` field, even though it's listed as `Required` in [the API](https://opensearch.org/docs/latest/monitoring-plugins/ad/api/#create-anomaly-detector). So, if a user submits an API request with all valid fields but a missing `description` field, then it leads to a missing `description` field in the system index, and parsing failures on the frontend and backend plugins, as it's expecting that field to exist (regardless of if it's value is populated or not). This creates confusion, since the raw response shows as successful when it actually is not. Details on the errors in the linked issue.

I see 2 possible paths forward:
1. Change to `Not required`, and handle the missing field in the backend similar to the other not required fields (e.g., `filter_query`/`window_delay`/`category_field`). Since it's listed as optional on the frontend, this aligns better from a user POV.
2. Leave as `Required`, and reject the request if it's missing in the body.

amitgalitz:

### Description
Changed the default description field to be set to an empty string instead of a null field. This means that description should also be changed from a `required` field as listed in [documentation](https://opensearch.org/docs/latest/monitoring-plugins/ad/api/#create-anomaly-detector) to an `optional` field since if detector description is missing then it will be filled as an empty string. 

Currently if there is a missing description field, the backend still creates the detector even though it is listed as `required` which causes issues on the frontend which tries to parse a null a value unsuccessfully when getting the relevant detector. 

Other option is to throw an exception if description is missing and hence enforce the `required` description which we don't enforce on the backend today and we also let users know it is optional on the frontend, where we just have it as empty if not filled out. For this reason, it makes more sense to just keep it consistent as an 'optional' field from UX point of view.

Reproduced the bug stated in the issue before change was made and tested that bug doesn't occur after this change 

### Issues Resolved
resolves #437 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -420,7 +420,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         TimeValue defaultDetectionWindowDelay
     ) throws IOException {
         String name = null;
-        String description = null;
+        String description = "";
         String timeField = null;
         List<String> indices = new ArrayList<String>();
         QueryBuilder filterQuery = QueryBuilders.matchAllQuery();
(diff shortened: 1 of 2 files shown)


ohltyler (approved):
@amitgalitz thanks for the quick work and owning this issue!

amitgalitz:
rebased with main @ohltyler @ylwu-amzn

ohltyler (approved):

ylwu-amzn (approved):

PR #447, opened 2022-03-16
Fixed ADTaskProfile toXContent bug and added to .gitignore

amitgalitz:

### Description
1. Fixed bug in `ADTaskProfile.toXContent()` which added `ENTITY_TASK_PROFILE_FIELD` instead of `LATEST_HC_TASK_RUN_TIME_FIELD` when `latestHCTaskRunTime` was present in constructor. Also added a test case for this. This previously would cause a parsing exception if a non null `latestHCTaskRunTime` was given to constructor.
2. Added `._.DS_Store` to `.gitignore` as its a file that gets auto created by macOS Finder which isn't relevant.
 
### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ out/
 .classpath
 .vscode
 bin/
+._.DS_Store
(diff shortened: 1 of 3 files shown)


ylwu-amzn (approved):
LGTM. Thanks for fixing this bug!

kaituo (approved):

Issue #451, opened 2022-03-21
Updating how we run integ tests


amitgalitz:
**Is your feature request related to a problem? Please describe.**
AD has experienced multiple flaky tests that are occurring at a rate that slows down development. AD’s Github workflows takes around 1 hour to complete, these are executed before merging and on each new commit that is made on an open PR. 

AD and other OpenSearch repos are relaying on Github’s runners that are quite small (2 x86_64 CPU cores and 7 GB RAM). This seems to be the main reason for many of AD’s flaky tests as there isn’t enough hardware processing power to consistently pass all our tests as we create indices, ingest data, start detectors, updates detectors and etc.

I propose that we consider a few possible solutions to at least reduce the amount of times we have to re-start our Github workflows or how long they take.

**Describe the solution you'd like**
Solution options (could be a variation of multiple options):

1. Explore the use of self hosted GitHub runners where we can take greater control of what machines are used to run our CI.
    1. This has a few downsides like increased cost, managing our own infrastructure and security issues (OpenSearch core currently isn’t pursuing this)
2. Wait for public Jenkins (potential release April 6th as mentioned in this issue’s comments) and AD CI too it, we can have an increased instance size on Jenkins to reduce the failing tests. Need to do more research on how this exactly would work and in terms of our GitHub commit process right now.
3. Further divide our workflow into smaller tasks. For example we can potentially separate the `Build and Run Tests` workflow into smaller tasks. We can also investigate separating out specific integ-tests into different workflows. Reducing how long each tasks runs has proven to reduce flakiness on OpenSearch core testing. *(This can be done anyways regardless of other options)*
    1. Additionally we can investigate if more workflows can run in parallel to reduce CI time (not necessarily flakiness)
4. Separate which specific integration tests we are running for each new commit/PR, when we merge to main and when infra team does validation testing. 
    1. Full integration testing could still be done on a new PR but for example run only a set of the integration tests when infra team does validation testing
5. https://blog.gradle.org/gradle-flaky-test-retry-plugin

PR #456, opened 2022-03-22
Adding test-retry plugin

amitgalitz:

### Description
Testing out new gradle plugin that retries failed test.
* Added this to both unit tests and integ tests
* `maxFailures`: "The maximum number of test failures that are allowed before retrying is disabled." (set to 10)
* `maxRetries`: "The maximum number of times to retry an individual test." (set to 10)
* `getFailOnPassedAfterRetry`: "Whether tests that initially fail and then pass on retry should fail the task." (set to **false** since if it passes once its good enough to pass build)

**Will be re-running workflows a few times on this PR to try this out and potentially adjust max failures/retries setting.** https://blog.gradle.org/gradle-flaky-test-retry-plugin

### Related Issues
#451 
 
### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ plugins {
     id 'nebula.ospackage' version "8.3.0" apply false
     id "com.diffplug.gradle.spotless" version "3.26.1"
     id 'java-library'
-    id 'org.gradle.test-retry' version '1.0.0'
+    id 'org.gradle.test-retry' version '1.3.1'
 }
 
 tasks.withType(JavaCompile) {
@@ -154,7 +154,8 @@ test {
     retry {
         if (isCiServer) {
             failOnPassedAfterRetry = false
-            maxRetries = 8
+            maxRetries = 6
+            maxFailures = 10
         }
     }
     include '**/*Tests.class'
@@ -172,7 +173,8 @@ integTest {
     retry {
         if (isCiServer) {
             failOnPassedAfterRetry = false
-            maxRetries = 8
+            maxRetries = 6
+            maxFailures = 10
         }
     }
     dependsOn "bundlePlugin"
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,7 @@ plugins {
     id 'nebula.ospackage' version "8.3.0" apply false
     id "com.diffplug.gradle.spotless" version "3.26.1"
     id 'java-library'
+    id 'org.gradle.test-retry' version '1.0.0'
 }
 
 tasks.withType(JavaCompile) {
@@ -150,6 +151,11 @@ def opensearch_tmp_dir = rootProject.file('build/private/opensearch_tmp').absolu
 opensearch_tmp_dir.mkdirs()
 
 test {
+    retry {
+        failOnPassedAfterRetry = false
+        maxFailures = 10
+        maxRetries = 10
+    }
     include '**/*Tests.class'
     systemProperty 'tests.security.manager', 'false'
 }
@@ -162,6 +168,11 @@ task integTest(type: RestIntegTestTask) {
 tasks.named("check").configure { dependsOn(integTest) }
 
 integTest {
+    retry {
+        failOnPassedAfterRetry = false
+        maxFailures = 10
+        maxRetries = 10
+    }
     dependsOn "bundlePlugin"
     systemProperty 'tests.security.manager', 'false'
     systemProperty 'java.io.tmpdir', opensearch_tmp_dir.absolutePath
--- a/build.gradle
+++ b/build.gradle
@@ -149,12 +149,13 @@ def _numNodes = findProperty('numNodes') as Integer ?: 1
 
 def opensearch_tmp_dir = rootProject.file('build/private/opensearch_tmp').absoluteFile
 opensearch_tmp_dir.mkdirs()
-
+boolean isCiServer = System.getenv().containsKey("CI")
 test {
     retry {
-        failOnPassedAfterRetry = false
-        maxFailures = 10
-        maxRetries = 10
+        if (isCiServer) {
+            failOnPassedAfterRetry = false
+            maxRetries = 8
+        }
     }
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -148,8 +149,15 @@ def _numNodes = findProperty('numNodes') as Integer ?: 1
...
 def opensearch_tmp_dir = rootProject.file('build/private/opensearch_tmp').absoluteFile
 opensearch_tmp_dir.mkdirs()
-
+boolean isCiServer = System.getenv().containsKey("CI")
 test {
+    retry {
+        if (isCiServer) {
+            failOnPassedAfterRetry = false
+            maxRetries = 6
+            maxFailures = 10
```
ylwu-amzn, line 158: Haven't used test retry, just asking: if `maxRetries` is 6, the max failure should't exceed 7? Why set `maxFailures = 10`?

src/test/java/org/opensearch/ad/e2e/DetectionResultEvalutationIT.java
```diff
@@ -592,26 +593,14 @@ private void indexTrainData(String datasetName, List<JsonObject> data, int train
...
-                    break;
-                } catch (Throwable throwable) {
-                    LOG.info("Retry restart test case", throwable);
-                    cleanUpCluster();
-                    wipeAllODFEIndices();
-                }
-            }
-            assertTrue("failed all retries", i < maxRetries);
+            disableResourceNotFoundFaultTolerence();
+            verifyRestart("synthetic", 1, 8);
```
ohltyler, line 603: Previous code had `cleanUpCluster()` and `wipeAllODFEIndices()` if there was some failure. Will the new retry logic handle this for us?

amitgalitz: good catch. I don't believe the retry plugin does these operations, maybe we should add these operations to the end of the test?

ohltyler: What about removing the for loop still, but adding back the try/catch? That way you can keep the same functionality, where if there's some exception it will run `cleanUpCluster()`/`wipeAllODFEIndices()`. Can also fail the test in catch block, else pass it.

amitgalitz: This makes sense to me and I'll add a `fail()` in catch block as you mention, only caveat is that there could be cases where `verifyRestart` fails and no exception thrown so it won't run clean and wipe methods. However, i think when this test is flaky it might always throw the exception and not just fail on assertion before. @kaituo can you confirm? I

ohltyler (commented):
LGTM! Just wondering, I feel this can cause flaky tests to stick around longer without fixing them. Is there still a way to get a report of the tests that may have failed / been retried? Ideally those are still brought to our attention so they can be fixed, assuming that the test suite is run in such a way or in a different env where they won't automatically be retried.

amitgalitz:
> LGTM! Just wondering, I feel this can cause flaky tests to stick around longer without fixing them. Is there still a way to get a report of the tests that may have failed / been retried? Ideally those are still brought to our attention so they can be fixed. That way, in case the test suite is run in such a way or in a different env where they won't automatically be retried, it's not a surprise.

Thats a valid concern, in order to mitigate this the retry-gradle will only work when running through github actions and not when building or testing locally so at least if developers build locally they can encounter flakiness first. However there are obviously times when we don't do the whole ./gradlew :build process locally before making a PR. If tests have failed and then passed on retry, the failures will show up on the action logs if developers check this, however it also makes sense that if a developer sees all checks passing then they wont check the logs. Opensearch has a bot that actually produces the reports and lets you see all failed tests in a nice format like we can do locally, opensearch core also has just implement gradle-retry and validated that they can see the flaky tests in their report( same as we can in our logs) https://github.com/opensearch-project/OpenSearch/pull/2638#issuecomment-1082503761 (if you click report it will download a zip that includes the nicely formatted test results as seen in screenshots in that PR)

build.gradle
```diff
@@ -148,8 +149,15 @@ def _numNodes = findProperty('numNodes') as Integer ?: 1
 
 def opensearch_tmp_dir = rootProject.file('build/private/opensearch_tmp').absoluteFile
 opensearch_tmp_dir.mkdirs()
-
+boolean isCiServer = System.getenv().containsKey("CI")
```
ohltyler, line 152: How did you come up with this value? Will this only work in github CI runners, or will it work in Jenkins hosts too? Ideally it's run in both so that we don't get surprises on test failures during infra builds.

ohltyler:
> 

Got it - yeah I think as long as this still fails in localhost & being able to view the reports from CI, should be ok. Can you help add the bot workflow that comments the zipped test output?

build.gradle
amitgalitz: This value was recommended by https://github.com/gradle/test-retry-gradle-plugin documentation and gradle documentation recommends this line when running something only in CI. I tested it and it works on github actions, I am not sure about jenkins. I saw this also recommended on the opensearch-core PR `if (BuildParams.isCi() == true)` but wasn't used in the end. I was thinking maybe its okay if we don't retry on jenkins as that is run on a larger instance and I didn't AD backend was as flaky there?

ohltyler:
> > 
> 
> Got it - yeah I think as long as this still fails in localhost & being able to view the reports from CI, should be ok. Can you help add the bot workflow that comments the zipped test output?

I see you've created #480 to track, sounds good

ohltyler (approved):
LGTM!

ylwu-amzn (approved):

dblock:
Nice to see plugins adopt this for flakey tests.

We had some discussions in https://github.com/opensearch-project/OpenSearch/pull/2638 about default values. I would take a look and lower the number of retries to catch flakey tests, and consider removing the "is CI" check to match the CI and developer experience, we had problems with this in core.

PR #495, opened 2022-04-08
Check if indices exist in the presence of empty search results

kaituo:
### Description
Previously, CompositeRetriever throws an IllegalArgumentException in the presence of empty results, which gets translated to internal failure and increments AD failure count. When the source index is a regex like blah*, we will get an empty response even if the index does not exist. This PR checks indices exist in the presence of empty search results. If yes, we throw an IndexNotFoundException that ends up being converted to EndRunException; if no, we still throw an IllegalArgumentException.

Testing done:
1. added unit tests
2. reproduced manually and verified the change fixed the issue.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
+++ b/src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
@@ -14,6 +14,7 @@ package org.opensearch.ad.feature;
 import java.io.IOException;
 import java.time.Clock;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
@@ -25,13 +26,17 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
+import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.model.Feature;
 import org.opensearch.ad.util.ParseUtils;
 import org.opensearch.client.Client;
+import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
+import org.opensearch.index.IndexNotFoundException;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.RangeQueryBuilder;
 import org.opensearch.search.aggregations.Aggregation;
@@ -66,6 +71,8 @@ public class CompositeRetriever extends AbstractRetriever {
     private final int pageSize;
     private long expirationEpochMs;
     private Clock clock;
+    private IndexNameExpressionResolver indexNameExpressionResolver;
+    private ClusterService clusterService;
 
     public CompositeRetriever(
         long dataStartEpoch,
@@ -77,7 +84,9 @@ public class CompositeRetriever extends AbstractRetriever {
         Clock clock,
         Settings settings,
         int maxEntitiesPerInterval,
-        int pageSize
+        int pageSize,
+        IndexNameExpressionResolver indexNameExpressionResolver,
+        ClusterService clusterService
     ) {
         this.dataStartEpoch = dataStartEpoch;
         this.dataEndEpoch = dataEndEpoch;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -301,12 +337,12 @@ private boolean shouldRetryDueToEmptyPage(SearchResponse response) {
...
         /**
          * Whether next page exists.  Conditions are:
-         * 1) we haven't fetched any page yet (totalResults == 0) or afterKey is not null
+         * 1) this is the first time we query (iterations == 0) or afterKey is not null
          * 2) next detection interval has not started
          * @return true if the iteration has more pages.
          */
         public boolean hasNext() {
-            return (totalResults == 0 || (totalResults > 0 && afterKey != null)) && expirationEpochMs > clock.millis();
+            return (iterations == 0 || (totalResults > 0 && afterKey != null)) && expirationEpochMs > clock.millis();
```
ylwu-amzn, line 345: Is it possible `iterations == 0` and `afterKey == null`? Not sure if this is enough `afterKey != null && expirationEpochMs > clock.millis()`

src/test/java/org/opensearch/ad/transport/MultiEntityResultTests.java
```diff
@@ -1283,4 +1287,45 @@ public void testSelectHigherExceptionInModelNode() throws InterruptedException,
...
+
+    /**
+     * A missing index will cause the search result to contain null aggregation
+     * like {"took":0,"timed_out":false,"_shards":{"total":0,"successful":0,"skipped":0,"failed":0},"hits":{"max_score":0.0,"hits":[]}}
+     *
+     * The test verifies we can handle such situation and won't throw exceptions
+     * @throws InterruptedException while waiting for execution gets interruptted
+     */
+    public void testMissingIndex() throws InterruptedException {
+        // when(indexNameResolver.concreteIndexNames(any(), any(), any(String[].class))).thenReturn(new String[] {});
```
amitgalitz, line 1299: Is this supposed to be commented out?

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -284,11 +303,28 @@ private boolean shouldRetryDueToEmptyPage(SearchResponse response) {
...
+            // Without regex, we will get IndexNotFoundException instead.
+            // {"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such
+            // index
+            // [blah]","index":"blah","resource.id":"blah","resource.type":"index_or_alias","index_uuid":"_na_"}],"type":"index_not_found_exception","reason":"no
+            // such index
+            // [blah]","index":"blah","resource.id":"blah","resource.type":"index_or_alias","index_uuid":"_na_"},"status":404}%
             if (response == null || response.getAggregations() == null) {
-                return Optional.empty();
+                List<String> sourceIndices = anomalyDetector.getIndices();
+                String[] concreteIndices = indexNameExpressionResolver
```
amitgalitz, line 317: Does this basically check if at least one of the indices in the pattern exists? Also This method won't get executed if detector has a single index used and it doesn't exist as an exception will be thrown earlier right?

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: > Is it possible iterations == 0 and afterKey == null
It is possible. 

> afterKey != null && expirationEpochMs > clock.millis()
This is not enough as at the beginning afterKey = null.

src/test/java/org/opensearch/ad/transport/MultiEntityResultTests.java
kaituo: yeah, removing.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: > Does this basically check if at least one of the indices in the pattern exists?

Yes

> Also This method won't get executed if detector has a single index used and it doesn't exist as an exception will be thrown earlier right?

For single index, you meant just index name without regex, right? If yes, right. We will get index not found exception before reaching here.

amitgalitz: Yeah meant just index name without regex.

amitgalitz (approved):
Thanks for making this fix!

ylwu-amzn (dismissed):
LGTM.

amitgalitz (approved):

ylwu-amzn (approved):

PR #501, opened 2022-04-14
Change 2.0-alpha1 to 2.0-rc1.

dblock:
Following https://github.com/opensearch-project/opensearch-build/pull/1863, change 2.0-alpha1 to 2.0-rc1.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,9 +17,9 @@ buildscript {
     ext {
         opensearch_group = "org.opensearch"
         isSnapshot = "true" == System.getProperty("build.snapshot", "true")
-        opensearch_version = System.getProperty("opensearch.version", "2.0.0-alpha1-SNAPSHOT")
-        buildVersionQualifier = System.getProperty("build.version_qualifier", "alpha1")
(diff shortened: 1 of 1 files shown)


amitgalitz:
I think we need job-scheduler to merge first with this change and then I'll re-run workflow

amitgalitz:
@peterzhuamazon still fails on CI. but locally passes. Might be a weird thing with the `tar` that is now added to the url: download link we have in `build.gradle` is `https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.0.0-rc1/latest/linux/x64/builds/opensearch/plugins/opensearch-job-scheduler-2.0.0.0-rc1.zip` doesn't work. But online `https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.0.0-rc1/latest/linux/x64/tar/builds/opensearch/plugins/opensearch-job-scheduler-2.0.0.0-rc1.zip` works. when I try the URI on chrome it works with tar but it also auto removes the tar from the link

amitgalitz:
Closing as #503 was merged instead

PR #519, opened 2022-04-22
bump rcf to 3.0-rc2.1

amitgalitz:

### Description
1. bumped RCF to 3.0-rc2.1 from maven
2. changed V1JsonToV2StateConverter usage V1JsonToV3StateConverter
3. Added test to deserialize model from rc1 and to see if scores are the same using rc1 checkpoint and rc2.1 dependency and rc1 dependency with the same data.

Other manual testing done:
1. real time single stream and HCAD (including create/start detector, see detector emit results; then stop cluster and restart it to see if results continue to show)
2. historical single stream and HCAD
3. backward compatible tests: v1 to v3 model, v2 to v3 model (B/G simulation)
4. Also answering this question -> "given a v2 model interpreted by rc1 and v2 model interpreted by rc2, would they produce the same score (rcf score and grade)? We expect those should be the same." (Added unit test for this)
5. Memory usage on average is 20-30% smaller, keeping the same memory formula however for worst case scenarios.
 
### Check List


src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
```diff
@@ -1000,4 +1005,77 @@ private double[] getPoint(int dimensions, Random random) {
...
+        String json = Files.readString(Paths.get(filePath), Charset.defaultCharset());
+        Map map = gson.fromJson(json, Map.class);
+        String model = (String) ((Map) ((Map) ((ArrayList) ((Map) map.get("hits")).get("hits")).get(0)).get("_source")).get("modelV2");
+        ThresholdedRandomCutForest forest = checkpointDao.toTrcf(model);
+        assertEquals(256, forest.getForest().getSampleSize());
+        assertEquals(8, forest.getForest().getShingleSize());
+        assertEquals(30, forest.getForest().getNumberOfTrees());
+    }
+
+    public void testDeserializeTRCFModel() throws Exception {
```
kaituo, line 1033: Could you add the test's purpose?

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
```diff
@@ -1000,4 +1005,77 @@ private double[] getPoint(int dimensions, Random random) {
         }
         return point;
     }
+
+    public void testDeserializeRCFModelPreINIT() throws Exception {
```
kaituo, line 1009: are the following 2 tests used for single-stream detector's checkpoints? If so, could you add comments?

kaituo (approved):

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -367,7 +367,7 @@ public Collection<Object> createComponents(
         mapper.setSaveExecutorContextEnabled(true);
         mapper.setSaveTreeStateEnabled(true);
         mapper.setPartialTreeStateEnabled(true);
-        V1JsonToV2StateConverter converter = new V1JsonToV2StateConverter();
+        V1JsonToV3StateConverter converter = new V1JsonToV3StateConverter();
```
ylwu-amzn, line 370: Do we need V2 to V3 converter?  How about add some comments to explain what's V1, V2 and V3?

amitgalitz: After testing and looking at RCF code, we don't need a converter between v2 to v3 as that is dealt with accordingly on RCF side. I also added a unit test where we parse a v2 checkpoint with RCF3.0-rc2.1 as a dependency(v3) and we get the same result as we do when parsing with rc1(v2). I'll add some comments explaining v1, v2, v3 in checkPointDAO class.

kaituo (dismissed):

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -117,6 +117,14 @@ public class CheckpointDao {
 
     private Gson gson;
     private RandomCutForestMapper mapper;
+
+    // For further reference v1, v2 and v3 refer to the different variations of RCF models
+    // used by AD. v1 was originally used with the lunch of OS 1.0. We later converted to v2
```
kaituo, line 122: lunch -> launch

amitgalitz: fixed

kaituo (dismissed):

ylwu-amzn (dismissed):

ylwu-amzn (approved):

kaituo (approved):

Issue #540, opened 2022-05-10
[BUG] Cold start not working after we have done cold start once and earlier historical data is deleted


kaituo:
**What is the bug?**
The bug happens because we use job enabled time as training data end time (https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java#L431). But if the historical data before that time is deleted or does not exist cold start might never finish. We should use current time as the training data end time and retry in a few intervals.

PR #547, opened 2022-05-13
Use current time as training data end time

kaituo:
### Description
The bug happens because we use job enabled time as training data end time. But if the historical data before that time is deleted or does not exist at all, cold start might never finish. This PR uses current time as the training data end time so that cold start has a chance to succeed later. This PR also removes the code that combines cold start data and existing samples in EntityColdStartWorker because we don't add samples until cold start succeeds. Combining cold start data and existing samples is thus unnecessary.

Testing done:
1. manually verified the bug is fixed.
2. fixed all related unit tests.

Note: I found the bug in 1.2, so I started with 1.2 branch. Will forward push to 1.3 and 2.x later.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/540

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -694,17 +694,17 @@ public class EntityColdStarter implements MaintenanceState {
      *
      * @param coldstartDatapoints training data generated from cold start
      * @param modelId model Id
-     * @param entityState entity State
+     * @param modelState entity State
      */
-    private void extractTrainSamples(List<double[][]> coldstartDatapoints, String modelId, ModelState<EntityModel> entityState) {
-        if (coldstartDatapoints == null || coldstartDatapoints.size() == 0 || entityState == null) {
+    private void extractTrainSamples(List<double[][]> coldstartDatapoints, String modelId, ModelState<EntityModel> modelState) {
+        if (coldstartDatapoints == null || coldstartDatapoints.size() == 0 || modelState == null) {
             return;
         }
 
-        EntityModel model = entityState.getModel();
+        EntityModel model = modelState.getModel();
         if (model == null) {
             model = new EntityModel(null, new ArrayDeque<>(), null);
-            entityState.setModel(model);
+            modelState.setModel(model);
         }
 
         Queue<double[]> newSamples = new ArrayDeque<>();
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -21,6 +21,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -46,7 +47,6 @@ import org.opensearch.ad.dataprocessor.Interpolator;
 import org.opensearch.ad.feature.FeatureManager;
 import org.opensearch.ad.feature.SearchFeatureDao;
 import org.opensearch.ad.model.AnomalyDetector;
-import org.opensearch.ad.model.AnomalyDetectorJob;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.model.IntervalTimeConfiguration;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
@@ -220,6 +220,22 @@ public class EntityColdStarter implements MaintenanceState {
     ) {
         logger.debug("Trigger cold start for {}", modelId);
 
+        if (modelState == null || entity == null) {
+            listener
+                .onFailure(
+                    new IllegalArgumentException(
+                        String
+                            .format(
+                                Locale.ROOT,
+                                "Cannot have empty model state or entity: model state [%b], entity [%b]",
+                                modelState == null,
+                                entity == null
+                            )
+                    )
+                );
+            return;
+        }
+
         if (lastThrottledColdStartTime.plus(Duration.ofMinutes(coolDownMinutes)).isAfter(clock.instant())) {
             listener.onResponse(null);
             return;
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -220,6 +220,22 @@ private void coldStart(
     ) {
         logger.debug("Trigger cold start for {}", modelId);
 
+        if (modelState == null || entity == null) {
```
amitgalitz, line 223: If data isn't present currently but will get ingested in the future and we are expecting a long initialization would either of this condition be met until then?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -404,42 +419,23 @@ private void getEntityColdStartData(String detectorId, Entity entity, ActionList
             ActionListener<Optional<Long>> minTimeListener = ActionListener.wrap(earliest -> {
                 if (earliest.isPresent()) {
                     long startTimeMs = earliest.get().longValue();
-                    nodeStateManager.getAnomalyDetectorJob(detectorId, ActionListener.wrap(jobOp -> {
-                        if (!jobOp.isPresent()) {
-                            listener.onFailure(new EndRunException(detectorId, "AnomalyDetector job is not available.", false));
```
amitgalitz, line 409: Before it seems like we used cold start data and samples for the training data. Can you further explain why we are moving away from this or was no sample data actually ever being used because we didn't add samples until success anyways (as you mention in description)?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -694,40 +690,30 @@ public void trainModelFromExistingSamples(ModelState<EntityModel> modelState, in
...
+     * Extract training data and put them into ModelState
      *
      * @param coldstartDatapoints training data generated from cold start
      * @param modelId model Id
      * @param entityState entity State
      */
-    private void combineTrainSamples(List<double[][]> coldstartDatapoints, String modelId, ModelState<EntityModel> entityState) {
-        if (coldstartDatapoints == null || coldstartDatapoints.size() == 0) {
+    private void extractTrainSamples(List<double[][]> coldstartDatapoints, String modelId, ModelState<EntityModel> entityState) {
+        if (coldstartDatapoints == null || coldstartDatapoints.size() == 0 || entityState == null) {
```
amitgalitz, line 700: Is this needed for extra safety to access `.getModel` later on? Since we know that if modelState was null earlier an exception would've been thrown so this might not be needed. Also should we keep naming consistent or indicate reason of transition from modelState to entityState name.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -694,40 +690,30 @@ public void trainModelFromExistingSamples(ModelState<EntityModel> modelState, in
...
             return;
         }
 
         EntityModel model = entityState.getModel();
         if (model == null) {
             model = new EntityModel(null, new ArrayDeque<>(), null);
+            entityState.setModel(model);
```
amitgalitz, line 707: What is the reason for setting an empty model now where previously we haven't done this.

amitgalitz:
To clarify with an example, is the bug here that we might have historical data from 1pm to 6pm. Current time is 6:10pm. Could job enabled time be something like 4pm after running cold start once or more already? And then we wont use data past 4 pm for training? With this PR we will use the current time of 6:10pm as training data end time? Will this add implications for long initializations that requires more data to be ingested for initialization to end anyways?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: This is method invariant and we don't any expect any of them to be true, regardless of whether data is present or not.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: No sample data actually ever being used because we didn't add samples until success.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: yes, it is for extra safety. It is not needed now but is helpful to prevent unintended bugs as we may forget the invariant.

yes, let me change to use modelState.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: not setting an empty model is a bug. But because the model will not be empty in current workflow so it does not manifest itself.

kaituo:
> 

The example is right except the timing number may need to change. Depends on the interval, the data fetching algorithm is different (check https://github.com/opensearch-project/anomaly-detection/commit/2ce24a04521c24dd9de5d0b4bfb2919ce8e800a1). In one scenario, we will use last 40 samples with two tries. At most, we will look back 80 intervals. So if the interval is at least 2 minutes, we will use some of the data older than 4 pm.

What you said can happen. If we keep current code, users will have to not delete data before enabled time or manipulate data timestamps to make it look old to make following cold start to succeed. After the fix, they can keep ingesting new data and the cold start will succeed eventually. The latter UX seems better to me.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -404,42 +419,23 @@ private void getEntityColdStartData(String detectorId, Entity entity, ActionList
...
-                        // anomalies in the process.
-                        // An ideal solution would be to record time stamps of data points and combine existing
-                        // samples and cold start samples and do interpolation afterwards. Recording time stamps
-                        // requires changes across the board like bwc in checkpoints. A pragmatic solution is to use
-                        // job enabled time as the end time of cold start period as it is easier to combine
-                        // existing samples with cold start data. We just need to appends existing samples after
-                        // cold start data as existing samples all happen after job enabled time. There might
-                        // be some gaps in between the last cold start sample and the first accumulated sample.
-                        // We will need to accept that precision loss in current solution.
-                        long endTimeMs = job.getEnabledTime().toEpochMilli();
```
ylwu-amzn, line 430: If HC detector realtime job not restarted, the enabled time won't change. If there is no enough data before job enabled time and user don't backfill historical data, there is no chance to pass cold start, right? Seems a critical bug if that's true. We'd better backfill to 1.x too.

kaituo: right. yes, will backfill.

ylwu-amzn (approved):
LGTM. Thanks for fixing!

amitgalitz (approved):
LGTM, thanks for making the fix!

Issue #539, opened 2022-05-10
[FEATURE] remove sampling/interpolation in cold start


kaituo:
**Is your feature request related to a problem?**
The initial precision of detection is low due to the effects of cold start using linear interpolation. Garbage in, Garbage out. We should use real data to train.

kaituo:
implemented in the above PR.

PR #575, opened 2022-06-20
Disable interpolation in HCAD cold start

kaituo:
### Description
Previously, we used interpolation in HCAD cold start for the purpose of efficiency. This caused problems for model accuracy. This PR removes interpolation in the cold start step.

Testing done:
1. added unit tests to verify precision boosted.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/539

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
@@ -93,6 +93,7 @@ public class EntityColdStarter implements MaintenanceState, CleanState {
     private final long rcfSeed;
     private final int maxRoundofColdStart;
     private final double initialAcceptFraction;
+    private final boolean allowInterpolation;
 
     /**
      * Constructor
@@ -118,6 +119,7 @@ public class EntityColdStarter implements MaintenanceState, CleanState {
      * @param checkpointWriteQueue queue to insert model checkpoints
      * @param rcfSeed rcf random seed
      * @param maxRoundofColdStart max number of rounds of cold start
+     * @param allowInterpolation use interpolation or not
      */
     public EntityColdStarter(
         Clock clock,
@@ -137,7 +139,8 @@ public class EntityColdStarter implements MaintenanceState, CleanState {
         Duration modelTtl,
         CheckpointWriteWorker checkpointWriteQueue,
         long rcfSeed,
-        int maxRoundofColdStart
+        int maxRoundofColdStart,
+        boolean allowInterpolation
     ) {
         this.clock = clock;
         this.lastThrottledColdStartTime = Instant.MIN;
@@ -160,6 +163,7 @@ public class EntityColdStarter implements MaintenanceState, CleanState {
         this.rcfSeed = rcfSeed;
         this.maxRoundofColdStart = maxRoundofColdStart;
         this.initialAcceptFraction = numMinSamples * 1.0d / rcfSampleSize;
+        this.allowInterpolation = allowInterpolation;
     }
 
     public EntityColdStarter(
@@ -179,7 +183,8 @@ public class EntityColdStarter implements MaintenanceState, CleanState {
         Settings settings,
         Duration modelTtl,
         CheckpointWriteWorker checkpointWriteQueue,
-        int maxRoundofColdStart
+        int maxRoundofColdStart,
+        boolean allowInterpolation
     ) {
         this(
             clock,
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -590,18 +597,24 @@ private int calculateColdStartDataSize(List<double[][]> coldStartData) {
...
+        if (allowInterpolation) {
+            long delta = detector.getDetectorIntervalInMinutes();
+
+            int strideLength = defaulStrideLength;
+            int numberOfSamples = defaultNumberOfSamples;
+            if (delta <= 30 && 60 % delta == 0) {
+                strideLength = (int) (60 / delta);
+                numberOfSamples = (int) Math.ceil((shingleSize + numMinSamples) / 24.0d) * 24;
+            } else {
+                strideLength = 1;
```
amitgalitz, line 610: nit:
Case number 2 in the algorithm where the interval is above 30 and not divisable by 60 is the same as thew new case with no interpolation. We can perhaps factor this out? I also understand why to keep it this way because they might have the same output but for different reasons. It is just a little confusing as I read the entire comment as whole, and now the "uncommon case" is the default one pretty much.

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
```diff
@@ -842,11 +843,56 @@ public int searchInsert(long[] timestamps, long target) {
...
     }
 
     public void testAccuracyThirteenMinuteInterval() throws Exception {
-        accuracyTemplate(13);
+        accuracyTemplate(13, 0.5f, 0.5f);
+    }
+
+    public void testAccuracyOneMinuteIntervalNoInterpolation() throws Exception {
+        // make sure AnomalyDetectorSettings.ALLOW_INTERPLATION_IN_COLDSTART is set to false now
+        assertEquals(false, AnomalyDetectorSettings.ALLOW_INTERPLATION_IN_COLDSTART);
```
amitgalitz, line 855: love showing the improvement! Should we disable interpolation for other tests or test for other interval lengths?

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: I thought about combining conditions of the uncommon code. It would be

if (!allowInterpolation || (delta > 20 || 60%delta != 0))) {
  // uncommon case
} else {
  ...
}

This looks less readable than current implementation. What do you think?

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
kaituo: The improvement for other tests are not that much since the interpolated data for those cases are much less than the one-minute case. Thus, the improvement is not that obvious.

I kept other tests unchanged so cover the path using interpolation. Otherwise, our coverage will drop.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
amitgalitz: Don't have a strong preference, if you considered both and decided current as more readable then we can keep as is. Since we are now using no interpolation, should we add more to the comment or change anything since it seems like we are taking what was "uncommon" in the past as now to be the default.

kaituo: I added comment "If we disable interpolation, strideLength is 1 and numberOfSamples is shingleSize + numMinSamples;". Anything else you want to see?

amitgalitz: Just saw the new comments, thanks for adding it

amitgalitz (approved):
LGTM, thanks for the change

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: Thanks for your suggestion

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -258,6 +258,8 @@ private AnomalyDetectorSettings() {}
 
     public static final int MAX_IMPUTATION_NEIGHBOR_DISTANCE = 2;
 
+    public static final boolean ALLOW_INTERPLATION_IN_COLDSTART = false;
```
ylwu-amzn, line 261: Should we make this as dynamic setting?

kaituo: changed

ylwu-amzn (approved):

amitgalitz (approved):

PR #579, opened 2022-06-21
Add support for reproducible builds

inglor:
### Description
Add support for reproducible builds

As per gradle [docs] add support to remove timestamps and package with same order which is required from
[reproducible] builds

[docs]: https://docs.gradle.org/current/userguide/working_with_files.html#sec:archives
[reproducible]: https://reproducible-builds.org/

### Issues Resolved
N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,10 @@ tasks.withType(Test) {
 tasks.withType(Javadoc) {
     options.encoding = 'UTF-8'
 }
+tasks.withType(AbstractArchiveTask).configureEach {
+    preserveFileTimestamps = false
+    reproducibleFileOrder = true
+}
 
 repositories {
     mavenLocal()


amitgalitz:
Hi @inglor, thanks for providing links for documentation on reproducible build and how they are useful. I was wondering if you have any further particular reasoning for adding it to Anomaly Detection at the time, do you have a use case for this? Just to understand more context for this change.

inglor:
Hi @amitgalitz, see https://github.com/opensearch-project/OpenSearch/pull/1995#issuecomment-1023744621. Similarly this project fails unless the above change is applied.
As part of my efforts to make OpenSearch reproducible in Arch Linux I am making PRs to each of the subprojects (slowly). Still got some to do [1] 

[1]: https://archlinux.org/devel/reports/non-reproducible-packages/artafinde/

amitgalitz:
Just to confirm since we don't utilize the Gradle GlobalBuildInfo plugin, we don't need the other change you introduced in the OS PR?

inglor:
@amitgalitz Correct. This change is tested and produces reproducible builds as-is

amitgalitz (approved):
LGTM, thanks for adding this! Will need one more approval before merging

ohltyler (approved):

dblock:
@inglor Have you looked into making a change in OpenSearch `opensearch.plugin` task(s) so that the plugins can automatically inherit this and avoid making the PRs in every plugin?

dblock:
I raised https://github.com/opensearch-project/opensearch-plugin-template-java/pull/38#issuecomment-1262592602, maybe we shouldn't put this in every plugin repo.

prudhvigodithi:
Added some thoughts https://github.com/opensearch-project/opensearch-plugin-template-java/pull/38#issuecomment-1264690748, may be we should consider putting this reproducible build logic to one of the common custom gradle plugins used.

PR #585, opened 2022-06-24
Adding HCAD data ingestion script to AD

amitgalitz:

### Description
Initially want to give credit to @kaituo for a majority of the logic in this code. 
First iteration of data ingestion script for AD testing.
This script generates cosine wave data with anomalies injected with 2 fields (potential features).
The current code is set to have two categorical fields where you can decide how many entities are created (must be at least 1), users also have control over the ingestion_frequency and how many points are created.
Other params are also explained with the `help` section seen in the code.

I want to initially add this script and I will be iterating over it with the ability to change the number of fields and easily create single entity data.


dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+       }
+    }
+
+    es.indices.create(index=INDEX_NAME, body=request_body)
+
+
+'''
+    Posts a document(s) to the index
+'''
+@retry(delay=2)
```
kaituo, line 123: the default configuration retries using exponential backoff. Is is better than a hardcoded 2 seconds?

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+                    nextTs += datetime.timedelta(seconds=INGESTION_FREQUENCY)
+                    if totalCount - lastTotalCount > 1000000:
+                        # report progress every 1 million inserts
+                        print("totalCount {} thread_index {}".format(totalCount,
+                                                                     thread_index))
+                        lastTotalCount = totalCount
+            j += 1
+
+        if len(bulk_payload) > 0:
+            post_log(bulk_payload, thread_index)
```
kaituo, line 210: need to reset bulk_payload:

bulk_payload = list()

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+    except Error as err:
+        print("error: {0}".format(err))
```
kaituo, line 212: Could you add retries if any exceptions? Sth like

```
retries = 0
    while keep_loop and retries < 10 and j < host_number:
        try:
            while keep_loop and j < host_number:
             ...
         except:
            print("error: {0}".format(err))
            retries += 1
            client[thread_index] = create_client(security, URL)
```

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+        for j in range(0, base_dimension):
+            phase[j] = random.randint(0, period)
+            amp[j] = (1 + 0.2 * random.random()) * amplitude
+        cosine_param[i] = np.array([phase, amp])
+    return cosine_param
+
+'''
+    Main entry method for script
+'''
+def main():
```
kaituo, line 233: This part might needs refactoring as we need to extract a method create_client to be used elsewhere. See above comments.

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+from random import Random
+from retry import retry
+import urllib3
+import concurrent.futures
+import argparse
+
+from opensearchpy import OpenSearch, RequestsHttpConnection
+from opensearchpy import helpers
+
+# https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
```
kaituo, line 24: Could you add docs on how to use the script in typical scenario?

dataGeneration/requirements.txt
```diff
@@ -0,0 +1,5 @@
+numpy==1.21.2
```
kaituo, line 1: how do we use this file? Or is it just documentation?

amitgalitz: we can use `pip install -r requirements.txt` will add a readme with that info

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: this retry package default is at 0 https://pypi.org/project/retry/. But I can add a backoff param set to 2

dataGeneration/requirements.txt
ylwu-amzn: Any requirement for python version?

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+    # First, delete the index if it exists
+    print("Deleting index if it exists...")
+    es.indices.delete(index=INDEX_NAME, ignore=[400, 404])
+
+    # Next, create the index
+    print("Creating index \"{}\"...".format(INDEX_NAME))
+    request_body = {
+        "settings":{
+          "number_of_shards":shard_number,
+          "number_of_replicas": 0, # increase this number after indexing
```
ylwu-amzn, line 92: Is it some manual action to "increase this number after indexing" ?

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,296 @@
...
+#default of 1008 points with ingestion frequency set to 600 means there will basically be 1008 intervals = 7 days * 144 intervals/day
+POINTS = args.points 
+INGESTION_FREQUENCY = args.ingestion_frequency
+BULK_SIZE = args.bulk_size
+USERNAME = args.username
+PASSWORD = args.password
+NUMBER_OF_HISTORICAL_DAYS = args.number_of_historical_days
+
+
+index_name = "_index"
```
ylwu-amzn, line 69: Why not use uppercase for these constants? And seems this is not for index name, this is index name field?

amitgalitz: These are fields that are used in the index itself so I wanted to differentiate them from constants that are set by params. Specifically for this index_name one I could remove it and just add "_index" as I do "_source" for other fields like host and process, those I plan to potentially make configurable in the future.

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: Yes after indexing is done, user can change the replica number if they want too but setting it at 0 during ingestion leads to faster ingestion times.

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: will make the changes after trying out this changes on enough points, will run it over the weekend.

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: will make the changes after trying out this changes on enough points, will run it over the weekend.

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: adding it here also

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: added readme

dataGeneration/requirements.txt
amitgalitz: I am using 3.7 only so I stated that as required, not sure how to know what is minimum without manual checking

dataGeneration/generate-cosine-data-multi-entity.py
```diff
@@ -0,0 +1,294 @@
...
+parser.add_argument("-ep", "--endpoint", help="cluster endpoint", required=True)
+parser.add_argument("-i", "--index-name", help=" ",required=True)
+parser.add_argument("-shards", "--shards", type=int, help="The number of shards for the given index", required=True)
+parser.add_argument("-t", "--threads", type=int, help="The number of threads to be used for data ingestion, make sure given machine has enough", required=True)
+parser.add_argument("-bulk", "--bulk-size", type=int, default=3000, help="Number of documents per bulk request, default to 3000", )
+parser.add_argument("-ingest", "--ingestion-frequency", type=int, default=600, help="how often each respective document is indexed, for example the default is 600 seconds which equates to every 10 minutes")
+parser.add_argument("-p", "--points", type=int, default=1008, help="total number of points ingested, for example with 1008 points and a frequency of 600s, there will be 7 days of data")
+
+parser.add_argument('--security', action='store_true')
+parser.add_argument('--no-security', dest='security', action='store_false')
```
ylwu-amzn, line 38: Can we use something like `--security false` to replace `--no-security` ? But I'm ok for keep '--no-security'. Not a big problem.

kaituo (approved):

dataGeneration/generate-cosine-data-multi-entity.py
amitgalitz: This was giving my issues because it was taking true/false param as a string as no type=boolean available. Thought it was less clean in code to add if "False", return false and elseif "True" return True. What do you think

ylwu-amzn: sure, sounds good, let's keep as is.

ylwu-amzn (approved):

PR #610, opened 2022-07-12
Increment version to 1.3.4 and fix zip fetching

### LINKED ISSUE #586: BWC Failure on version increment
**What is the bug?**

Currently in order to run our BWC tests we need to pull the latest AD artifact from ci.opensearch.org. However, when increment version for AD we run into circular problem. For example when upgrading from 2.0 to 2.1, our CI tries to pull the latest AD version which is 2.1 from ci.openearch.org but since we haven't yet pushed our version increment commit, the artifact doesn't exist yet.

Version Increment PR: https://github.com/opensearch-project/anomaly-detection/pull/567

**Do you have any additional context?**

At the moment we are bypassing this check for the first intial PR by just merging the version increment which then leads to its upload so it can be fetched later on. We then either manually run the CI again on the main branch or check for BWC on the next upcoming PR.

This however isn't the right way to do things because we could miss a bwc failure on the initial PR merge. I propose that we add some sort of check that if there is no artifact available from AD we will build it during our CI run and utilize that zip itself.

amitgalitz:

### Description
bumped to 1.3.4
Changed fetching new AD version from gradle assemble instead of from ci.distribution.org

### Issues Resolved
#586

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -29,28 +29,28 @@ jobs:
 
       - name: Assemble anomaly-detection
         run: |
-          ./gradlew assemble -Dopensearch.version=1.3.3-SNAPSHOT
-          echo "Creating ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.3.3.0-SNAPSHOT ..."
-          mkdir -p ./src/test/resources/org/opensearch/ad/bwc/anomaly-detection/1.3.3.0-SNAPSHOT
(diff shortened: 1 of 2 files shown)


zelinh:
Version increment is part of the tasks for [Release 1.3.4](https://github.com/opensearch-project/opensearch-build/issues/2205)

ohltyler (approved):

zelinh:
@opensearch-project/anomaly-detection Can we get another approval on this PR so we could merge this and unblock for 1.3.4 release.

ylwu-amzn (approved):

PR #613, opened 2022-07-13
Incremented version to 2.1.1.

gaiksaya:
Coming from https://github.com/opensearch-project/opensearch-build/issues/1818 post release activities.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,9 +17,9 @@ buildscript {
     ext {
         opensearch_group = "org.opensearch"
         isSnapshot = "true" == System.getProperty("build.snapshot", "true")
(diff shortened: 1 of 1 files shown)


amitgalitz:
@gaiksaya the issue here is that the CI is trying to fetch 2.1.1 from online but it doesn't exist yet. This was fixed and merged into main and 2.x but not in time for 2.1. We could just merge this now and this CI will pass for the next 2.1.1 build that AD is in. Do you recommend doing that since we probably shouldn't merge this CI fix for 2.1?

gaiksaya:
Closing this as 2.2.0 is released and this version won't be released now.
Thanks!

Issue #22, opened 2021-04-24
Rename plugin


dblock:
**Is your feature request related to a problem? Please describe.**

Folder name (FolderID) is still using old naming (OpenDistro) .

Plugins should follow this naming convention:
Plugins ID / Folder Name after installation: opensearch-<pluginname> (All Lowercase)
Example ES: https://github.com/opensearch-project/index-management/blob/main/build.gradle#L104
Example Kibana: https://github.com/opensearch-project/index-management-dashboards-plugin/blob/main/package.json#L2
Plugin ZIP file Name: opensearch-<pluginname>-1.0.0.0-beta1.zip (All Lowercase)

ylwu-amzn:
And we should rename `opendistro-job-scheduler` to `opensearch-job-scheduler` for all plugins depending on job scheduler like [AD](https://github.com/opensearch-project/anomaly-detection/blob/main/build.gradle#L97), [ISM](https://github.com/opensearch-project/index-management/blob/main/build.gradle#L107).

PR #617, opened 2022-07-20
[Feature/extensions] Remove CommonUtils ThreadContext User from AD user checks

dbwiddis:
Signed-off-by: Daniel Widdis <widdis@gmail.com>

### Description
This copies the `User` class over from the Common Utils into a `UserIdentity` class.   Other than the class rename and addition of javadocs, there are no changes to the class.  This is done primarily to remove the external dependency but also preserve the intention (user and roles) as a placeholder until a replacement security authentication mechanism is implemented, and retains existing functionality.

In places where the User is retrieved from the thread context, a null user is used, as the original designers of this plugin determined that null = super user.  This will not be the path going forward but it is at least very convenient for us now.

Final changes were made to tests to recognize the lack of failed access, and coverage which is reduced by never executing code to respond to failures.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-sdk/issues/23

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
@@ -24,6 +24,7 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.ad.auth.UserIdentity;
 import org.opensearch.ad.indices.AnomalyDetectionIndices;
 import org.opensearch.ad.model.DetectionDateRange;
 import org.opensearch.ad.rest.handler.IndexAnomalyDetectorJobActionHandler;
@@ -36,7 +37,6 @@ import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
-import org.opensearch.commons.authuser.User;
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
 
@@ -88,7 +88,7 @@ public class AnomalyDetectorJobTransportAction extends HandledTransportAction<An
         ActionListener<AnomalyDetectorJobResponse> listener = wrapRestActionListener(actionListener, errorMessage);
 
         // By the time request reaches here, the user permissions are validated by Security plugin.
-        User user = getUserContext(client);
+        UserIdentity user = getUserContext(client);
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
             resolveUserAndExecute(
                 user,
@@ -126,7 +126,7 @@ public class AnomalyDetectorJobTransportAction extends HandledTransportAction<An
         long primaryTerm,
         String rawPath,
         TimeValue requestTimeout,
-        User user,
+        UserIdentity user,
         ThreadContext.StoredContext context
     ) {
         IndexAnomalyDetectorJobActionHandler handler = new IndexAnomalyDetectorJobActionHandler(
--- a/src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
@@ -15,7 +15,7 @@ import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_START_DETEC
 import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_STOP_DETECTOR;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.REQUEST_TIMEOUT;
-import static org.opensearch.ad.util.ParseUtils.getUserContext;
+import static org.opensearch.ad.util.ParseUtils.getNullUser;
 import static org.opensearch.ad.util.ParseUtils.resolveUserAndExecute;
 import static org.opensearch.ad.util.RestHandlerUtils.wrapRestActionListener;
 
@@ -88,7 +88,8 @@ public class AnomalyDetectorJobTransportAction extends HandledTransportAction<An
         ActionListener<AnomalyDetectorJobResponse> listener = wrapRestActionListener(actionListener, errorMessage);
 
         // By the time request reaches here, the user permissions are validated by Security plugin.
-        UserIdentity user = getUserContext(client);
+        // Temporary null user for AD extension without security. Will always execute detector.
+        UserIdentity user = getNullUser();
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
             resolveUserAndExecute(
                 user,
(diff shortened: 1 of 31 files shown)


dbwiddis:
Submitted for review.  Can't get tests to run due to 403 on snapshot repo.  I believe the below test should fail and will comment it out if so.
https://github.com/opensearch-project/anomaly-detection/blob/4baf75ca46da02c4e70016d34982f2c5a34d0acd/src/test/java/org/opensearch/ad/transport/handler/ADSearchHandlerTests.java#L85-L92

dbwiddis:
@owaiskazi19 @saratvemulapalli This is ready for review.  Summary of commits:
 - 66de852324fe9c439ba31b433dc450d6d1b99c13 Copied `User` class from Common Utils to a `UserIdentity` class here. Intention is:
   - Remove external dependency without changing other code (yet). This minimizes unnecessary code churn.
   - Keep a class as a placeholder where we will be using it to pass access tokens under the new security model.  
 - 4473415a487d8a942a8b83dc44a6ce6632ae70dd Deprecated the code which fetched the user from the Thread Context. Created a new temporary "null User" method that simply returns null (granting super-admin privileges).  
   - Could have modified the other code to substitute `null` instead of calling a "null User" method but this approach makes it easy to find those methods again in the future by tracing call hierarchy, instead of looking for uses of "null" in the codebase.
   - Eventually there will be a new method here to query the security plugin/core API/etc. to get the user access token (split token model) which will replace the temporary "null user" code.
 - bd83a39abc38256bf5bb3abbd5595e01d2f428e7 Updated tests which were intended to fail the user access checks, which now always pass thanks to the null (superadmin) user, so all test pass
 - 633c18d093fd930ce30d14e450b0e75469b9de1f The classes impacted by lack of failures have reduced code coverage.  Excluded them from JaCoCo coverage temporarily.  When the new security user code

saratvemulapalli (approved):
Thanks @dbwiddis for the changes.
Also thanks for putting in placeholders. 

@peternied could you take a look as well.

src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
```diff
@@ -88,7 +88,8 @@ protected void doExecute(Task task, AnomalyDetectorJobRequest request, ActionLis
         ActionListener<AnomalyDetectorJobResponse> listener = wrapRestActionListener(actionListener, errorMessage);
 
         // By the time request reaches here, the user permissions are validated by Security plugin.
-        User user = getUserContext(client);
+        // Temporary null user for AD extension without security. Will always execute detector.
+        UserIdentity user = getNullUser();
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
peternied, line 93: Seems like all the methods for ThreadContext were added for preserving/resetting the user from the thread context, a quick search says this can be removed.  What do you think about making that change now?

peternied (commented):
> In places where the User is retrieved from the thread context, a null user is used, as the original designers of this plugin determined that null = super user. This will not be the path going forward but it is at least very convenient for us now.

I would rather those checks for a null user became checks for `UserIdentity.Admin` that was returned from getUser when security was disabled.

This makes the more dangerous kind of actions much clearer in appearance as oppose to 'input validation' where we are checking for nulls. 

```java
if (getUser() == UserIdentity.Admin) {
   // DO EVERYTHING!
   ...
}
```

How does this align to what you were thinking?

src/main/java/org/opensearch/ad/transport/AnomalyDetectorJobTransportAction.java
dbwiddis: That's actually being handled in #22.  [You can see the diff here,](https://github.com/dbwiddis/anomaly-detection/compare/remove-user...dbwiddis:anomaly-detection:remove-client) awaiting this PR to be merged. :)

I went a step further and actually removed the whole thread pool since it's not needed for anything other than the thread context, although it remains an argument in the `Plugin` constructor so would be available for multithreading applications in the future.

dbwiddis: Also, with the changes in the above diff, `getUserContext(client)` now returns null so that null user placeholder could be removed again... but I prefer this naming as a reminder that we need to make some new method when we've re-done the `UserIdentity` class.

dbwiddis:
> I would rather those checks for a null user became checks for `UserIdentity.Admin` that was returned from getUser when security was disabled.

I completely agree with this preference and a future direction, but:
 - We haven't finished defining the `UserIdenity` class. Right now it's a carbon copy of `User` from Common Utils but we probably don't need any of the guts inside it.  They are there more as a reference for what a user used to be.
 - The fact is that the current code accepts `null` as a superuser, and hiding that flaw makes it less likely to fix it later!  I'd love to simultaneously switch from `null` user to an admin user along with disabling the superpower of `null`.  This is like putting a big orange cone on (or in) a pothole.  Doesn't fix the hole but at least you know driving down the street requires care and you can remove the cone when you fix the road.
 - There will be even more work with integrating identity in the future that we can integrate into the `UserIdentity` class, and the less I require now the more flexibility we have in defining exactly what we need later.

But I hope my snarky comments in the `getNullUser()` class indicate my feelings about that *not* being permanent. :)

dbwiddis:
> This makes the more dangerous kind of actions much clearer

Another thought, there should actually be no need under a new RBAC model for the extension to even know anything about whether the user is a superadmin or an anonymous user.  All we need is the access token that will be provided from the security features being moved from the plugin.  We won't know what access it gives, we'll just pass it along with our API requests and the security features will identify the access roles at that point.  So for the purposes of an extension, all users are the same.

peternied (approved):

src/main/java/org/opensearch/ad/util/ParseUtils.java
```diff
@@ -456,14 +456,30 @@ public static SearchSourceBuilder addUserBackendRolesFilter(User user, SearchSou
...
+
+    /**
+     * Placeholder method for running AD as an extension without security. User == null when security is disabled or user is super-admin.
+     *
+     * @return null.
+     */
+    public static UserIdentity getNullUser() {
+        // The author of this temporary code thinks it's a really bad idea to have a null object default to super-admin privileges.
+        // But that's how it is so here is the infinity gauntlet.
+        return null;
```
owaiskazi19, line 478: For my knowledge, if we are passing null here the role would be super-admin. Will that cause any impact of the functionality of AD plugin? I'm thinking more in a general way where we provide more access to admin as compared to other users

src/main/java/org/opensearch/ad/transport/DeleteAnomalyResultsTransportAction.java
```diff
@@ -61,7 +61,8 @@ protected void doExecute(Task task, DeleteByQueryRequest request, ActionListener
     }
 
     public void delete(DeleteByQueryRequest request, ActionListener<BulkByScrollResponse> listener) {
-        User user = getUserContext(client);
+        // Temporary null user for AD extension without security. Will always validate role.
+        UserIdentity user = getNullUser();
         try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
```
owaiskazi19, line 66: We are not targeting threadpool for SDKClient. Does having threadContext here will impact the client?

dbwiddis: The removal of the thread pool will be done in a pending PR that I'll submit as soon as this one is merged.  See earlier similar comment and my response here: https://github.com/opensearch-project/anomaly-detection/pull/617#discussion_r929516077

src/main/java/org/opensearch/ad/util/ParseUtils.java
dbwiddis: I'm not clear what you mean by "impact of the functionality".  It does provide the same level of access as either super-admin or security disabled, which is what we want in the near term.  See the changes in the tests, where a few tests that were supposed to fail for denied access ended up succeeding.

When we have developed security/identity features which include an access token, we will replace these null users with calls associated with the access token, which we will pass to OpenSearch via our REST API/Transport, and the access will be handled there.

owaiskazi19: Those tests answer my question. Thank you!

owaiskazi19 (approved):
LGTM

PR #619, opened 2022-07-26
[Feature/extensions] Remove client

dbwiddis:
### Description
This PR removes all dependencies on the Common Utils Thread Context.  It was initially used for users and roles, with most of the user code removed in #617.  This cleans up the remaining (useless/unused) fetching of the thread context from the code.

The thread context was pulled from the ThreadPool, which is also no longer needed or desired in an extension and has been removed in most code.  

There is still a remaining threadpool parameter in the `createComponents()` method which inherits/overrides `Plugin` and saves it as a field in the runner object, but it is okay to ignore that and not use it in later code; eventually that will no longer be needed.  A thread pool is also part of the `Client` constructor and there is a `threadpool()` getter from it.  While it may be harmless to keep that method around I think having it throw an exception as unused, or be deprecated, would be preferable.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-sdk/issues/22

### DIFF
--- a/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
+++ b/src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
@@ -95,7 +95,6 @@ import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.CheckedRunnable;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
-import org.opensearch.commons.InjectSecurity;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.RangeQueryBuilder;
 import org.opensearch.index.query.TermQueryBuilder;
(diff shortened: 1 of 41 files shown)


src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -1169,7 +1166,7 @@ private void detectAnomaly(
                 internalListener,
                 anomalyResults,
                 resultIndex,
-                () -> injectSecurity.close()
+                null
```
peternied, line 1169: Can we remove this parameter?

peternied (approved):
Great work, only a minor recommendation

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
dbwiddis: That was my initial gut feel, as this is the only known usage of that method so it possibly *could* be removed ...

But the method signature is more general with a `CheckedRunnable<?> runBefore` to support chaining result requests. In this particular usage, we aren't chaining a request, so `null` terminates the (potential) sequence of Runnables.  I figured it does no harm to leave it as is.

owaiskazi19 (approved):
LGTM!

saratvemulapalli (approved):

Issue #338, opened 2021-12-09
Even out hourly maintenance work for hourly cpu spike 


kaituo:
**Is your feature request related to a problem? Please describe.**
During the million-entity evaluation, CPU is mostly around 1% with hourly spike up to 65%.  An internal hourly maintenance job can account for the spike due to saving hundreds of thousands of model checkpoints, clears unused models, and performs bookkeeping for internal states.

![cpu_2](https://user-images.githubusercontent.com/5303417/145493982-7078140d-3011-40d9-ad37-667c66a4436e.png)

**Describe the solution you'd like**
 This may affect a cluster with constant workload. We may want do deep dive the spike and do something to even it out into larger window.

kaituo:
fixed in above PR.

PR #620, opened 2022-07-28
make 1M1min possible

kaituo:
### Description
This PR improves performance to make the 1M1min experiment possible. First, I changed coordinating node pagination from sync to async mode in AnomalyResultTransportAction so that the coordinating node does not have to wait for model nodes' responses before fetching the next page. Second, during the million-entity evaluation, CPU is mostly around 1% with hourly spikes up to 65%. An internal hourly maintenance job can account for the spike due to saving hundreds of thousands of model checkpoints, clearing unused models, and performing bookkeeping for internal states. This PR evens out the resource usage more fairly across a large maintenance window by introducing CheckpointMaintainWorker. Third, during a model corruption, I retrigger cold start for mitigation. Check ModelManager.score, EntityResultTransportAction, and CheckpointReadWorker.

Testing done:
1. Added unit tests.
2. Manually confirmed 1M1min is possible after the above changes.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/338

### DIFF
--- a/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
@@ -18,7 +18,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Optional;
-import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
@@ -36,7 +35,6 @@ import org.opensearch.ad.ratelimit.CheckpointMaintainRequest;
 import org.opensearch.ad.ratelimit.CheckpointMaintainWorker;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
 import org.opensearch.ad.ratelimit.RequestPriority;
-import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.util.DateUtils;
 
 /**
@@ -80,8 +78,7 @@ public class CacheBuffer implements ExpiringState {
     private final Clock clock;
     private final CheckpointWriteWorker checkpointWriteQueue;
     private final CheckpointMaintainWorker checkpointMaintainQueue;
-    private final Random random;
-    private final int checkpointIntervalHrs;
+    private int checkpointIntervalHrs;
 
     public CacheBuffer(
         int minimumCapacity,
@@ -92,8 +89,8 @@ public class CacheBuffer implements ExpiringState {
         Duration modelTtl,
         String detectorId,
         CheckpointWriteWorker checkpointWriteQueue,
-        Random random,
-        CheckpointMaintainWorker checkpointMaintainQueue
+        CheckpointMaintainWorker checkpointMaintainQueue,
+        int checkpointIntervalHrs
     ) {
         this.memoryConsumptionPerEntity = memoryConsumptionPerEntity;
         setMinimumCapacity(minimumCapacity);
@@ -108,9 +105,8 @@ public class CacheBuffer implements ExpiringState {
         this.clock = clock;
         this.priorityTracker = new PriorityTracker(clock, intervalSecs, clock.instant().getEpochSecond(), MAX_TRACKING_ENTITIES);
         this.checkpointWriteQueue = checkpointWriteQueue;
-        this.random = random;
         this.checkpointMaintainQueue = checkpointMaintainQueue;
-        this.checkpointIntervalHrs = AnomalyDetectorSettings.CHECKPOINT_SAVING_FREQ.toHoursPart();
+        setCheckpointIntervalHrs(checkpointIntervalHrs);
     }
 
     /**
@@ -350,7 +346,7 @@ public class CacheBuffer implements ExpiringState {
         List<CheckpointMaintainRequest> modelsToSave = new ArrayList<>();
         List<ModelState<EntityModel>> removedStates = new ArrayList<>();
         Instant now = clock.instant();
-        int currentHour = DateUtils.getHourOfDay(now);
+        int currentHour = DateUtils.getUTCHourOfDay(now);
         int currentSlot = currentHour % checkpointIntervalHrs;
         items.entrySet().stream().forEach(entry -> {
             String entityModelId = entry.getKey();
@@ -370,8 +366,8 @@ public class CacheBuffer implements ExpiringState {
                     removedStates.add(remove(entityModelId));
                 } else if (Math.abs(entityModelId.hashCode()) % checkpointIntervalHrs == currentSlot) {
                     // checkpoint is relatively big compared to other queued requests
-                    // save checkpoints with 1/6 probability as we expect to save
-                    // all every 6 hours statistically
+                    // Evens out the resource usage more fairly across a large maintenance window
+                    // by adding saving requests to CheckpointMaintainWorker.
                     //
                     // Background:
                     // We will save a checkpoint when
@@ -539,5 +535,18 @@ public class CacheBuffer implements ExpiringState {
         }
         this.minimumCapacity = minimumCapacity;
         this.reservedBytes = memoryConsumptionPerEntity * minimumCapacity;
+    }
+
+    public void setCheckpointIntervalHrs(int checkpointIntervalHrs) {
+        this.checkpointIntervalHrs = checkpointIntervalHrs;
+        // 0 can cause java.lang.ArithmeticException: / by zero
+        // negative value is meaningless
+        if (checkpointIntervalHrs <= 0) {
+            this.checkpointIntervalHrs = 1;
+        }
+    }
+
+    public int getCheckpointIntervalHrs() {
+        return checkpointIntervalHrs;
     }
 }
--- a/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/org/opensearch/ad/caching/CacheBuffer.java
@@ -32,8 +32,12 @@ import org.opensearch.ad.MemoryTracker.Origin;
 import org.opensearch.ad.ml.EntityModel;
 import org.opensearch.ad.ml.ModelState;
 import org.opensearch.ad.model.InitProgressProfile;
+import org.opensearch.ad.ratelimit.CheckpointMaintainRequest;
+import org.opensearch.ad.ratelimit.CheckpointMaintainWorker;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
 import org.opensearch.ad.ratelimit.RequestPriority;
+import org.opensearch.ad.settings.AnomalyDetectorSettings;
+import org.opensearch.ad.util.DateUtils;
 
 /**
  * We use a layered cache to manage active entities’ states.  We have a two-level
@@ -75,7 +79,9 @@ public class CacheBuffer implements ExpiringState {
     private final PriorityTracker priorityTracker;
     private final Clock clock;
     private final CheckpointWriteWorker checkpointWriteQueue;
+    private final CheckpointMaintainWorker checkpointMaintainQueue;
     private final Random random;
+    private final int checkpointIntervalHrs;
 
     public CacheBuffer(
         int minimumCapacity,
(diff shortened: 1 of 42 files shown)


src/main/java/org/opensearch/ad/caching/CacheBuffer.java
```diff
@@ -322,7 +368,7 @@ public List<ModelState<EntityModel>> maintenance() {
                     // already in the cache
                     // remove method saves checkpoint as well
                     removedStates.add(remove(entityModelId));
-                } else if (random.nextInt(6) == 0) {
+                } else if (Math.abs(entityModelId.hashCode()) % checkpointIntervalHrs == currentSlot) {
                     // checkpoint is relatively big compared to other queued requests
                     // save checkpoints with 1/6 probability as we expect to save
                     // all every 6 hours statistically
```
ohltyler, line 374: nit: can update this comment

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
```diff
@@ -342,7 +342,14 @@ Optional<CompositeAggregation> getComposite(SearchResponse response) {
          * @return true if the iteration has more pages.
          */
         public boolean hasNext() {
-            return (iterations == 0 || (totalResults > 0 && afterKey != null)) && expirationEpochMs > clock.millis();
+            long now = clock.millis();
+            if (expirationEpochMs <= now) {
+                LOG.info("Time is up, afterKey: " + afterKey + " " + expirationEpochMs + " " + now);
+            }
+            if ((iterations > 0 && afterKey == null) || totalResults > maxEntities) {
+                LOG.info("Finished in " + (now - startMs) + " msecs. ");
```
ohltyler, line 350: nit - should these be debug level logging?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -180,6 +180,9 @@ private AnomalyDetectorSettings() {}
 
     public static final Duration HOURLY_MAINTENANCE = Duration.ofHours(1);
 
+    // saving checkpoint every 6 hours.
+    public static final Duration CHECKPOINT_SAVING_FREQ = Duration.ofHours(6);
```
ohltyler, line 184: Is this the final decided value? I remember in a separate discussion that every 12 hours was a possibility.

Based on the 1 million entities experiments you've done, does 6 allow for the majority of models to be saved? Obviously the smaller the number the better, to allow for models to be saved more frequently.

ohltyler: Also, should this be exposed as a public setting for customers to adjust?

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -573,6 +595,18 @@ private AnomalyDetectorSettings() {}
...
+     * CheckpointMaintainRequest has model Id (roughly 256 bytes), and QueuedRequest
+     * fields including detector Id(roughly 128 bytes), expirationEpochMs (long,
+     *  8 bytes), and priority (12 bytes).
+     * Plus Java object size (12 bytes), we have roughly 416 bytes per request.
+     * We don't want the total size exceeds 0.1% of the heap.
+     * We can have at most 0.1% heap / 416 = heap / 416,000.
+     * For t3.small, 0.1% heap is of 1MB. The queue's size is up to
+     * 10^ 6 / 416 = 2403
+     */
+    public static int CHECKPOINT_MAINTAIN_REQUEST_SIZE_IN_BYTES = 416;
```
ohltyler, line 608: So given a single t3.small instance, we can optimally support 2403 entities per interval right? So given a frequency of 6 hrs, then given a detector with ~14k models would be able to be entirely saved. 

Is there any guidance from product or community on the tradeoff regarding the frequency of saving models / the importance of saving all models vs. how often they are saved?

src/main/java/org/opensearch/ad/util/DateUtils.java
```diff
@@ -0,0 +1,32 @@
...
+public class DateUtils {
+    public static final ZoneId UTC = ZoneId.of("Z");
+
+    /**
+     * Get hour of day of the input time in UTC
+     * @param instant input time
+     *
+     * @return Hour of day
+     */
+    public static int getHourOfDay(Instant instant) {
```
ohltyler, line 28: can this be renamed to make it more explicit that it's getting the UTC hour of day?

src/test/java/org/opensearch/ad/caching/CacheBufferTests.java
```diff
@@ -117,11 +118,58 @@ public void testMaintenance() {
...
 
+    @SuppressWarnings("unchecked")
+    public void testMaintainByHourNothingToSave() {
+        // hash code 49 % 6 = 1
+        String modelId1 = "1";
+        // hash code 50 % 6 = 2
+        String modelId2 = "2";
+        // hash code 51 % 6 = 3
+        String modelId3 = "3";
+        // hour 17. 17 % 6 (check point frequency) = 5
```
ohltyler, line 135: if there is changes in the check point frequency then this may fail.

src/main/java/org/opensearch/ad/caching/CacheBuffer.java
kaituo: updated.

src/main/java/org/opensearch/ad/feature/CompositeRetriever.java
kaituo: yeah, changed

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: yeah, will change to 12 hours in the new commit. 6 hours may or may not be enough. I changed to expose it as a public setting so that users can adjust based on their needs.

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
kaituo: It is a tradeoff: the more frequent, the more refreshed the saved model, but more resource consuming. I can add that as new doc in 2.2.

kaituo: The request size is one aspect. Together with queue size, they determine how many requests we can queue. But how many checkpoints we can really save depends on a few more factors (e.g., cluster load, the number of models in the cluster).

src/main/java/org/opensearch/ad/util/DateUtils.java
kaituo: yes

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -304,6 +303,9 @@ class PageListener implements ActionListener<CompositeRetriever.Page> {
 
         @Override
         public void onResponse(CompositeRetriever.Page entityFeatures) {
+            if (pageIterator.hasNext()) {
```
ylwu-amzn, line 306: If there are too many pages and not reach the last page until current job interval ends, will current job still iterate next page while next job triggers?

kaituo: yes

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -158,14 +162,14 @@ private ActionListener<Optional<AnomalyDetector>> onGetDetector(
     ) {
         return ActionListener.wrap(detectorOptional -> {
             if (!detectorOptional.isPresent()) {
-                listener.onFailure(new EndRunException(detectorId, "AnomalyDetector is not available.", true));
+                listener.onFailure(new EndRunException(detectorId, "AnomalyDetector is not available.", false));
```
ylwu-amzn, line 165: Why change to `false`? If detector is not found, do we still need to run job?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -158,14 +162,14 @@ private ActionListener<Optional<AnomalyDetector>> onGetDetector(
...
                 return;
             }
 
             AnomalyDetector detector = detectorOptional.get();
 
             if (request.getEntities() == null) {
-                listener.onResponse(null);
+                listener.onFailure(new EndRunException(detectorId, "Fail to get any entities from request.", false));
```
ylwu-amzn, line 172: For sparse data, it's possible that some interval has no data. With this change, we will stop AD job if there are no data for X intervals?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -196,35 +200,51 @@ private ActionListener<Optional<AnomalyDetector>> onGetDetector(
...
+                                    detectorId,
+                                    result.getGrade() > 0 ? RequestPriority.HIGH : RequestPriority.MEDIUM,
+                                    resultToSave,
+                                    detector.getResultIndex()
+                                )
+                            );
+                    }
+                } catch (IllegalArgumentException e) {
+                    // fail to score likely due to model corruption. Re-cold start to recover.
+                    cache.get().removeEntityModel(detectorId, modelId);
```
ylwu-amzn, line 238: Add some todo to track model corruption times?

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: It might be the node with the detector config is temporarily unavailable. Changing to false to to accommodate that.

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: if there is no data, model node should not receive the request in the first place. If request.getEntities() == null, there might be some bug in our code.

src/main/java/org/opensearch/ad/transport/EntityResultTransportAction.java
kaituo: I added the tracking code. Will send the new commit.

kaituo:
BWC test failed because we deprecated ODFE and the download link is archived.

ohltyler:
> BWC test failed because we deprecated ODFE and the download link is archived.

Have a fix coming soon for this.

kaituo:
> > BWC test failed because we deprecated ODFE and the download link is archived.
> 
> Have a fix coming soon for this - see #625

Thanks Tyler

kaituo:
Besides BWC tests, do you have other comments? @ohltyler @ylwu-amzn

ohltyler:
> Besides BWC tests, do you have other comments? @ohltyler @ylwu-amzn

I've taken a look at your latest commit - LGTM

src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -525,25 +537,25 @@ private AnomalyDetectorSettings() {}
     // expected execution time per cold entity request. This setting controls
     // the speed of cold entity requests execution. The larger, the faster, and
     // the more performance impact to customers' workload.
-    public static final Setting<Integer> EXPECTED_COLD_ENTITY_EXECUTION_TIME_IN_SECS = Setting
+    public static final Setting<Integer> EXPECTED_COLD_ENTITY_EXECUTION_TIME_IN_MILLISECS = Setting
         .intSetting(
-            "plugins.anomaly_detection.expected_cold_entity_execution_time_in_secs",
-            3,
+            "plugins.anomaly_detection.expected_cold_entity_execution_time_in_millisecs",
```
ylwu-amzn, line 542: Why change from second to millisecond? Will it have any bwc issue?

kaituo: it gives more control on the speed. Previously, we can only specify down to 1 second. Now we can specify down to 1 millisecond. It won't as these settings are not documented in public.

ylwu-amzn (approved):
LGTM. Thanks for the change!

ohltyler (approved):
LGTM!

ohltyler:
Failed tests will be fixed in #625, I think we can merge this and backport first.

PR #624, opened 2022-08-02
Staging for version increment automation

prudhvigodithi:
Description
This is the staging PR for to execute workflows that auto raise the version increment PR's
Example: https://github.com/prudhvigodithi/anomaly-detection/pull/1

Issues Resolved
Part of: https://github.com/opensearch-project/opensearch-build/issues/1375
From solution: https://github.com/opensearch-project/opensearch-build/issues/1375#issuecomment-1159476686
Related issue: https://github.com/opensearch-project/anomaly-detection/issues/605

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -687,3 +687,20 @@ validateNebulaPom.enabled = false
 tasks.withType(licenseHeaders.class) {
     additionalLicense 'AL   ', 'Apache', 'Licensed under the Apache License, Version 2.0 (the "License")'
 }
+
+// versionIncrement: Task to auto increment to the next development iteration
+task versionIncrement {
+    onlyIf { System.getProperty('newVersion') }
+    doLast {
+        ext.newVersion = System.getProperty('newVersion')
+        println "Setting version to ${newVersion}."
+        // String tokenization to support -SNAPSHOT
+        ant.replaceregexp(match: opensearch_version.tokenize('-')[0], replace: newVersion.tokenize('-')[0], flags:'g', byline:true) {
+            fileset(dir: projectDir) {
+                // Include the required files that needs to be updated with new Version
+                include(name: ".github/workflows/CI.yml")
+                include(name: "build.gradle")
+            }
+        }
+    }
+} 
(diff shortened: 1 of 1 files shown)


prudhvigodithi:
Referred old 1.x version increment PR https://github.com/opensearch-project/anomaly-detection/pull/611

ohltyler (approved):

amitgalitz:
@prudhvigodithi It seems like infra isn't actively building 1.4.0. Should that start happening before we continue developing on 1.x set on 1.4.0. For example this branch will continue being unstable with no 1.4.0 JS or common-utils until we either move to a different version or infra starts building 1.4.0 as well.

ohltyler:
@prudhvigodithi because we have already bumped `1.x` to `1.4`, can you rebase to branch `1.3` for a `1.3.5` release? Thanks!

amitgalitz:
wouldn't moving this branch to 1.3.5 cause issues for the auto version increment plugin since it is based on our version being the same as core for Y branch. Core currently points to 1.4, so based off the effort to make every plugin be the same as core. All plugins should be pointing to 1.4 and we should actively be building 1.4.@prudhvigodithi can you give further clarification on this? Our documentation says that 1.x should point to next minor release which would be 1.4 but this isn't getting built for some reason

prudhvigodithi:
Hey 1.x is for next minor version thats true, but any feature, enhancement we push, its should to 1.x and backported to other 1.x (example 1.3) branches (unless the fix or patch is only for 1.3 then the PR should be directly to 1.3 branch).
So @amitgalitz @ohltyler here this version increment task will be added to 1.x and should be backported to 1.3 branch.
Thank you

ohltyler:
> Hey 1.x is for next minor version thats true, but any feature, enhancement we push, its should to 1.x and backported to other 1.x (example 1.3) branches (unless the fix or patch is only for 1.3 then the PR should be directly to 1.3 branch). So @amitgalitz @ohltyler here this version increment task will be added to 1.x and should be backported to 1.3 branch. Thank you

Understood - thanks for the explanation!

amitgalitz (approved):
Thank you for pushing this to 1.x branch and I understand why it needs to be here. I just meant to highlight in above comments why 1.x branch CI will always fail

PR #674, opened 2022-09-23
Fix window delay test

amitgalitz:

### Description
Changed integration test so it doesn't default to exact window delay that assumes creation of data and test run time is in a few seconds. When I ran this test on my mac, the test took longer to run meaning more time has passed between the data creation and the validate API call. 

Meaning that it would fail with something like, "expected window delay to be 4 minutes but got 6 minutes if test took 2 minutes longer". I am not 100% sure why my macbook was so slow for this test where it didn't even fail on the github CI runner but regardless new logic makes more sense.

However with this fixed we can now, run, test and build AD on macos. 

### DIFF
--- a/src/test/java/org/opensearch/ad/e2e/DetectionResultEvalutationIT.java
+++ b/src/test/java/org/opensearch/ad/e2e/DetectionResultEvalutationIT.java
@@ -537,6 +537,8 @@ public class DetectionResultEvalutationIT extends ODFERestTestCase {
         List<JsonObject> data = createData(2000, recDetectorIntervalMillis);
         indexTrainData("validation", data, 2000, client);
         long detectorInterval = 4;
+        long expectedWindowDelayMillis = Instant.now().toEpochMilli() - data.get(0).get("timestamp").getAsLong();
+        long minutes = TimeUnit.MILLISECONDS.toMinutes(expectedWindowDelayMillis);
         String requestBody = String
             .format(
                 Locale.ROOT,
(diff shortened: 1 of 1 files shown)


src/test/java/org/opensearch/ad/e2e/DetectionResultEvalutationIT.java
```diff
@@ -563,8 +565,7 @@ public void testValidationWindowDelayRecommendation() throws Exception {
             .extractValue("model", responseMap);
         // adding plus one since window delay always rounds up another minute
         assertEquals(
-            String
-                .format(Locale.ROOT, CommonErrorMessages.WINDOW_DELAY_REC, +recDetectorIntervalMinutes + 1, recDetectorIntervalMinutes + 1),
+            String.format(Locale.ROOT, CommonErrorMessages.WINDOW_DELAY_REC, minutes + 1, minutes + 1),
```
ohltyler, line 568: Just to confirm - this means that before it was hardcoded to a 3 min, and expected an error message suggesting 4 min. And now it's dynamically based on the latest timestamp of the ingested data, which if it takes 1 min to ingest, would be set to 4 min, and expect an error message suggesting 5 min?

amitgalitz: Yes the hardcoded 3 minutes will always be there as its the delay of the first point in the data set and ingestion time is added now like in your example.

kaituo: Could you add comments to explain why you use minutes + 1 ?

amitgalitz: I have comment stating: `// adding plus one since window delay always rounds up another minute` since we round up in window delay recommendation, should I add more to this?

amitgalitz: I'll just change to `(long) Math.ceil(expectedWindowDelayMillis / 60000.0)` which has same effect and is cleaner

kaituo (approved):

src/test/java/org/opensearch/ad/e2e/DetectionResultEvalutationIT.java
kaituo: This is enough. Thanks.

ohltyler (approved):

PR #684, opened 2022-10-07
Create Component Integration

ryanbogan:
Create Component Integration

Signed-off-by: Ryan Bogan <rbogan@amazon.com>

### Description
Integrates createComponents

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #692, opened 2022-10-13
Create Detector Workflow

owaiskazi19:
### Description
<s>This is an initial draft PR for create detector workflow. 
1. When creating an [index](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java#L538) and having the method markMappingUpToDate which requires an actionListener. I have passed it as null for now but we need to figure out a way for this mapping. Or do we even need a listener because we can get the Response directly while creating the AD_Index from SDKClient.
2. Need to wait for createComponent [integration](https://github.com/opensearch-project/anomaly-detection/pull/684) to finish so that we will get an object for AnomalyDetectionIndices to call markMappingUpToDate
3. Not able to start AD Extension because while creating the handler I'm passing null for settings and clusterService
 `handler = List.of(new RestCreateDetectorAction(null, null)); ` as these objects will be available through createComponent </s>

This PR adds:
1. Create Detector workflow for AD Extension
2. Migrate client transports to Apache HttpClient / Core 5.x

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/58

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -14,6 +14,7 @@ import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.sdk.*;
 import org.opensearch.sdk.Extension;
 import org.opensearch.sdk.ExtensionRestHandler;
@@ -45,8 +46,9 @@ public class AnomalyDetectorExtension implements Extension {
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
         List<ExtensionRestHandler> handler = null;
+        Settings settings = Settings.builder().put("deprecated.foo", true).build();
         try {
-            handler = List.of(new RestCreateDetectorAction());
+            handler = List.of(new RestCreateDetectorAction(settings));
         } catch (IOException e) {
             e.printStackTrace();
         }
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -110,7 +110,6 @@ public class AnomalyDetectorExtension implements Extension {
 
     public OpenSearchClient getClient() throws IOException {
         SDKClient sdkClient = new SDKClient();
-        // Need to read this from extensions.yml
         OpenSearchClient client = sdkClient.initializeClient(settings.getHostAddress(), Integer.parseInt(settings.getHostPort()));
         return client;
     }
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -15,6 +15,8 @@ import org.opensearch.sdk.Extension;
 import org.opensearch.sdk.ExtensionRestHandler;
 import org.opensearch.sdk.ExtensionSettings;
 import org.opensearch.sdk.ExtensionsRunner;
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.sdk.*;
 
 import com.google.common.collect.ImmutableList;
 
@@ -39,7 +41,14 @@ public class AnomalyDetectorExtension implements Extension {
 
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
-        return List.of(new RestCreateDetectorAction());
+        List<ExtensionRestHandler> handler = null;
+        // Get the settings anc clusterService from createComponent
+        try {
+            handler = List.of(new RestCreateDetectorAction(null, null));
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return handler;
     }
 
     @Override
@@ -89,6 +98,12 @@ public class AnomalyDetectorExtension implements Extension {
             throw new IOException("Failed to initialize Extension settings. No port bound.");
         }
         return settings;
+    }
+
+    public static OpenSearchClient getClient() throws IOException {
+        SDKClient sdkClient = new SDKClient();
+        OpenSearchClient client = sdkClient.initializeClient("127.0.0.1", 9200);
+        return client;
     }
 
     public static void main(String[] args) throws IOException {
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -45,9 +45,8 @@ public class AnomalyDetectorExtension implements Extension {
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
         List<ExtensionRestHandler> handler = null;
-        // Get the settings anc clusterService from createComponent
         try {
-            handler = List.of(new RestCreateDetectorAction(null, null));
+            handler = List.of(new RestCreateDetectorAction());
         } catch (IOException e) {
             e.printStackTrace();
         }
@@ -110,7 +109,7 @@ public class AnomalyDetectorExtension implements Extension {
 
     public OpenSearchClient getClient() throws IOException {
         SDKClient sdkClient = new SDKClient();
-        OpenSearchClient client = sdkClient.initializeClient(settings.getHostAddress(), Integer.parseInt(settings.getHostPort()));
+        OpenSearchClient client = sdkClient.initializeClient("localhost", 9200);
         return client;
     }
 
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -14,7 +14,6 @@ import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Setting;
-import org.opensearch.common.settings.Settings;
 import org.opensearch.sdk.*;
 import org.opensearch.sdk.Extension;
 import org.opensearch.sdk.ExtensionRestHandler;
@@ -46,9 +45,8 @@ public class AnomalyDetectorExtension implements Extension {
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
         List<ExtensionRestHandler> handler = null;
-        Settings settings = Settings.builder().put("deprecated.foo", true).build();
         try {
-            handler = List.of(new RestCreateDetectorAction(settings));
+            handler = List.of(new RestCreateDetectorAction());
         } catch (IOException e) {
             e.printStackTrace();
         }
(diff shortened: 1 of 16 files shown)


src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -39,7 +44,14 @@ public ExtensionSettings getExtensionSettings() {
 
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
-        return List.of(new RestCreateDetectorAction());
+        List<ExtensionRestHandler> handler = null;
+        // Get the settings anc clusterService from createComponent
+        try {
+            handler = List.of(new RestCreateDetectorAction(null, null));
```
dbwiddis, line 50: First null is `Settings`, second is cluster service.  
Settings should be available via `ExtensionsRunner.sendEnvironmentSettingsRequest()`.  

BaseExtension createComponents already stores cluster state, just need to actually call that from the ExtensionsRunner at some point, and also add settings to that.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -39,7 +44,14 @@ public ExtensionSettings getExtensionSettings() {
...
+        } catch (IOException e) {
```
dbwiddis, line 51: does instantiating this object throw an exception?

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -83,6 +95,11 @@ public List<Setting<?>> getSettings() {
         );
     }
 
+    @Override
+    public Collection<Object> createComponents(SDKClient sdkClient, ClusterService clusterService, ThreadPool threadPool) {
+        return null;
```
dbwiddis, line 100: Not sure this is ever read anywhere, but isn't returning an empty collection better than null?

We don't need to override this if we implement in BaseExtension.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -3,18 +3,23 @@
...
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.opensearch.ad.rest.RestCreateDetectorAction;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.settings.EnabledSetting;
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Setting;
+import org.opensearch.sdk.*;
```
dbwiddis, line 17: Given how much confusion I know was generated in creating this PR with similarly-named classes, this star import should be a big red flag. :-)

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -39,7 +44,13 @@ public ExtensionSettings getExtensionSettings() {
 
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
-        return List.of(new RestCreateDetectorAction());
+        List<ExtensionRestHandler> handler = null;
+        try {
+            handler = List.of(new RestCreateDetectorAction());
```
dbwiddis, line 49: We don't need a catch block for this object instantiation.  The constructor is a no-op.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -91,6 +107,12 @@ private static ExtensionSettings initializeSettings() throws IOException {
         return settings;
     }
 
+    public OpenSearchClient getClient() throws IOException {
+        SDKClient sdkClient = new SDKClient();
+        OpenSearchClient client = sdkClient.initializeClient("localhost", 9200);
```
dbwiddis, line 112: Let's see if we can fetch these parameters from somewhere else.

src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
```diff
@@ -1,134 +1,241 @@
...
-import static org.opensearch.rest.RestRequest.Method.GET;
-import static org.opensearch.rest.RestRequest.Method.PUT;
-import static org.opensearch.rest.RestStatus.BAD_REQUEST;
-import static org.opensearch.rest.RestStatus.NOT_ACCEPTABLE;
-import static org.opensearch.rest.RestStatus.NOT_FOUND;
-import static org.opensearch.rest.RestStatus.NOT_MODIFIED;
+import static org.opensearch.ad.model.AnomalyDetector.ANOMALY_DETECTORS_INDEX;
+import static org.opensearch.ad.settings.AnomalyDetectorSettings.ANOMALY_DETECTORS_INDEX_MAPPING_FILE;
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.rest.RestRequest.Method.*;
```
dbwiddis, line 6: I love static imports but not ones with *.

src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
```diff
@@ -1,134 +1,241 @@
...
-    @Override
-    public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
-        Method method = request.method();
+    private String getAnomalyDetectorMappings() throws IOException {
+        URL url = AnomalyDetectionIndices.class.getClassLoader().getResource(ANOMALY_DETECTORS_INDEX_MAPPING_FILE);
+        return Resources.toString(url, Charsets.UTF_8);
+    }
+
+    private IndexResponse indexAnomalyDetector(AnomalyDetector anomalyDetector) throws IOException {
+        AnomalyDetector detector = new AnomalyDetector(
```
dbwiddis, line 77: It looks here like we're pretty much cloning the input parameter.  Why is this needed?

src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
```diff
@@ -1,134 +1,241 @@
...
+    @Override
+    public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
+        if (!EnabledSetting.isADPluginEnabled()) {
+            throw new IllegalStateException(CommonErrorMessages.DISABLED_ERR_MSG);
+        }
+        Method method = request.method();
+
+        NamedXContentRegistry xContentRegistry = new NamedXContentRegistry(getNamedXWriteables());
+        XContentParser parser = null;
+        try {
```
dbwiddis, line 191: I think we can try to wrap most of this handler method in a try/catch, where the catch returns an ExtensionRestResponse failure with the exception details.

src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
owaiskazi19: IndexRequest requires a model class object to document it with the index.
```
IndexRequest<AnomalyDetector> indexRequest = new IndexRequest.Builder<AnomalyDetector>()
            .index(ANOMALY_DETECTORS_INDEX)
            .id("1")
            .document(detector)
            .build();

File: src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java (hunk 6 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java:106]: Raised https://github.com/opensearch-project/opensearch-sdk-java/pull/207 for the same

File: src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java (hunk 1 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java]: No need of settings and clusterService anymore. Removed it :)

File: src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java (hunk 2 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java]: Removed the exception from SDK: https://github.com/opensearch-project/opensearch-sdk-java/pull/234

File: src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java (hunk 3 above)
Comment (owaiskazi19) [reply src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java:93]: Not extending from `AbstractAnomalyDetectorAction` anymore. So not needed, removed!

File: src/main/java/org/opensearch/ad/model/AnomalyDetector.java [hunk 10]
```diff
@@ -65,15 +65,14 @@
...
-
     public static final String PARSE_FIELD_NAME = "AnomalyDetector";
     public static final NamedXContentRegistry.Entry XCONTENT_REGISTRY = new NamedXContentRegistry.Entry(
         AnomalyDetector.class,
         new ParseField(PARSE_FIELD_NAME),
         it -> parse(it)
     );
     public static final String NO_ID = "";
-    public static final String ANOMALY_DETECTORS_INDEX = ".opendistro-anomaly-detectors";
+    public static final String ANOMALY_DETECTORS_INDEX = "opendistro-anomaly-detectors";
```
dbwiddis, line 75: (Random thought, don't hold up PR approval for this)

Do we want to keep the opendistro name for backwards compatibility?  Although removing the dot (per 3.x) breaks that anyway, so we might want a new name.

src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
```diff
@@ -1,134 +1,227 @@
...
-        String worldNameWithRandomAdjective = worldAdjectives.isEmpty()
-            ? worldName
-            : String.join(" ", worldAdjectives.get(rand.nextInt(worldAdjectives.size())), worldName);
-        return new ExtensionRestResponse(request, OK, String.format(GREETING, worldNameWithRandomAdjective));
+    public List<NamedXContentRegistry.Entry> getNamedXWriteables() {
+        List<NamedXContentRegistry.Entry> entries = new ArrayList<>();
+        entries.add(AnomalyDetector.XCONTENT_REGISTRY);
+        entries.add(AnomalyResult.XCONTENT_REGISTRY);
+        entries.add(DetectorInternalState.XCONTENT_REGISTRY);
+        entries
```
dbwiddis, line 110: Note to other reviewers: these four `entries.add()` here and below are placeholder code and will be replaced by getNamedXContent work in progress.

This comment may be resolved when ready to merge.

dbwiddis (approved):
LGTM! Great job! A few comments (can be resolved to merge).

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
owaiskazi19: This is really good point. Do you think we should discuss this on an issue on SDK repo or if you have a better name I can replace it in this PR itself?
I can think of `opensearch-anomaly-detector`

owaiskazi19: Actually, let me open an issue for this because I think it will require some discussion

owaiskazi19: https://github.com/opensearch-project/opensearch-sdk-java/issues/241

ryanbogan (approved):

PR #696, opened 2022-10-19
Fixing issues caused by 3.0 version bumping

jackiehanyang:
### Description
- Fix issues caused by upstream breaking change - https://github.com/opensearch-project/OpenSearch/commit/4833e080cc506237fee5aa90767d6f49d952bec3#
- Use version 2.4 for bwc test

### Issues Resolved
https://github.com/opensearch-project/opensearch-plugins/issues/142


build.gradle
```diff
@@ -136,7 +136,7 @@ configurations.all {
         force "joda-time:joda-time:${versions.joda}"
         force "com.fasterxml.jackson.core:jackson-core:2.13.4"
         force "commons-logging:commons-logging:${versions.commonslogging}"
-        force "org.apache.httpcomponents:httpcore:${versions.httpcore}"
+        force "org.apache.httpcomponents:httpcore5:${versions.httpcore5}"
```
amitgalitz, line 139: Could we possible just remove this `force`? and just inherit it as is from opensearch core?

build.gradle
```diff
@@ -372,7 +372,7 @@ task integTestRemote(type: RestIntegTestTask) {
     }
 }
 
-String bwcVersion = "1.1.0.0"
+String bwcVersion = "2.4.0.0"
```
amitgalitz, line 375: nit: We can maybe set this at the top of constants and also use it for fetching correct version in line 41/43. just to minimize number of place we are changing version number on each increment in main?

ohltyler:
Can you link the related changes in core here for more context? For example, not sure of the reasoning of the new apache dependencies

build.gradle
jackiehanyang: do you know what's the initial reason of adding this `force` instead of inheriting it from opensearch core?

amitgalitz: Not sure, seems to be with our initial set up of AD. Usually from my understanding its used if we need a specific version of a dependency that doesn't align with core so we force it to be something else. I believe though since we aren't in need of anything specific we should be able to remove this line and build/testing should be fine.

ohltyler: Before doing that I would confirm with @ylwu-amzn or @kaituo on the reasoning. When that's figured out, if it's still needed, we should add a clarifying comment above.

jackiehanyang:
> Can you link the related changes in core here for more context? For example, not sure of the reasoning of the new apache dependencies

Updated the description.

build.gradle
jackiehanyang: > nit: We can maybe set this at the top of constants and also use it for fetching correct version in line 41/43. just to minimize number of place we are changing version number on each increment in main?

moved bwc related constants to the build script block

amitgalitz (approved):

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
```diff
@@ -62,7 +63,7 @@ public ToXContentObject[] getHistoricalAnomalyDetector(String detectorId, boolea
         return getAnomalyDetector(detectorId, header, false, returnTask, client);
     }
 
-    public ADTaskProfile getADTaskProfile(String detectorId) throws IOException {
+    public ADTaskProfile getADTaskProfile(String detectorId) throws IOException, ParseException {
```
ohltyler, line 66: What changed where this can throw a `ParseException` now?

src/test/java/org/opensearch/ad/ODFERestTestCase.java
```diff
@@ -139,7 +143,7 @@ protected RestClient buildClient(Settings settings, HttpHost[] hosts) throws IOE
     @After
     protected void wipeAllODFEIndices() throws IOException {
         Response response = adminClient().performRequest(new Request("GET", "/_cat/indices?format=json&expand_wildcards=all"));
-        XContentType xContentType = XContentType.fromMediaType(response.getEntity().getContentType().getValue());
+        XContentType xContentType = XContentType.fromMediaType(response.getEntity().getContentType());
```
ohltyler, line 146: Was this due to a breaking change in `XContentType`?

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
jackiehanyang: the toString() method in `EntityUtils.toString(profileResponse.getEntity());`

src/test/java/org/opensearch/ad/ODFERestTestCase.java
jackiehanyang: yes

src/test/java/org/opensearch/ad/HistoricalAnalysisRestTestCase.java
ohltyler: Was this due to a dependency change? Can you link the method that has changed? I don't see anything explicitly changed here

jackiehanyang:
```
    /**
     * Reads the contents of an entity and return it as a String.
     * The content is converted using the character set from the entity (if any),
     * failing that, "ISO-8859-1" is used.
     *
     * @param entity the entity to convert to a string; must not be null
     * @return String containing the content.
     * @throws ParseException if header elements cannot be parsed
     * @throws IllegalArgumentException if entity is null or if content length &gt; Integer.MAX_VALUE
     * @throws IOException if an error occurs reading the input stream
     * @throws java.nio.charset.UnsupportedCharsetException Thrown when the named charset is not available in
     * this instance of the Java virtual machine
     */
    public static String toString(final HttpEntity entity) throws IOException, ParseException {
        return toString(entity, DEFAULT_ENTITY_RETURN_MAX_LENGTH);
    }
```

ohltyler: Is this from one of the apache dependency changes? Sorry for the confusion just wondering where this changed exactly.

jackiehanyang: yes, it's from the apache dependency change. You can find the import of this exception is `org.apache.hc.core5.http.ParseException`

ohltyler: I was just wondering where that method was coming from - I now see it's from `EntityUtils` which is a dependency

ohltyler (approved):

PR #703, opened 2022-10-27
windows CI for AD

### LINKED ISSUE #274: Add Windows Support
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/95, add Windows support.

- [ ] Passing CI on Windows
- [ ] Documentation

amitgalitz:
### Description
Adds windows CI for main branch

1. Changed the usage of getPath() to first use .toURI() in some of the test cases. This is because just getting the path like we used to was returning a string like this:` \C:\Sample\sample.txt`. This isn't supported on Windows. Converting with `toUri()` works on all platforms.
### Issues Resolved
resolves #274 

### DIFF
--- a/src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
+++ b/src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
@@ -1045,6 +1045,7 @@ public class CheckpointDaoTests extends OpenSearchTestCase {
         // Model in file rc1_trcf_model_direct is a checkpoint creatd by RCF-3.0-rc1
         URI uri = ClassLoader.getSystemResource("org/opensearch/ad/ml/rc1_trcf_model_direct.json").toURI();
         String filePath = Paths.get(uri).toString();
+        String json = Files.readString(Paths.get(filePath), Charset.defaultCharset());
         // For the parsing of .toTrcf to work I had to manually change "\u003d" in code back to =.
         // In the byte array it doesn't seem like this is an issue but whenever reading the byte array response into a file it
         // converts "=" to "\u003d" https://groups.google.com/g/google-gson/c/JDHUo9DWyyM?pli=1
--- a/src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
+++ b/src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
@@ -32,6 +32,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


amitgalitz:
This is currently failing because a change was made in main to now allow Opensearch to run as root user. This was reverted an hour with this PR: https://github.com/opensearch-project/OpenSearch/pull/4949. Will try running CI again once validated this change is in latest snapshot

amitgalitz:
@ohltyler @jackiehanyang  Lets merge this into main and then merge following backport PR. I will close the other windows-ci PR when this is merged and passing CI.

ohltyler:
Nice! LGTM. Just recommend changing the PR description from `2.x` -> `main`

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
```diff
@@ -1010,7 +1011,8 @@ private double[] getPoint(int dimensions, Random random) {
     // The checkpoint used for this test is from a single-stream detector
     public void testDeserializeRCFModelPreINIT() throws Exception {
         // Model in file 1_3_0_rcf_model_pre_init.json not passed initialization yet
-        String filePath = getClass().getResource("1_3_0_rcf_model_pre_init.json").getPath();
```
ylwu-amzn, line 1013: This code won't work on windows?

ohltyler:
Also, if you add `closes` or `resolves` keyword in front of the issue number it will automatically link to the PR

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
ohltyler: See PR description

amitgalitz: Yeah it doesn't, the reason is because `.getPath()` adds a `/` to beginning of path which is fine for linux but not for windows. It causes `java.nio.file.InvalidPathException: Illegal char <:> at index 2`

ylwu-amzn: I'm not sure, as ml-commons also using `getClass().getResource` and it passed windows Github CI , check code https://github.com/opensearch-project/ml-commons/blob/2.x/ml-algorithms/src/test/java/org/opensearch/ml/engine/algorithms/text_embedding/TextEmbeddingModelTest.java#L90

ylwu-amzn: I think it's fine to use either way. Just curious why AD can't work.

amitgalitz: ML-Commons uses `toUri()` not `getPath(). Thats what I changed to using. `toUri()` is platform independent

ohltyler (approved):

ylwu-amzn (approved):

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
ylwu-amzn: oh, got it, I should read it more carefully. Thanks, that make sense.

PR #707, opened 2022-10-27
Fix Docker CI for security enabled tests

amitgalitz:
### Description
_Describe what this change achieves._

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #710, opened 2022-10-28
Fixing Docker CI for security enabled tests 

### LINKED ISSUE #275: Add MacOS Support
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/96, add MacOS (Darwin) support.

- [ ] Passing CI on MacOS
- [ ] Documentation

amitgalitz:
### Description
Split our CI into 3 different workflows. 
1. CI.yml -> This includes building AD, which includes UT and IT plus running multi node integration tests (linux, macos, windows). Windows is seperate here because I was not able to run it correctly with the `-Dtest.logs=true` option, only on linux and macos, I believe its okay for now as we only need that part for precision and recall and its enough to just see it on those two platforms.

2. Security-Test-CI.yml -> Assembles AD and sets up docker container to run AD integ tests with security plugin if its available. (Linux only cause no infra set up yet for windows and macos)
3. Bwc-test-CI.yml -> Runs BWC separately for AD (Linux only cause we don't have an older version with windows and macos)

Additionally it might be hard to notice cause of the new files but I changed some of the parsing logic in security-test-ci as it was failing to parse the correct docker version before. 
### Issues Resolved
resolves #275 


amitgalitz:
Security tests are failing here, I think this is a 3.0 specific problem as the same tests passed on this PR for 2.x: https://github.com/opensearch-project/anomaly-detection/pull/707
The errors are also related to http client which had some changes in 3.0 branch. 
If rest of checks pass I believe we should still merge this in, backport the change and open an issue on security to further investigate errors.

ohltyler:
Can we rename the workflows to follow a consistent convention? I think all underscore (verb-first) like mentioned [here](https://nimblehq.co/compass/development/code-conventions/github-actions/#naming) is a good way, or all-`-`. I see our workflows are already a bit inconsistent

.github/workflows/Security-Test-CI.yml
```diff
@@ -0,0 +1,81 @@
...
+        if: env.imagePresent == 'true'
+        run: |
+          security=`curl -XGET https://localhost:9200/_cat/plugins?v -u admin:admin --insecure |grep opensearch-security|wc -l`
+          if [ $security -gt 0 ]
+          then
+            echo "Security plugin is available"
+            ./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=true -Duser=admin -Dpassword=admin
+          else
+            echo "Security plugin is NOT available"
+            ./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster"
```
ohltyler, line 80: Can we remove running this if security plugin isn't available? I think it's fine to leave the echo statement so it doesn't fail, but no necessary to run w/o security since it's covered in regular CI

amitgalitz: removed

ohltyler (approved):

amitgalitz:
More context on security failures here: https://github.com/opensearch-project/anomaly-detection/issues/712

kaituo (approved):

PR #714, opened 2022-11-04
remove bwc code related to 1.0 and earlier version

kaituo:
### Description
This PR addresses AD compile failure because the core removed 1x Version constants. (check https://github.com/opensearch-project/OpenSearch/pull/5021) OpenSearch does not support N-2 version compatibility. This is inherited from Elasticsearch and Lucene. So version 1x is not compatible with 3.0. Thus removal of deprecated 1x code. This PR follows suit and removes backward compatibility code on OpenSearch 1.0 and older versions. So we won't support direct upgrade from 1.x domains to 3.x.

Testing done:
1. gradle build. Note that CI workflow will fail due to missing job scheduler. We are using a job scheduler from distribution. But due to a circular dependency on -SNAPSHOT builds being published to maven that require the distribution build to be successful (check https://github.com/opensearch-project/opensearch-build/issues/1463), AD compilation failure caused the job scheduler to be taken out.  Not including the latest job scheduler will cause AD build to fail. So we have a chicken and egg problem: this PR fixes AD build failure and the PR itself cannot build due to missing job scheduler. To run gradle build, I changed to use local job scheduler and verified gradle build succeeded. Once I merge this PR. job scheduler should be able to build and be published to maven. Future AD CI should be unblocked after that.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/713


src/main/java/org/opensearch/ad/util/Bwc.java
```diff
@@ -1,32 +0,0 @@
-/*
```
ohltyler, line 1: just wondering, do you think it's worth keeping this file around just as skeleton code? I imagine we'll need to re-create something like this in the future for other breaking changes that require conditional testing based on the version

src/test/java/org/opensearch/StreamInputOutputTests.java
```diff
@@ -0,0 +1,293 @@
...
+import org.opensearch.cluster.node.DiscoveryNode;
+import org.opensearch.common.io.stream.BytesStreamOutput;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.transport.TransportAddress;
+
+/**
+ * Put in core package so that we can using Version's package private constructor
+ *
+ */
+public class StreamInputOutputTests extends AbstractADTest {
```
ohltyler, line 51: is the main idea here that it's the same as [bwc](src/main/java/org/opensearch/ad/util/Bwc.java), but removes the tests and conditional statements about 1.x requests/responses? Anything in particular that's been added in this file? It's a lot of lines to try to compare

src/main/java/org/opensearch/ad/util/Bwc.java
kaituo: We can refer to older versions or this PR to bring it back. Having a skeleton code there will make the coverage 0 and will fail our coverage check tool.

src/test/java/org/opensearch/StreamInputOutputTests.java
kaituo: yes, it is almost the same except I added a few more lines of code to boost the test coverage. You can deem it the same.

src/main/java/org/opensearch/ad/util/Bwc.java
ohltyler: Fair enough, I'm ok to remove

ohltyler (approved):
LGTM!

src/test/java/org/opensearch/BwcTests.java
```diff
@@ -1,564 +0,0 @@
-/*
```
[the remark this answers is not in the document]
kaituo: We have other BWC tests related to model (e.g., rcf 1.0 upgrading to 2.0).

ylwu-amzn (approved):

PR #600, opened 2022-06-28
Expose model accuracy metrics in tests

kaituo:
### Description
This PR adds an option flag to print logs during tests and turn on the flag in CI workflow. The flag is disabled by default. By doing this, we can record model accuracy metrics in git workflows and later retrieve it for analysis.

Testing done:
1. We can turn on/off logs during tests.
2. The accuracy logs are recorded.

Signed-off-by: Kaituo Li <kaituo@amazon.com>

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #716, opened 2022-11-07

kaituo:
### Description
This PR adds a AD model performance benchmark so that we can compare model performance across versions.

We also backported https://github.com/opensearch-project/anomaly-detection/pull/600 so that we can capture the performance data in CI output.

Testing done:
1. added unit tests to run the benchmark.


kaituo:
Cannot get CI to run as OpenSearch 1.0.0 is not actively maintained:

 Execution failed for task ':buildSrc:compileJava'.
> Could not resolve all files for configuration ':buildSrc:compileClasspath'.
   > Could not download nebula-publishing-plugin-4.4.4.jar (com.netflix.nebula:nebula-publishing-plugin:4.4.4)
      > Could not get resource 'https://jcenter.bintray.com/com/netflix/nebula/nebula-publishing-plugin/4.4.4/nebula-publishing-plugin-4.4.4.jar'.
         > Could not GET 'https://jcenter.bintray.com/com/netflix/nebula/nebula-publishing-plugin/4.4.4/nebula-publishing-plugin-4.4.4.jar'.
            > Read timed out

kaituo:
Run gradle build locally as I have cached 1.0.0 OpenSearch build before:

./gradlew build -Dtest.logs=true
...
        at org.apache.lucene.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:64) [lucene-test-framework-8.8.2.jar:8.8.2 a92a05e195b775b30ca410bc0a26e8e79e7b3bfb - mdrob - 2021-04-06 16:33:29]
        at org.apache.lucene.util.TestRuleIgnoreTestSuites$1.evaluate(TestRuleIgnoreTestSuites.java:54) [lucene-test-framework-8.8.2.jar:8.8.2 a92a05e195b775b30ca410bc0a26e8e79e7b3bfb - mdrob - 2021-04-06 16:33:29]
        at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36) [randomizedtesting-runner-2.7.1.jar:?]
        at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:368) [randomizedtesting-runner-2.7.1.jar:?]
        at java.lang.Thread.run(Thread.java:832) [?:?]
    [2022-11-08T00:46:12,571][INFO ][o.o.a.AbstractADTest     ] [testFail] Cannot delete old checkpoint.
    [2022-11-08T00:46:12,571][INFO ][o.o.a.c.DailyCronTests   ] [testFail] after test

org.opensearch.ad.cluster.DailyCronTests > testNormal STANDARD_OUT
    [2022-11-08T00:46:12,574][INFO ][o.o.a.c.DailyCronTests   ] [testNormal] before test
    [2022-11-08T00:46:12,595][INFO ][o.o.a.c.DailyCron        ] [testNormal] 10 checkpoint docs get deleted
    [2022-11-08T00:46:12,595][INFO ][o.o.a.AbstractADTest     ] [testNormal] 10 checkpoint docs get deleted
    [2022-11-08T00:46:12,595][INFO ][o.o.a.c.DailyCronTests   ] [testNormal] after test

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 4m 41s
28 actionable tasks: 28 executed
(base)
(22-11-07 14:46:16) <0> [~/code/github/opensearch-ad]
dev-dsk-kaituo-2b-bf84c4db %

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
```diff
@@ -268,15 +324,19 @@ private void trainModelFromDataSegments(List<double[][]> dataPoints, String mode
...
             .dimensions(rcfNumFeatures)
             .sampleSize(rcfSampleSize)
             .numberOfTrees(numberOfTrees)
             .lambda(rcfTimeDecay)
             .outputAfter(numMinSamples)
-            .parallelExecutionEnabled(false)
-            .build();
+            .parallelExecutionEnabled(false);
+
+        if (rcfSeed != -1) {
```
amitgalitz, line 336: We are simply initializing it in AD so we have access to control what the seed is in our testing right? No behavioral change cause RCF would initialize a seed if not given one by AD?

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
```diff
@@ -458,4 +509,258 @@ public void testEmptyDataRange() throws InterruptedException {
...
+                List<Entry<Long, Long>> ranges = invocation.getArgument(1);
+                List<Optional<double[]>> coldStartSamples = new ArrayList<>();
+
+                Collections.sort(ranges, new Comparator<Entry<Long, Long>>() {
+                    @Override
+                    public int compare(Entry<Long, Long> p1, Entry<Long, Long> p2) {
+                        return Long.compare(p1.getKey(), p2.getKey());
+                    }
+                });
+                for (int j = 0; j < ranges.size(); j++) {
```
amitgalitz, line 658: Do the ranges we get here have actual values to allow us to mock the actual expected results by mapping them to the timestamps?

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
```diff
@@ -458,4 +509,258 @@ public void testEmptyDataRange() throws InterruptedException {
...
+                    int valueIndex = searchInsert(timestamps, start);
+                    coldStartSamples.add(Optional.of(data[valueIndex]));
+                }
+
+                ActionListener<List<Optional<double[]>>> listener = invocation.getArgument(4);
+                listener.onResponse(coldStartSamples);
+                return null;
+            }).when(searchFeatureDao).getColdStartSamplesForPeriods(any(), any(), any(), anyBoolean(), any());
+
+            String entityNamez = entityName + z;
```
amitgalitz, line 670: I'm a little confused about this part, are we creating a new Entity model that doesn't exist with name "abcz" and then training the model with that one entity so we treat all the data points only for that one entity as we are only training one model? basically simulating HCAD logic with only one entity? Does that mean the categories given earlier when creating the detector and going through cold start don't matter?

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
```diff
@@ -458,4 +509,258 @@ public void testEmptyDataRange() throws InterruptedException {
...
+                ModelType.ENTITY.getName(),
+                clock,
+                priority
+            );
+
+            released = new AtomicBoolean();
+
+            inProgressLatch = new CountDownLatch(1);
+            listener = ActionListener.wrap(() -> {
+                released.set(true);
```
amitgalitz, line 686: I don't think we need this listener, atomicBoolean and CountDownLatch. In current code our `trainModel()` utilizes a listener to know if to pull another request from cold start queue which I assume is why we could have used a latchCountDown to release threads being used by the listener in this loop but I don't see it in 1.0 code.

src/test/java/org/opensearch/ad/util/LabelledAnomalyGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+                    if (noiseprg.nextDouble() < 0.01 && noiseprg.nextDouble() < 0.3) {
+                        double factor = 5 * (1 + noiseprg.nextDouble());
+                        double change = noiseprg.nextDouble() < 0.5 ? factor * noise : -factor * noise;
+                        data[i][j] += newChange[j] = change;
+                        changedTimestamps[i] = timestamps[i];
+                        changes[i] = newChange;
+                    }
+                }
+            } else {
+                boolean flag = (noiseprg.nextDouble() < 0.01);
```
amitgalitz, line 97: For the case where anomalies are dependent on each other, should we not by default just move both the 0.01 and 0.3 check out? Are we doing it this way so there is still some variance if all features have an anomaly or not.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
kaituo: yeah, no behavior change will happen as by default we don't provide a seed.

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
kaituo: yes

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
kaituo: yeah, what you described is correct. I am simulate HCAD logic with only one entity. It takes a long time run all of these steps for one entity (1.0 is relatively fast, 1.1 takes 25+ minutes since we disabled bounding box caching). Yeah, the categories given earlier does not matter.

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
kaituo: you are right. Removed those 3 variables.

src/test/java/org/opensearch/ad/util/LabelledAnomalyGenerator.java
kaituo: yeah, we are doing this way so there is still some variance. So not all features have anomalies.

src/test/java/org/opensearch/ad/ml/EntityColdStarterTests.java
amitgalitz: was the reason we needed the other anomaly detector and entity just to get samples for that period?

kaituo: We need the detector object as EntityColdStarter.getEntityColdStartData requires a full detector and then access its properties.

amitgalitz (approved):

kaituo:
Ran gradle build on the new commit. Build Successful.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,79 @@
...
+          echo "Copying ./build/distributions/*.zip to ../src/test/resources/job-scheduler ..."
+          ls ./build/distributions/
+          cp ./build/distributions/*.zip ../src/test/resources/job-scheduler
+          echo "Copied ./build/distributions/*.zip to ../src/test/resources/job-scheduler ..."
+          ls ../src/test/resources/job-scheduler
+
+      - name: Build and Run Tests
+        run: |
+          ./gradlew ':test' --tests "org.opensearch.ad.ml.HCADModelPerfTests" -Dtests.seed=2AEBDBBAE75AC5E0 -Dtests.security.manager=false -Dtests.locale=es-CU -Dtests.timezone=Chile/EasterIsland -Dtest.logs=true -Dhcad-benchmark=true
+          ./gradlew integTest --tests "org.opensearch.ad.e2e.DetectionResultEvalutationIT.testDataset" -Dtests.seed=60CDDB34427ACD0C -Dtests.security.manager=false -Dtests.locale=kab-DZ -Dtests.timezone=Asia/Hebron -Dtest.logs=true
```
amitgalitz, line 79: wont these tests just run twice now, as we run integTest in CI.yml workflow? Also are we trying to keep the same seed across all versions?

kaituo: HCADModelPerfTests is only ran in standalone mode. I excluded them in gradle build run. DetectionResultEvalutationIT.testDataset is ran twice. I thought about remove DetectionResultEvalutationIT.testDataset from gradle build, but I feel it is a good signal to check if single stream real time detectors work from end to end. So I kept them. Yes, the seed across versions are the same.

amitgalitz: Will it be possible to inject seed without changing the API? We don't have access too it then.

kaituo: right now it is not easy to do it without changing the API.

src/main/java/org/opensearch/ad/ml/EntityColdStarter.java
amitgalitz: in 1.1 PR you set the seed too 2051L for the other tests that will use this setup. I assume it doesn't matter for the HCADModelPerfTests as we re-establish seed to be loop number. Should we add the same seed tho in the setup in the AbstractCosineDataTest class here?

kaituo: you meant to use loop number in the setup? The 2051L number will be used in other tests in EntityColdStarterTests. HCADModelPerfTests create a different EntityColdStarter and doesn't need the seed number 2051L.

src/test/java/org/opensearch/ad/util/LabelledAnomalyGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Regarding benchmark data, we randomly generated synthetic data with known anomalies
+ * inserted throughout the signal. In particular, these are one/two/four dimensional
+ * data where each dimension is a noisy cosine wave. Anomalies are inserted into one
+ * dimension with 0.003 probability. Anomalies across each dimension can be independent
+ * or dependent.  We have approximately 5000 observations per data set. The data set is
+ * generated using the same random seed so the result is comparable across versions.
+ *
+ */
+public class LabelledAnomalyGenerator {
```
ohltyler, line 29: This is nice to have. I wonder if we could expose this on the UI somewhere to let users play around with some data and test out detectors for themselves.

ohltyler (approved):

src/test/java/org/opensearch/ad/util/LabelledAnomalyGenerator.java
ohltyler: Also, what was the reasoning for only testing 1/2/4 dimensional data? Do higher dimensions bring too much variability/noise regarding precision/recall results?

kaituo: yeah, this could be part of HCAD sample detectors.

kaituo: By default, we support up to 5 features, even though we can increase to 10. Testing 1/2/4 dimensional data would cover power of 2 dimensions less than 5. Yeah, we can do all 5 of them. I assume 1/2/4 can give us a glimpse of the whole picture. Yes, high dimensions brings curse of dimensionality (https://en.wikipedia.org/wiki/Curse_of_dimensionality).

kaituo: When the number of features increase to 4, the precision/recall is lesser. Take 2.4 as an example:

| Precision      | Recall |  Number of features |
| ----------- | ----------- |
| 0.5697 | 0.4548  | 4 |
| 0.62214 | 0.64622| 2 |
| 0.62944 | 0.58078 | 1 |

There is only 1 unique entity involved. You can see I only have one entity in HCADModelPerfTests.averageAccuracyTemplate.

```
entity = Entity.createSingleAttributeEntity("field", entityName + z);
```

Issue #719, opened 2022-11-11
[BUG]Build failure for 1.3 branch


b4sjoo:
**What is the bug?**
Build failure for 1.3 branch due to 3 failed tests:

SearchFeatureDaoTests in org.opensearch.ad.feature.SearchFeatureDaoTests
ModelManagerTests in org.opensearch.ad.ml.ModelManagerTests
GetAnomalyDetectorActionTests in org.opensearch.ad.transport.GetAnomalyDetectorActionTests

**How can one reproduce the bug?**
`./gradlew build`

**What is the expected behavior?**
```
SearchFeatureDaoTests initializationError
java.lang.NoClassDefFoundError: javassist/NotFoundException
	at org.powermock.core.transformers.TestClassTransformerBuilder$RemovesTestMethodAnnotation.fromMethods(TestClassTransformerBuilder.java:62)
	at org.powermock.tests.utils.impl.AbstractCommonTestSuiteChunkerImpl.createDefaultMockLoader(AbstractCommonTestSuiteChunkerImpl.java:126)
	at org.powermock.tests.utils.impl.AbstractCommonTestSuiteChunkerImpl.chunkClass(AbstractCommonTestSuiteChunkerImpl.java:89)
	at org.powermock.tests.utils.impl.AbstractCommonTestSuiteChunkerImpl.<init>(AbstractCommonTestSuiteChunkerImpl.java:50)
	at org.powermock.tests.utils.impl.AbstractCommonTestSuiteChunkerImpl.<init>(AbstractCommonTestSuiteChunkerImpl.java:44)
	at org.powermock.tests.utils.impl.AbstractTestSuiteChunkerImpl.<init>(AbstractTestSuiteChunkerImpl.java:57)
	at org.powermock.modules.junit4.common.internal.impl.JUnit4TestSuiteChunkerImpl.<init>(JUnit4TestSuiteChunkerImpl.java:57)
	at org.powermock.modules.junit4.common.internal.impl.AbstractCommonPowerMockRunner.<init>(AbstractCommonPowerMockRunner.java:36)
	at org.powermock.modules.junit4.PowerMockRunner.<init>(PowerMockRunner.java:34)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
[...]

PR #722, opened 2022-11-11

kaituo:
### Description

We also backported #600 so that we can capture the performance data in CI output.

Testing done:

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/719

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,10 @@ buildscript {
                 'opensearch-anomaly-detection-1.1.0.0.zip'
         bwcOpenSearchJSDownload = 'https://ci.opensearch.org/ci/dbc/bundle-build/1.1.0/20210930/linux/x64/builds/opensearch/plugins/' +
                 'opensearch-job-scheduler-1.1.0.0.zip'
+        // gradle build won't print logs during test by default unless there is a failure.
+        // It is useful to record intermediately information like prediction precision and recall.
+        // This option turn on log printing during tests. 
+        printLogs = "true" == System.getProperty("test.logs", "false")
     }
 
     repositories {
@@ -175,6 +179,12 @@ test {
     }
     include '**/*Tests.class'
     systemProperty 'tests.security.manager', 'false'
+
+    if (System.getProperty("model-benchmark") == null || System.getProperty("model-benchmark") == "false") {
+        filter {
+            excludeTestsMatching "org.opensearch.ad.ml.HCADModelPerfTests"
+        }
+    }
 }
 
 task integTest(type: RestIntegTestTask) {
@@ -220,6 +230,12 @@ integTest {
         }
     }
 
+    if (System.getProperty("model-benchmark") == null || System.getProperty("model-benchmark") == "false") {
+        filter {
+            excludeTestsMatching "org.opensearch.ad.e2e.SingleStreamModelPerfIT"
+        }
+    }
+
     // The 'doFirst' delays till execution time.
     doFirst {
         // Tell the test JVM if the cluster JVM is running under a debugger so that tests can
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -670,6 +692,7 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.7.2'
     testRuntimeOnly 'org.junit.vintage:junit-vintage-engine:5.7.2'
     testCompileOnly 'junit:junit:4.13.2'
+    implementation group: 'org.javassist', name: 'javassist', version:'3.28.0-GA'
```
amitgalitz, line 695: Did you figure out why we suddenly needed this?

src/test/java/org/opensearch/ad/e2e/AbstractSyntheticDataTest.java
```diff
@@ -0,0 +1,258 @@
...
+public class AbstractSyntheticDataTest extends ODFERestTestCase {
+
+    /**
+     * In real time AD, we mute a node for a detector if that node keeps returning
+     * ResourceNotFoundException (5 times in a row).  This is a problem for batch mode
+     * testing as we issue a large amount of requests quickly. Due to the speed, we
+     * won't be able to finish cold start before the ResourceNotFoundException mutes
+     * a node.  Since our test case has only one node, there is no other nodes to fall
+     * back on.  Here we disable such fault tolerance by setting max retries before
+     * muting to a large number and the actual wait time during muting to 0.
```
ohltyler, line 55: What's the requests being made that's causing the exceptions?

ohltyler: If it's the data ingestion, how much data loss is occurring?

src/test/java/org/opensearch/ad/e2e/AbstractSyntheticDataTest.java
```diff
@@ -0,0 +1,258 @@
...
+        String categoryField,
+        long windowDelayInMins
+    ) throws Exception {
+        Request request = new Request("POST", "/_plugins/_anomaly_detection/detectors/");
+        String requestBody = null;
+        if (Strings.isEmpty(categoryField)) {
+            requestBody = String
+                .format(
+                    Locale.ROOT,
+                    "{ \"name\": \"test\", \"description\": \"test\", \"time_field\": \"timestamp\""
```
ohltyler, line 175: I know these long JSON strings exist a lot in the tests, but wondering if you think it's reasonable to pull these out into JSON objs in constants somewhere, and pulling them in and converting to strings here. Could remove duplicate code and clean up the test readability quite a bit

amitgalitz (approved):

build.gradle
kaituo: I didn't. The 1.3 CI started failing due to missing javassit since Oct 5.

src/test/java/org/opensearch/ad/e2e/AbstractSyntheticDataTest.java
kaituo: It is the checkpoint lookup before cold start request being made that's causing the exception. We don't lose data, just the cold start will take some intervals to finish. In real time, it is usually within one interval (our minimum interval is 1 minute). In a batch testing mode, this can last several interval.

src/test/java/org/opensearch/ad/e2e/AbstractSyntheticDataTest.java
kaituo: It's reasonable. These are old code and I am moving it to a separate file. I can do it in 3.0 PR (https://github.com/opensearch-project/anomaly-detection/pull/728) so newer branches can inherit the changes. Doing it in all of the PRs takes too much time :)

ohltyler (approved):

PR #625, opened 2022-08-03
Update BWC zip links

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description
Update BWC links to point to OpenSearch 1.1.0, since Open Distro is now officially unsupported. See OpenSearch [maintenance policy](https://opensearch.org/releases.html#maintenance-policy) on supported versions.

This PR also cleans up BWC tests to be compatible with 1.1 as the baseline. Note there was some changes to historical analysis between ODFE 1.13.0.0 (based on ES 7.10.2) => OpenSearch 1.1.0.0.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #723, opened 2022-11-11

kaituo:
### Description

We also backported #600 so that we can capture the performance data in CI output. We backported https://github.com/opensearch-project/anomaly-detection/pull/625 to update bwc zip since odfe zip is unavaiable.

Testing done:

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,33 @@
+name: Run AD benchmark
+on:
+  push:
+    branches:
+      - "*"
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  Build-ad:
+    strategy:
+      matrix:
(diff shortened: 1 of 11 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,33 @@
...
+      - "*"
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  Build-ad:
+    strategy:
+      matrix:
+        java: [17]
```
ohltyler, line 14: 2.0 CI introduced testing on java 11 as well - should we add that here too?

ohltyler: Fair enough, I'm ok with just this

ohltyler (approved):

amitgalitz (approved):

PR #724, opened 2022-11-11

kaituo:
### Description

We also backported #600 so that we can capture the performance data in CI output. We backported https://github.com/opensearch-project/anomaly-detection/pull/625 to update bwc zip since odfe zip is unavaiable.

Testing done:


build.gradle
```diff
@@ -366,7 +379,7 @@ task integTestRemote(type: RestIntegTestTask) {
     }
 }
 
-String bwcVersion = "1.13.0.0"
+String bwcVersion = "1.1.0.0"
```
ohltyler, line 382: thanks for updating these too :)

ohltyler (approved):

amitgalitz (approved):

build.gradle
kaituo: np :)

PR #245, opened 2021-09-29
Add utf-8 encoding settings in gradle settings

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

### Description
Add utf-8 encoding settings in gradle settings
 
### Issues Resolved
#244 
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #726, opened 2022-11-13
[Feature/extensions] CreateDetectorAction extends BaseExtensionRestHandler

dbwiddis:
NOTE: Requires merging [SDK #246](https://github.com/opensearch-project/opensearch-sdk-java/pull/246/files) first.

### Description

Simplifies `CreateDetectorAction` by extending `BaseExtensionRestHandler`:
 - changes `routes()` to `routeHandlers()` which auto-checks the method and path
 - removes the unnecessary check for the method and path
 - renames the handler method to avoid name conflict/override

### Issues Resolved

Fixes [SDK #245](https://github.com/opensearch-project/opensearch-sdk-java/issues/245)

### DIFF
--- a/src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
@@ -15,7 +15,6 @@ import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedT
 import static org.opensearch.rest.RestRequest.Method.POST;
 import static org.opensearch.rest.RestStatus.BAD_REQUEST;
 import static org.opensearch.rest.RestStatus.CREATED;
-import static org.opensearch.rest.RestStatus.NOT_FOUND;
 import static org.opensearch.rest.RestStatus.OK;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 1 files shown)


owaiskazi19 (approved):

owaiskazi19:
@dbwiddis rebase with feature/extensions?

ryanbogan (approved):

src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java
```diff
@@ -66,8 +61,8 @@ public RestCreateDetectorAction(ExtensionsRunner runner, AnomalyDetectorExtensio
     }
 
     @Override
-    public List<Route> routes() {
-        return List.of(new Route(POST, "/detectors"));
+    protected List<RouteHandler> routeHandlers() {
+        return List.of(new RouteHandler(POST, "/detectors", (r) -> handlePostRequest(r)));
```
dbwiddis, line 65: @owaiskazi19 FYI, I tried to use the same shortened format as the `RestHelloAction` but the internal use of `client` and `xContentRegistry` inside a `Function` falls afoul of "must be final" (even though they are defined final!)

So I had to keep it as a standard method and do the lambda version of the `Function` inline here.

PR #732, opened 2022-11-16
[Feature/extensions] Add Action for Validate detector API call

mloufra:
Signed-off-by: Frank Lou <mloufra@amazon.com>

### Description
Create an action implementing `ExtensionRestHandler` with routes for the Post detector API calls.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-sdk-java/issues/221

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -10,6 +10,7 @@ import java.util.stream.Stream;
 
 import org.opensearch.ad.rest.RestCreateDetectorAction;
 import org.opensearch.ad.rest.RestGetDetectorAction;
+import org.opensearch.ad.rest.RestValidateDetectorAction;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.client.opensearch.OpenSearchClient;
@@ -45,7 +46,7 @@ public class AnomalyDetectorExtension implements Extension {
 
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
-        return List.of(new RestCreateDetectorAction(), new RestGetDetectorAction());
+        return List.of(new RestCreateDetectorAction(), new RestGetDetectorAction(), new RestValidateDetectorAction());
     }
 
     @Override
--- a/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
@@ -21,7 +21,7 @@ public class RestValidateDetectorAction implements ExtensionRestHandler {
 
     @Override
     public List<Route> routes() {
-        return List.of(new Route(POST, "/detectors"), new Route(POST, "/module"));
+        return List.of(new Route(POST, "/module"));
     }
 
     @Override
--- a/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
@@ -21,7 +21,7 @@ public class RestValidateDetectorAction implements ExtensionRestHandler {
 
     @Override
     public List<Route> routes() {
-        return List.of(new Route(POST, "/module"));
+        return List.of(new Route(POST, "/detectors/_validate"), new Route(POST, "/detectors/_validate"));
     }
 
     @Override
--- a/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
@@ -1,6 +1,7 @@
 package org.opensearch.ad.rest;
 
 import static org.opensearch.rest.RestRequest.Method.POST;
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -0,0 +1,44 @@
...
+import org.opensearch.rest.RestHandler.Route;
+import org.opensearch.rest.RestRequest.Method;
+import org.opensearch.sdk.ExtensionRestHandler;
+
+public class RestValidateDetectorAction implements ExtensionRestHandler {
+    private final Logger logger = LogManager.getLogger(RestValidateDetectorAction.class);
+
+    @Override
+    public List<Route> routes() {
+        return List.of(new Route(POST, "/detectors"));
```
dbwiddis, line 24: Need the new routes here.

dbwiddis (changes requested):

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
mloufra: Since Validate detector is used to return the detector configuration has any issue that prevent Opensearch from creating the detector, do you means we need add new route `Head` here for verifying that resource is available?

owaiskazi19: Refer https://opensearch.org/docs/latest/monitoring-plugins/ad/api/#validate-detector for the route @mloufra.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -0,0 +1,45 @@
...
+import org.opensearch.rest.RestHandler.Route;
+import org.opensearch.rest.RestRequest.Method;
+import org.opensearch.sdk.ExtensionRestHandler;
+
+public class RestValidateDetectorAction implements ExtensionRestHandler {
+    private final Logger logger = LogManager.getLogger(RestValidateDetectorAction.class);
+
+    @Override
+    public List<Route> routes() {
+        return List.of(new Route(POST, "/detectors"), new Route(GET, "/detectors"));
```
owaiskazi19, line 25: We don't have `GET` route for validate detector

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
dbwiddis: @mloufra there are 3 requests that will be handled, but only two routes: the wildcard `{type}` will handle two of them.  The specific routes you need to map are in the issue for [SDK #221](https://github.com/opensearch-project/opensearch-sdk-java/issues/221).

mloufra: @dbwiddis @owaiskazi19 Thank you for your comments. For my understand, Validate detector having two option of validation. First one is focus on the detector configuration to find any issues which will block detector creation. Second one is focus on the source data to check the probability of complete model training. So we need two routes `POST _plugins/_anomaly_detection/detectors/_validate/detector`, one for detector, and one for model `POST _plugins/_anomaly_detection/detectors/_validate/model`.

owaiskazi19: Based on the issue we will 2 routes:
```
POST /detectors/_validate
POST /detectors/_validate/{type}

**Review by dbwiddis — APPROVED** [review]
LGTM!

File: src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java (hunk 1 above)
Comment (dbwiddis) [reply src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java]: > We are not covering the below route as a part of this issue
> 
> ```
> POST _plugins/_anomaly_detection/detectors/_validate/model
> ```

Actually, the `{type}` will include either `detector` or `model` that we'll extract with a `param.get("type")` as part of the logic later.

We could have listed the routes separately (`_validate/detector` and `_validate/model`) but using `{type}` parallels the existing plugin usage.

**Comment by dbwiddis** [comment]
@mloufra @owaiskazi19 build is failing due to changes in (merged) SDK [PR #253](https://github.com/opensearch-project/opensearch-sdk-java/pull/253) that have not yet been merged here in #725.  Hopefully when #725 is merged this will pass.

File: src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java [hunk 3]
```diff
@@ -43,7 +44,7 @@ public AnomalyDetectorExtension() {
 
     @Override
     public List<ExtensionRestHandler> getExtensionRestHandlers() {
-        return List.of(new RestCreateDetectorAction(extensionsRunner, this), new RestGetDetectorAction());
+        return List.of(new RestCreateDetectorAction(extensionsRunner, this), new RestGetDetectorAction(), new RestValidateDetectorAction());
```
owaiskazi19, line 47: We should add arguments `extensionsRunner, this` for other routes as well

dbwiddis: @owaiskazi19 that can wait until we do the implementation and know exactly what we need... if we add that then we should also update all the handler code to the new format.

owaiskazi19 (approved):

PR #740, opened 2022-11-30
Use sonatype to fetch JS instead of ci.opensearch.org

amitgalitz:

### Description
Added Job-scheduler as a dependency so it is pulled from either mavenlocal, snapshots and etc so we aren't relying on a potentially failed build from ci.opensearch.org. This would allow us to not be constantly stuck fixing bugs that cause a build failure which in turn means JS itself failed to build leading too failed integration tests.

Had issues using zipArchive later in the build.gradle file so moved dependencies up in the file

### Issues Resolved


ohltyler:
Seems some failure in BWC action: 
```
Execution failed for task ':adBwcCluster#fullRestartClusterTask'.
See https://docs.gradle.org/7.4.2/userguide/command_line_interface.html#sec:command_line_warnings
> class org.gradle.api.internal.file.FileCollectionBackedFileTree cannot be cast to class org.gradle.api.provider.Provider (org.gradle.api.internal.file.FileCollectionBackedFileTree and org.gradle.api.provider.Provider are in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @1794d431)
```

amitgalitz:
> Seems some failure in BWC action:
> 
> ```
> Execution failed for task ':adBwcCluster#fullRestartClusterTask'.
> See https://docs.gradle.org/7.4.2/userguide/command_line_interface.html#sec:command_line_warnings
> > class org.gradle.api.internal.file.FileCollectionBackedFileTree cannot be cast to class org.gradle.api.provider.Provider (org.gradle.api.internal.file.FileCollectionBackedFileTree and org.gradle.api.provider.Provider are in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @1794d431)
> ```

Yeah just noticed this from github CI, seems like I'm not able to cast the file collection as a provider. Looking into this, not 100% sure yet on solution

build.gradle
```diff
@@ -96,6 +92,70 @@ repositories {
     maven { url "https://plugins.gradle.org/m2/" }
     maven { url "https://d1nvenhzbhpy0q.cloudfront.net/snapshots/lucene/" }
 }
+configurations {
+    zipArchive
+    testImplementation {
+        exclude group: 'org.hamcrest', module: 'hamcrest-core'
```
kaituo, line 98: Why do you need to exclude hamcrest?

amitgalitz: It was already there, I just moved this from later in the file. The reason we originally have it is cause without excluding it we have a jarHell Exception caused by some conflict of hamcrest in jdk11. It was added in the initial release PR for AD in opendistro.

amitgalitz:
bwc test is failing here due to an issue outside of AD or this new changes in gradle: https://github.com/opensearch-project/OpenSearch/issues/5076

I also checked that making the same changes in 2.x work

amitgalitz:
Not sure why security is failing on java 17 only, continuing to locally test, also bwc as mentioned above is due to unrelated issue. @kaituo @ohltyler are we okay with merging this so we can at least successfully run integ test and unit test on 3.0. Also so we can run `./gradlew :run` on main branch now unrelated too if the distribution build is broken.

ohltyler:
> Not sure why security is failing on java 17 only, continuing to locally test, also bwc as mentioned above is due to unrelated issue. @kaituo @ohltyler are we okay with merging this so we can at least successfully run integ test and unit test on 3.0. Also so we can run `./gradlew :run` on main branch now unrelated too if the distribution build is broken.

Yeah SGTM. In general I think we can relax restrictions for CI & builds for now until 3.0 is stable where it's worth our time to fix any actual breaking changes on our end.

ohltyler (dismissed):

build.gradle
kaituo: could you add a comment?

amitgalitz: added

ohltyler (approved):

vibrantvarun (approved):

b4sjoo (approved):

PR #222, opened 2021-09-23
Recalculate model memory size

kaituo:
Note: since there are inter-file references, I split a big PR into multiple smaller PRs based on files to save time (1.1 cut off is due this Friday). The build will fail due to missing dependencies. I will use that PR just for review. will not merge it. Will have a big PR in the end and merge once after all review PRs get approved. For the complete code, check https://github.com/kaituo/anomaly-detection-1/tree/thresholdRCF2

### Description
This PR recalculates model memory size formula after introducing ThresholdedRandomCutForest.  Please check comments for details.

Testing done:
1. unit tests.

Experiment data:
https://gist.github.com/kaituo/77b985921eda631a534157e8e11f6c57
 
### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #211, opened 2021-09-09
Check if memory usage exceeds limit for historical analysis


ylwu-amzn:
https://github.com/opensearch-project/anomaly-detection/pull/209#discussion_r704862754

Kaituo's comment
> Do you think it is necessary to add maintenance for memory tracker just in case you don't release memory? I have one in PriorityCache for HC detector: https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/caching/PriorityCache.java#L517-L539

PR #742, opened 2022-12-01
[Feature/extensions] Validate Detector Action Handle request

vibrantvarun:
Signed-off-by: Varun Jain <varunudr@amazon.com>

### Description
Template of Handle Validate Detector request in RestValidateDetectorAction

Part of Issue: 
[https://github.com/opensearch-project/opensearch-sdk-java/issues/222](https://github.com/opensearch-project/opensearch-sdk-java/issues/222)

### DIFF
--- a/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
@@ -1,14 +1,16 @@
 package org.opensearch.ad.rest;
 
+import static org.opensearch.ad.util.RestHandlerUtils.TYPE;
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.rest.RestRequest.Method.POST;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.Collections;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
@@ -33,32 +35,29 @@ import org.opensearch.sdk.BaseExtensionRestHandler;
 import org.opensearch.sdk.ExtensionsRunner;
 import org.opensearch.sdk.RouteHandler;
 
-import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
-import static org.opensearch.ad.util.RestHandlerUtils.TYPE;
-
 public class RestValidateDetectorAction extends BaseExtensionRestHandler {
     private final Logger logger = LogManager.getLogger(RestValidateDetectorAction.class);
     private final OpenSearchClient sdkClient;
     private final NamedXContentRegistry xContentRegistry;
 
     public static final Set<String> ALL_VALIDATION_ASPECTS_STRS = Arrays
-            .asList(ValidationAspect.values())
-            .stream()
-            .map(aspect -> aspect.getName())
-            .collect(Collectors.toSet());
+        .asList(ValidationAspect.values())
+        .stream()
+        .map(aspect -> aspect.getName())
+        .collect(Collectors.toSet());
 
-    public RestValidateDetectorAction(ExtensionsRunner runner, AnomalyDetectorExtension extension){
+    public RestValidateDetectorAction(ExtensionsRunner runner, AnomalyDetectorExtension extension) {
         this.xContentRegistry = runner.getNamedXContentRegistry().getRegistry();
         this.sdkClient = extension.getClient();
     }
 
     @Override
     protected List<RouteHandler> routeHandlers() {
-        return List.of(new RouteHandler(POST, "/detectors/_validate",(r)->handleRequest(r)));
+        return List.of(new RouteHandler(POST, "/detectors/_validate", (r) -> handleRequest(r)));
     }
 
     @Override
-    public ExtensionRestResponse handleRequest(ExtensionRestRequest request){
+    public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
         if (!EnabledSetting.isADPluginEnabled()) {
             throw new IllegalStateException(CommonErrorMessages.DISABLED_ERR_MSG);
         }
@@ -67,28 +66,24 @@ public class RestValidateDetectorAction extends BaseExtensionRestHandler {
         XContentBuilder builder = null;
         ValidateAnomalyDetectorRequest validateAnomalyDetectorRequest;
         try {
-            parser= request.contentParser(this.xContentRegistry);
+            parser = request.contentParser(this.xContentRegistry);
             ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
             String typesStr = validateTypeString(request);
-            DetectorValidationIssue issue=null;
-            try{
-                detector= AnomalyDetector.parse(parser);
-//                validateAnomalyDetectorRequest= new ValidateAnomalyDetectorRequest(
-//                        detector,
-//                        typesStr,
-//                        extension,
-//                        maxMultiEntityDetectors,
-//                        maxAnomalyFeatures,
-//                        requestTimeout
-//                );
-            }catch (Exception e){
-                if(e instanceof ADValidationException){
+            DetectorValidationIssue issue = null;
+            try {
+                detector = AnomalyDetector.parse(parser);
+                // validateAnomalyDetectorRequest= new ValidateAnomalyDetectorRequest(
+                // detector,
+                // typesStr,
+                // extension,
+                // maxMultiEntityDetectors,
+                // maxAnomalyFeatures,
+                // requestTimeout
+                // );
+            } catch (Exception e) {
+                if (e instanceof ADValidationException) {
                     ADValidationException ADException = (ADValidationException) e;
-                    issue = new DetectorValidationIssue(
-                            ADException.getAspect(),
-                            ADException.getType(),
-                            ADException.getMessage()
-                    );
+                    issue = new DetectorValidationIssue(ADException.getAspect(), ADException.getType(), ADException.getMessage());
                 }
             }
 
@@ -97,13 +92,11 @@ public class RestValidateDetectorAction extends BaseExtensionRestHandler {
                 builder.startObject();
                 builder.field("issue", issue);
                 builder.endObject();
-            }catch (IOException e){
+            } catch (IOException e) {
                 e.printStackTrace();
             }
 
-
-
-        }catch (Exception e){
+        } catch (Exception e) {
             return new ExtensionRestResponse(request, RestStatus.BAD_REQUEST, builder);
         }
         return new ExtensionRestResponse(request, RestStatus.OK, "placeholder");
@@ -114,7 +107,7 @@ public class RestValidateDetectorAction extends BaseExtensionRestHandler {
         return (!Collections.disjoint(typesInRequest, ALL_VALIDATION_ASPECTS_STRS));
     }
 
-    private String validateTypeString(ExtensionRestRequest request){
+    private String validateTypeString(ExtensionRestRequest request) {
         String typesStr = request.param(TYPE);
 
         // if type param isn't blank and isn't a part of possible validation types throws exception
--- a/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
@@ -53,11 +53,10 @@ public class RestValidateDetectorAction extends BaseExtensionRestHandler {
 
     @Override
     protected List<RouteHandler> routeHandlers() {
-        return List.of(new RouteHandler(POST, "/detectors/_validate", (r) -> handleRequest(r)));
+        return List.of(new RouteHandler(POST, "/detectors/_validate", (r) -> handleValidateDetectorRequest(r)));
     }
 
-    @Override
-    public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
+    private ExtensionRestResponse handleValidateDetectorRequest(ExtensionRestRequest request) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
 package org.opensearch.ad.rest;
 
 import static org.opensearch.rest.RestRequest.Method.POST;
-import static org.opensearch.rest.RestStatus.NOT_FOUND;
-import static org.opensearch.rest.RestStatus.OK;
 
-import java.util.List;
+import java.io.IOException;
+import java.util.*;
```
owaiskazi19, line 6: Avoid wildcard imports

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
...
+        XContentBuilder builder = null;
+        ValidateAnomalyDetectorRequest validateAnomalyDetectorRequest;
+        try {
+            parser= request.contentParser(this.xContentRegistry);
+            ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+            String typesStr = validateTypeString(request);
+            DetectorValidationIssue issue=null;
+            try{
+                detector= AnomalyDetector.parse(parser);
+                //TODO Dependent on https://github.com/opensearch-project/opensearch-sdk-java/issues/211
```
owaiskazi19, line 76: What's the dependency here?

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
...
     private final Logger logger = LogManager.getLogger(RestValidateDetectorAction.class);
+    private final OpenSearchClient sdkClient;
+    private final NamedXContentRegistry xContentRegistry;
 
+    public static final Set<String> ALL_VALIDATION_ASPECTS_STRS = Arrays
+            .asList(ValidationAspect.values())
+            .stream()
+            .map(aspect -> aspect.getName())
+            .collect(Collectors.toSet());
+    public RestValidateDetectorAction(ExtensionsRunner runner, AnomalyDetectorExtension extension){
```
owaiskazi19, line 51: New line before the method please :)

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
...
 import org.opensearch.rest.RestHandler.Route;
 import org.opensearch.rest.RestRequest.Method;
+import org.opensearch.rest.action.RestToXContentListener;
 import org.opensearch.sdk.ExtensionRestHandler;
+import org.opensearch.sdk.ExtensionsRunner;
+
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.ad.util.RestHandlerUtils.TYPE;
+import static org.opensearch.ad.util.RestHandlerUtils.VALIDATE;
+import static org.opensearch.rest.RestStatus.*;
```
owaiskazi19, line 39: Let's avoid using wildcard imports

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
...
+                            ADException.getType(),
+                            ADException.getMessage()
+                    );
+                }
+            }
+
+            try {
+                builder = XContentBuilder.builder(XContentType.JSON.xContent());
+                builder.startObject();
+                builder.field("issue", issue);
```
owaiskazi19, line 99: Why are we adding `DetectorValidationIssue` as a part of the builder response?

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
...
+            }
+
+
+
+        }catch (Exception e){
+            return new ExtensionRestResponse(request, BAD_REQUEST, builder);
         }
-        // do things with request
-        return new ExtensionRestResponse(request, OK, "placeholder");
+        return new ExtensionRestResponse(request, OK, builder);
```
owaiskazi19, line 110: Looks like we are sending the same `builder` in the response for `BAD_REQUEST` and `OK` response.

owaiskazi19:
Is this PR a template for the logic of ValidateDetector since there's no request call for the same?

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,130 @@
...
             throw new IllegalStateException(CommonErrorMessages.DISABLED_ERR_MSG);
         }
-        Method method = request.method();
-
-        if (!Method.POST.equals(method)) {
-            return new ExtensionRestResponse(
-                request,
-                NOT_FOUND,
-                "Extension REST action improperly configured to handle " + request.toString()
-            );
```
owaiskazi19, line 39: Why are we removing this check here?

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: Ack

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: Ack

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: Ack

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: @dbwiddis will be following the work which is being done by you in getDetector to get an idea of how to import some parameters in validation.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: yeah  I think we can keep it as initially I thought to be a template just to fill the method. But, as this is a check to validate method as POST we can keep that.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: Actually once the logic will be implemented in validate detector, sending builder in response object will make more sense. But, yeah as of now we are keeping it as template I am replacing it with the "placeholder"

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: We are adding DetectorValidationIssue in builder response as when there is a ADValidationexception it sends issue object in response which in turns what exactly was the issue in parsing. When there is no error then issue object remains null. In the plugin arch it is like when there is an error then return issue object and in case of successful execution no issue object is sent in response.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
dbwiddis: We could remove the method check if we use the new `routeHandlers()` syntax.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,144 @@
...
+            .map(aspect -> aspect.getName())
+            .collect(Collectors.toSet());
 
+    public RestValidateDetectorAction(ExtensionsRunner runner, AnomalyDetectorExtension extension){
+        this.xContentRegistry = runner.getNamedXContentRegistry().getRegistry();
+        this.sdkClient = extension.getClient();
+    }
     @Override
     public List<Route> routes() {
         return List.of(new Route(POST, "/detectors/_validate"), new Route(POST, "/detectors/_validate/{type}"));
```
dbwiddis, line 62: While this works, there's a better syntax available with the `BaseExtensionRestHandler`.  So you can:
1. change `implements ExtensionRestHandler` to `extends BaseExtensionRestHandler`.
2. change this method to `List<RouteHandler> routeHandlers()`
3. make the list of `new RouteHandler(POST, "/detectors/_validate", (r) -> handlePostRequest(r))` (and another one for the other route pointing to the same method).

See 
https://github.com/opensearch-project/anomaly-detection/blob/71010ca7f99c889b9d8a9a9cba3a3b2f0ba8afc9/src/main/java/org/opensearch/ad/rest/RestCreateDetectorAction.java#L63-L66

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
dbwiddis: +1. The other half of validation requires more complex work with settings and searching existing detectors (e.g., won't let you create too many detectors).  Sinceo @vibrantvarun only had a short time to work on this I requested he submit what he did so far. I'll pick up [SDK #222](https://github.com/opensearch-project/opensearch-sdk-java/issues/222) next week.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
dbwiddis: The full response will be part of [SDK #224](https://github.com/opensearch-project/opensearch-sdk-java/issues/224).  The intent of issue 222 was just to create placeholders here.  This is a good start.

dbwiddis (changes requested):
Great start.  Can we use the new `routeHandler()` syntax?

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
owaiskazi19: @vibrantvarun so let's not call it as a dependency on #211 in TODO

owaiskazi19 (approved):

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: ACk

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,122 @@
...
+    public RestValidateDetectorAction(ExtensionsRunner runner, AnomalyDetectorExtension extension) {
+        this.xContentRegistry = runner.getNamedXContentRegistry().getRegistry();
+        this.sdkClient = extension.getClient();
+    }
 
     @Override
-    public List<Route> routes() {
-        return List.of(new Route(POST, "/detectors/_validate"), new Route(POST, "/detectors/_validate/{type}"));
+    protected List<RouteHandler> routeHandlers() {
+        return List.of(new RouteHandler(POST, "/detectors/_validate", (r) -> handleRequest(r)));
```
dbwiddis, line 56: The `handleRequest()` method below is an override of the superclass method which actually does the filtering/parsing of the route handler.  So in reality you'll never actually use the `(r) -> handleRequest(r)` part here; it'll jump straight to the overridden method and ignore all the snazzy route handler method-and-path-checking goodness.

While that happens to give us the result we want in this case, it's not clear.  This is "shadowing" and is almost always a Bad™️ thing.

So pick any other name for the request handler below, so that it doesn't override.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
```diff
@@ -1,44 +1,122 @@
...
+                builder.field("issue", issue);
+                builder.endObject();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+
+        } catch (Exception e) {
+            return new ExtensionRestResponse(request, RestStatus.BAD_REQUEST, builder);
+        }
+        return new ExtensionRestResponse(request, RestStatus.OK, "placeholder");
```
dbwiddis, line 102: I know the issue said to just send a placeholder response, but in our case we know that the success response is an empty JSON object.   So we could:
```java
return new ExtensionRestResponse(
    request,
    RestStatus.OK,
    JsonXContent.contentBuilder()
        .startObject()
        .endObject()
);
```

dbwiddis (changes requested):
Shadowing is not a good thing.

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/rest/RestValidateDetectorAction.java
vibrantvarun: `JsonXContent.contentBuilder()
        .startObject()
        .endObject()`
        
The above code throws an IOException. If I add throws IO exception in the method signature then it will impact the call which is been made from routeHandlers() method.  Another way of handling the exception could be adding a try/catch block just to handle an IOException, which here I think won't be a good idea as adding the try/catch block would be adding some boilerplate code for just a template.

dbwiddis (approved):

PR #743, opened 2022-12-01
Speed up cold start

kaituo:
### Description
If historical data is enough, a single stream detector takes 1 interval for cold start to be triggered + 1 interval for the state document to be updated. Similar to single stream detectors, HCAD cold start needs 2 intervals and one more interval to make sure an entity appears more than once. So HCAD needs three intervals to complete cold starts. Long initialization is the single most complained problem of AD. This PR reduces both single stream and HCAD detectors' initialization time to one interval + 1 minute by
* delaying real time cache update by one minute so we will have trained models by then and update the state document accordingly.
* disable door keeper by default so that we won't wait extra interval in HCAD

Testing done:
1. verified the cold start time is reduced using the frontend.

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -60,6 +61,7 @@ import org.opensearch.ad.util.DiscoveryNodeFilterer;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
 import org.opensearch.common.xcontent.XContentBuilder;
@@ -502,7 +504,12 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
             // For a HCAD detector, we don't need to save on the detector level.
             // We return 0 or Double.NaN rcf score if there is no error.
             if ((response.getAnomalyScore() <= 0 || Double.isNaN(response.getAnomalyScore())) && response.getError() == null) {
-                updateRealtimeTask(response, detectorId);
+                threadPool
+                    .schedule(
+                        () -> { updateRealtimeTask(response, detectorId); },
+                        new TimeValue(60, TimeUnit.SECONDS),
+                        AnomalyDetectorPlugin.AD_THREAD_POOL_NAME
+                    );
                 return;
             }
             IntervalTimeConfiguration windowDelay = (IntervalTimeConfiguration) jobParameter.getWindowDelay();
--- a/src/main/java/org/opensearch/ad/caching/PriorityCache.java
+++ b/src/main/java/org/opensearch/ad/caching/PriorityCache.java
@@ -55,6 +55,7 @@ import org.opensearch.ad.model.ModelProfile;
 import org.opensearch.ad.ratelimit.CheckpointMaintainWorker;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
+import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.ad.util.DateUtils;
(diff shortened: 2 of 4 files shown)


src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -502,7 +504,12 @@ private void indexAnomalyResult(
             // For a HCAD detector, we don't need to save on the detector level.
             // We return 0 or Double.NaN rcf score if there is no error.
             if ((response.getAnomalyScore() <= 0 || Double.isNaN(response.getAnomalyScore())) && response.getError() == null) {
-                updateRealtimeTask(response, detectorId);
+                threadPool
+                    .schedule(
+                        () -> { updateRealtimeTask(response, detectorId); },
+                        new TimeValue(60, TimeUnit.SECONDS),
```
ohltyler, line 510: Could this be set to an even shorter amount of time? Or in your experience is 60s a good amount of time to ensure trained models?

kaituo: 60s is a good amount of time to ensure trained models. But it can be longer due to various reasons.

ohltyler:
If HCAD cold start time is reduced to 1 interval + 1 minute how is it ensured that an entity appears more than once?

kaituo:
> If HCAD cold start time is reduced to 1 interval + 1 minute how is it ensured that an entity appears more than once?

I made the tradeoff to not do that by default. Users can set it to true if they want it.

ohltyler:
> > If HCAD cold start time is reduced to 1 interval + 1 minute how is it ensured that an entity appears more than once?
> 
> I made the tradeoff to not do that by default. Users can set a dynamic setting to true if they want it.

Got it. What's the impact of the tradeoff roughly? Is it that the entities stored for the model after one interval may not be the most ideal?

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
ohltyler: If it's longer, the only impact is that the cold start will take one more interval, correct? (assuming it is fully trained by the next interval)

kaituo:
> > > If HCAD cold start time is reduced to 1 interval + 1 minute how is it ensured that an entity appears more than once?
> > 
> > 
> > I made the tradeoff to not do that by default. Users can set a dynamic setting to true if they want it.
> 
> Got it. What's the impact of the tradeoff roughly? Is it that the entities stored for the model after one interval may not be the most ideal?

the entities stored for the model may only appear once and stayed in memory. So we could have more memory usage if there are a lot of such entities.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
kaituo: yes

ohltyler (approved):

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
amitgalitz: By ensure trained model? what does this exactly include? I understand we need one interval first to trigger cold start, these 60 seconds are to get back the model from RCF from the data from the the one interval we needed? make sure its on the checkpoint index?

amitgalitz:
> > > > If HCAD cold start time is reduced to 1 interval + 1 minute how is it ensured that an entity appears more than once?
> > > 
> > > 
> > > I made the tradeoff to not do that by default. Users can set a dynamic setting to true if they want it.
> > 
> > 
> > Got it. What's the impact of the tradeoff roughly? Is it that the entities stored for the model after one interval may not be the most ideal?
> 
> the entities stored for the model may only appear once and stayed in memory. So we could have more memory usage if there are a lot of such entities.

I am still a little confused on this, if doorkeeper is enabled then we check if a modelID only appears once and if it does after the next interval then we delete it so save memory? Also if an entity only appeared once in the first interval and not in the next we would be delaying HCAD start until at least one entity appears twice?

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
amitgalitz: we also update our real-time cache every-time we get a non-null, no error or non 0 anomaly score right?

amitgalitz:
Should we merge this to 2.x instead and backport to release branches and then forward port to main?

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
kaituo: 60s is to ensure cold start and writing back to checkpoint index. 

For our real-time cache, we don't update if the current value is the same as the cached values.

kaituo:
> Should we merge this to 2.x instead and backport to release branches and then forward port to main?

yes, I will close this PR and create a new one to main branch. Did further optimization to reduce the cold start to 1 minute. @ohltyler @amitgalitz

Issue #502, opened 2022-04-14
[BUG] Discrepancy between Profile API and realtime tasks API


amitgalitz:
**What is the bug?**
There is discrepancy between the results returned from the profile API and search tasks API. This occurs when restarting an Opensearch cluster that already has a HCAD running. 
**How can one reproduce the bug?**
1.  Start Opensearch cluster and dashboards with AD plugin
2. create and wait for HCAD to be running and returning anomaly results
3. stop and start OpenSearch cluster
4. HCAD will go back to showing as initializing again on frontend
5. Profile API for detector will return response of `running`
6. Search request to `..._plugins/_anomaly_detection/detectors/tasks/_search` with body of:
```
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "detector_id": "N64oKYABE8ZUzJlQp33e"
          }
        },
        {
          "term": {
            "task_type": "REALTIME_HC_DETECTOR"
          }
        },
        {
          "term": {
            "is_latest": "true"
          }
        }
      ]
    }
  }
}
```
returns response with `initializing`

**What is the expected behavior?**
HCAD to show as `running` on frontend and all relevant API responses

**What is your host/environment?**
 - Opensearch 2.0.0-alpha1-snapshot (4/14/22)

ylwu-amzn:
For HC detector:
- Realtime task calculate state&init_progress with RCF total updates, check https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java#L550-L556, If detector restarted, the RCF model cache will be cleaned, and detector will not execute until next job run. So the realtime task get 0 total updates and will show as `INIT`.

- Profile API will select AD result index to check if there are any result generated for detector to tell whether it passed initialization of not , check https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/AnomalyDetectorProfileRunner.java#L451-L461, if detector passed initialization, it will generate some AD result and store in result index. When detector restarted, the AD result still in result index, so profile API can find some result and regard the state as `RUNNING`.

kaituo:
I experienced it myself without restarting a node. My case is when the data runs out the HCAD detector status change to init. Will debug about this.

kaituo:
[...]

PR #753, opened 2022-12-09
Speed up cold start

kaituo:
### Description
If historical data is enough, a single stream detector takes 1 interval for cold start to be triggered + 1 interval for the state document to be updated. Similar to single stream detectors, HCAD cold start needs 2 intervals and one more interval to make sure an entity appears more than once. So HCAD needs three intervals to complete cold starts. Long initialization is the single most complained problem of AD. This PR reduces both single stream and HCAD detectors' initialization time to 1 minute by
* delaying real time cache update by one minute when we receive ResourceNotFoundException in single stream detectors or when the init progress of HCAD real time cache is 0. Thus, we can finish the cold start and writing checkpoint one minute later and update the state document accordingly. This optimization saves one interval to wait for the state document update.
* disable the door keeper by default so that we won't have to wait an extra interval in HCAD.
* trigger cold start when starting a real time detector. This optimization saves one interval to wait for the cold start to be triggered.

Testing done:
* verified the cold start time is reduced to 1 minute.
* added tests for new code.

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
@@ -14,21 +14,18 @@ package org.opensearch.ad;
 import static org.opensearch.action.DocWriteResponse.Result.CREATED;
 import static org.opensearch.action.DocWriteResponse.Result.UPDATED;
 import static org.opensearch.ad.AnomalyDetectorPlugin.AD_THREAD_POOL_NAME;
-import static org.opensearch.ad.constant.CommonErrorMessages.CAN_NOT_FIND_LATEST_TASK;
 import static org.opensearch.ad.util.RestHandlerUtils.XCONTENT_WITH_TYPE;
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
 import java.io.IOException;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
 import org.opensearch.action.ActionListener;
 import org.opensearch.action.get.GetRequest;
 import org.opensearch.action.get.GetResponse;
@@ -37,15 +34,10 @@ import org.opensearch.action.support.WriteRequest;
 import org.opensearch.ad.common.exception.AnomalyDetectionException;
 import org.opensearch.ad.common.exception.EndRunException;
 import org.opensearch.ad.common.exception.InternalFailure;
-import org.opensearch.ad.common.exception.ResourceNotFoundException;
-import org.opensearch.ad.indices.ADIndex;
 import org.opensearch.ad.indices.AnomalyDetectionIndices;
 import org.opensearch.ad.model.ADTaskState;
+import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.AnomalyDetectorJob;
-import org.opensearch.ad.model.AnomalyResult;
-import org.opensearch.ad.model.DetectorProfileName;
-import org.opensearch.ad.model.FeatureData;
-import org.opensearch.ad.model.IntervalTimeConfiguration;
 import org.opensearch.ad.rest.handler.AnomalyDetectorFunction;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.task.ADTaskManager;
@@ -53,12 +45,7 @@ import org.opensearch.ad.transport.AnomalyResultAction;
 import org.opensearch.ad.transport.AnomalyResultRequest;
 import org.opensearch.ad.transport.AnomalyResultResponse;
 import org.opensearch.ad.transport.AnomalyResultTransportAction;
-import org.opensearch.ad.transport.ProfileAction;
-import org.opensearch.ad.transport.ProfileRequest;
-import org.opensearch.ad.transport.handler.AnomalyIndexHandler;
-import org.opensearch.ad.util.DiscoveryNodeFilterer;
 import org.opensearch.client.Client;
-import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
@@ -66,7 +53,6 @@ import org.opensearch.common.xcontent.XContentBuilder;
 import org.opensearch.common.xcontent.XContentParser;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.commons.InjectSecurity;
-import org.opensearch.commons.authuser.User;
 import org.opensearch.jobscheduler.spi.JobExecutionContext;
 import org.opensearch.jobscheduler.spi.LockModel;
 import org.opensearch.jobscheduler.spi.ScheduledJobParameter;
@@ -88,11 +74,11 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
     private int maxRetryForEndRunException;
     private Client client;
     private ThreadPool threadPool;
-    private AnomalyIndexHandler<AnomalyResult> anomalyResultHandler;
     private ConcurrentHashMap<String, Integer> detectorEndRunExceptionCount;
     private AnomalyDetectionIndices anomalyDetectionIndices;
-    private DiscoveryNodeFilterer nodeFilter;
     private ADTaskManager adTaskManager;
+    private NodeStateManager nodeStateManager;
+    private ExecuteADResultResponseRecorder recorder;
 
     public static AnomalyDetectorJobRunner getJobRunnerInstance() {
         if (INSTANCE != null) {
@@ -120,10 +106,6 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
         this.threadPool = threadPool;
     }
 
-    public void setAnomalyResultHandler(AnomalyIndexHandler<AnomalyResult> anomalyResultHandler) {
-        this.anomalyResultHandler = anomalyResultHandler;
-    }
-
     public void setSettings(Settings settings) {
         this.settings = settings;
         this.maxRetryForEndRunException = AnomalyDetectorSettings.MAX_RETRY_FOR_END_RUN_EXCEPTION.get(settings);
@@ -137,8 +119,12 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
         this.anomalyDetectionIndices = anomalyDetectionIndices;
     }
 
-    public void setNodeFilter(DiscoveryNodeFilterer nodeFilter) {
-        this.nodeFilter = nodeFilter;
+    public void setNodeStateManager(NodeStateManager nodeStateManager) {
+        this.nodeStateManager = nodeStateManager;
+    }
+
+    public void setExecuteADResultResponseRecorder(ExecuteADResultResponseRecorder recorder) {
+        this.recorder = recorder;
     }
 
     @Override
@@ -159,28 +145,49 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
         final LockService lockService = context.getLockService();
 
         Runnable runnable = () -> {
-            if (jobParameter.getLockDurationSeconds() != null) {
-                lockService
-                    .acquireLock(
-                        jobParameter,
-                        context,
-                        ActionListener
-                            .wrap(lock -> runAdJob(jobParameter, lockService, lock, detectionStartTime, executionStartTime), exception -> {
-                                indexAnomalyResultException(
-                                    jobParameter,
-                                    lockService,
-                                    null,
-                                    detectionStartTime,
-                                    executionStartTime,
-                                    exception,
-                                    false
-                                );
-                                throw new IllegalStateException("Failed to acquire lock for AD job: " + detectorId);
-                            })
-                    );
-            } else {
-                log.warn("Can't get lock for AD job: " + detectorId);
-            }
+            nodeStateManager.getAnomalyDetector(detectorId, ActionListener.wrap(detectorOptional -> {
+                if (!detectorOptional.isPresent()) {
+                    log.error(new ParameterizedMessage("fail to get detector [{}]", detectorId));
+                    return;
+                }
+                AnomalyDetector detector = detectorOptional.get();
+
+                if (jobParameter.getLockDurationSeconds() != null) {
+                    lockService
+                        .acquireLock(
+                            jobParameter,
+                            context,
+                            ActionListener
+                                .wrap(
+                                    lock -> runAdJob(
+                                        jobParameter,
+                                        lockService,
+                                        lock,
+                                        detectionStartTime,
+                                        executionStartTime,
+                                        recorder,
+                                        detector
+                                    ),
+                                    exception -> {
+                                        indexAnomalyResultException(
+                                            jobParameter,
+                                            lockService,
+                                            null,
+                                            detectionStartTime,
+                                            executionStartTime,
+                                            exception,
+                                            false,
+                                            recorder,
+                                            detector
+                                        );
+                                        throw new IllegalStateException("Failed to acquire lock for AD job: " + detectorId);
+                                    }
+                                )
+                        );
+                } else {
+                    log.warn("Can't get lock for AD job: " + detectorId);
+                }
+            }, e -> log.error(new ParameterizedMessage("fail to get detector [{}]", detectorId), e)));
         };
 
         ExecutorService executor = threadPool.executor(AD_THREAD_POOL_NAME);
@@ -195,13 +202,17 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
      * @param lock lock to run job
      * @param detectionStartTime detection start time
      * @param executionStartTime detection end time
+     * @param recorder utility to record job execution result
+     * @param detector associated detector accessor
      */
     protected void runAdJob(
         AnomalyDetectorJob jobParameter,
         LockService lockService,
         LockModel lock,
         Instant detectionStartTime,
-        Instant executionStartTime
+        Instant executionStartTime,
+        ExecuteADResultResponseRecorder recorder,
+        AnomalyDetector detector
     ) {
         String detectorId = jobParameter.getName();
         if (lock == null) {
@@ -212,7 +223,9 @@ public class AnomalyDetectorJobRunner implements ScheduledJobRunner {
                 detectionStartTime,
                 executionStartTime,
                 "Can't run AD job due to null lock",
-                false
+                false,
+                recorder,
+                detector
             );
             return;
         }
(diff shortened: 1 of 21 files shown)


kaituo:
BWC tests failed:

 * What went wrong:
> `cluster{::adBwcCluster0}` failed to wait for cluster health yellow after 40 SECONDS
    503 Service Unavailable

amitgalitz:
>This PR reduces both single stream and HCAD detectors' initialization time to 1 minute by
Initial question, I see this PR has an improvement to reduce initialization time to just 1 minute not 1 interval + 1 minute. Why did we wait one interval before starting cold start? Does this mean if there is enough historical data, even if the interval is 10 minutes long, we can have a detector running in 1 minute?

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -0,0 +1,268 @@
...
+                        response.getDetectorIntervalInMinutes(),
+                        response.getError()
+                    );
+                }, e -> { log.error("Failed to update latest realtime task for " + detectorId, e); }));
+            };
+            if (!adTaskManager.isHCRealtimeTaskStartInitializing(detectorId)) {
+                // real time init progress is 0 may mean this is a newly started detector
+                // Delay real time cache update by one minute. If we are in init status, the delay may give the model training time to
+                // finish. We can change the detector running immediately instead of waiting for the next interval.
+                threadPool.schedule(profileHCInitProgress, new TimeValue(60, TimeUnit.SECONDS), AnomalyDetectorPlugin.AD_THREAD_POOL_NAME);
```
amitgalitz, line 148: Previous PR had this extra 60 time wait when `response.getAnomalyScore() <= 0 || Double.isNaN(response.getAnomalyScore())) && response.getError() == null` was true. 
1. Is limiting this wait time to when init is 0 for hcad and when single stream throws `ResourceNotFoundException` narrowing the scope of this wait only happening when detector is first being created? 
2. Just to make sure we aren't going to get the discrepancy that sometimes happen between profile api and real time task api cause here we are internally checking if the realtime task already exists and we are above init of 0?

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -0,0 +1,268 @@
...
+            );
+            String resultIndex = detector.getResultIndex();
+            if (resultIndex != null && !anomalyDetectionIndices.doesIndexExist(resultIndex)) {
+                // Set result index as null, will write exception to default result index.
+                anomalyResultHandler.index(anomalyResult, detectorId, null);
+            } else {
+                anomalyResultHandler.index(anomalyResult, detectorId, resultIndex);
+            }
+
+            if (errorMessage.contains(CommonErrorMessages.NO_CHECKPOINT_ERR_MSG) && !detector.isMultiCategoryDetector()) {
```
amitgalitz, line 237: Do single entity detectors also need 1 minute? if we restarted the detector is there any reason for us to believe initialization will be any quicker in the case we already have some results or when stopping detector we delete the model checkpoint along with any other relevant stored info?

kaituo:
> > This PR reduces both single stream and HCAD detectors' initialization time to 1 minute by
> 
> Initial question, I see this PR has an improvement to reduce initialization time to just 1 minute not 1 interval + 1 minute. Why did we wait one interval before starting cold start? Does this mean if there is enough historical data, even if the interval is 10 minutes long, we can have a detector running in 1 minute?

Back then, I did manage to focus on improve cold start speed. This is long overdue.

Yes, even if the interval is 10 minutes long, we can have a detector running in 1 minute.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -159,28 +145,49 @@ public void runJob(ScheduledJobParameter scheduledJobParameter, JobExecutionCont
...
-                                    exception,
-                                    false
-                                );
-                                throw new IllegalStateException("Failed to acquire lock for AD job: " + detectorId);
-                            })
-                    );
-            } else {
-                log.warn("Can't get lock for AD job: " + detectorId);
-            }
+            nodeStateManager.getAnomalyDetector(detectorId, ActionListener.wrap(detectorOptional -> {
```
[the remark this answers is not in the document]
kaituo: yes, when the workflow reaches this line, detector index should exist. Otherwise, we will get index not found exception.

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -234,11 +234,12 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
     private ClientUtil clientUtil;
     private DiscoveryNodeFilterer nodeFilter;
     private IndexUtils indexUtils;
-    private ADTaskCacheManager adTaskCacheManager;
```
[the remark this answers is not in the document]
kaituo: The variable is only used in createComponents and can be created as a local variable instead of an instance variable.

src/main/java/org/opensearch/ad/settings/EnabledSetting.java
```diff
@@ -39,8 +39,9 @@ public class EnabledSetting extends AbstractSetting {
 
     public static final String LEGACY_OPENDISTRO_AD_BREAKER_ENABLED = "opendistro.anomaly_detection.breaker.enabled";
 
-    public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED =
-        "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";;
+    public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED = "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";
+
+    public static final String DOOR_KEEPER_IN_CACHE_ENABLED = "plugins.anomaly_detection.door_keeper_in_cache.enabled";;
```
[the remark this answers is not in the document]
kaituo: created https://github.com/opensearch-project/anomaly-detection/issues/755

src/main/java/org/opensearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -696,7 +696,7 @@ private Exception coldStartIfNoModel(AtomicReference<Exception> failure, Anomaly
         }
         LOG.info("Trigger cold start for {}", detector.getDetectorId());
         coldStart(detector);
-        return previousException.orElse(new InternalFailure(adID, NO_MODEL_ERR_MSG));
+        return previousException.orElse(exp);
```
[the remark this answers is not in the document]
kaituo: To catch the resource not found exception on the job scheduler and delay updating real time cache by 1 minute. Otherwise, we only get internal failure and has to process its error msgs.

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartWorker.java
```diff
@@ -114,15 +120,36 @@ protected void executeRequest(EntityRequest coldStartRequest, ActionListener<Voi
...
+                try {
+                    if (!detectorOptional.isPresent()) {
+                        LOG.error(new ParameterizedMessage("fail to get detector [{}]", detectorId));
+                        return;
+                    }
+                    AnomalyDetector detector = detectorOptional.get();
+                    EntityModel model = modelState.getModel();
+                    // load to cache if cold start succeeds
+                    if (model != null && model.getTrcf() != null) {
+                        cacheProvider.get().hostIfPossible(detector, modelState);
```
amitgalitz, line 134: previously did we not load to cache? Cant find where this logic was previously if it was

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -0,0 +1,268 @@
...
+            anomalyResultHandler.index(anomalyResult, detectorId, resultIndex);
+            updateRealtimeTask(response, detectorId);
+        } catch (EndRunException e) {
+            throw e;
+        } catch (Exception e) {
+            log.error("Failed to index anomaly result for " + detectorId, e);
+        }
+    }
+
+    private void updateRealtimeTask(AnomalyResultResponse response, String detectorId) {
```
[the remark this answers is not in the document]
kaituo: added comment:

Update real time task (one document per detector in state index). If the real-time task has no changes compared with local cache, the task won't update. Task only updates when the state changed, or any error happened, or AD job stopped. Task is mainly consumed by the front-end to track detector status. For single-stream detectors, we embed model total updates in AnomalyResultResponse and update state accordingly. For HCAD, we won't wait for model finishing updating before returning a response to the job scheduler since it might be long before all entities finish execution. So we don't embed model total updates in AnomalyResultResponse. Instead, we issue a profile request to poll each model node and get the maximum total updates among all models.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -2820,6 +2820,13 @@ public boolean skipUpdateHCRealtimeTask(String detectorId, String error) {
             && Objects.equals(error, realtimeTaskCache.getError());
     }
 
+    public boolean isHCRealtimeTaskStartInitializing(String detectorId) {
+        ADRealtimeTaskCache realtimeTaskCache = adTaskCacheManager.getRealtimeTaskCache(detectorId);
+        return realtimeTaskCache != null
+            && realtimeTaskCache.getInitProgress() != null
+            && realtimeTaskCache.getInitProgress().floatValue() > 0;
```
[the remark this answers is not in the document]
kaituo: We don't need it since JVM will add .floatValue() automatically via unboxing. Prefer to keep it as it does not hurt and make the logic clear.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: 1. yes, you are right.
2. I haven't fixed the discrepancy issue. Tracked here: https://github.com/opensearch-project/anomaly-detection/issues/502 Plan to do it after the speed up PR is merged so the PR is not too big.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: yeah, 1 minute is to give enough leeway for training and checkpoint write, regardless of single stream or HCAD detectors.

For restart detector, as mentioned earlier, plan to do it after the speed up PR is merged so the PR is not too big. Currently, it is still an issue.

When stopping detector, we will delete everything except results. It will have to go through a new round of cold start.

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartWorker.java
kaituo: Previously, we don't load to cache when cold start finishes. We would have to wait for the next interval to make it happen.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
amitgalitz: sounds good

ohltyler: For the restart scenario, since the checkpoint would have been deleted already, wouldn't it automatically go through this flow and have the 1 minute delay? In other words, what needs to be done differently in the restart scenario vs. the initial start scenario?

I may answer my own question as I keep reading

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -0,0 +1,278 @@
...
+                        updateLatestRealtimeTask(
+                            detectorId,
+                            taskState,
+                            totalUpdates,
+                            detector.getDetectorIntervalInMinutes(),
+                            totalUpdates > 0 ? "" : errorMessage
+                        );
+                    }, e -> {
+                        log.error("Fail to eecute RCFRollingAction", e);
+                        updateLatestRealtimeTask(detectorId, taskState, null, null, errorMessage);
```
ohltyler, line 266: should the realtime task error message be `e` instead of `errorMessage`? The latter will always be the one containing `CommonErrorMessages.NO_CHECKPOINT_ERR_MSG`. I'm not sure if it's more appropriate to keep that as the error message here, or if it's more useful to use the `e` error as returned by the polling failure. Maybe the `e` could be too unrelated to issues with the detector itself since it could be an issue with cluster, a particular node, etc.

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartWorker.java
```diff
@@ -114,15 +120,36 @@ protected void executeRequest(EntityRequest coldStartRequest, ActionListener<Voi
...
 
-        ActionListener<Void> failureListener = ActionListener.delegateResponse(listener, (delegateListener, e) -> {
-            if (ExceptionUtil.isOverloaded(e)) {
-                LOG.error("OpenSearch is overloaded");
-                setCoolDownStart();
+        ActionListener<Void> coldStartListener = ActionListener.wrap(r -> {
+            nodeStateManager.getAnomalyDetector(detectorId, ActionListener.wrap(detectorOptional -> {
+                try {
+                    if (!detectorOptional.isPresent()) {
+                        LOG.error(new ParameterizedMessage("fail to get detector [{}]", detectorId));
```
ohltyler, line 127: Can we add a debug log here that the detector failed to get added to cache after cold start?

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -0,0 +1,278 @@
...
+                        // if there are updates, don't record failures
+                        updateLatestRealtimeTask(
+                            detectorId,
+                            taskState,
+                            totalUpdates,
+                            detector.getDetectorIntervalInMinutes(),
+                            totalUpdates > 0 ? "" : errorMessage
+                        );
+                    }, e -> {
+                        log.error("Fail to eecute RCFRollingAction", e);
```
ohltyler, line 265: nit: typo in `execute`

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: note it is restarting cluster not detector. So the checkpoint won't be deleted.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -0,0 +1,278 @@
...
+                        log.error("Can't find latest realtime task of detector " + detectorId);
+                        adTaskManager.removeRealtimeTaskCache(detectorId);
+                    } else {
+                        log.error("Failed to update latest realtime task for detector " + detectorId, e);
+                    }
+                })
+            );
+    }
+
+    public void indexAnomalyResultException(
```
ohltyler, line 211: can you add a comment here that it's not only indexing the result with the exception, but also updating the task state after 60s if the exception is related to cold start (index not found exceptions)?

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorJobActionHandler.java
```diff
@@ -106,37 +110,76 @@ public IndexAnomalyDetectorJobActionHandler(
...
 
     /**
      * Start anomaly detector job.
      * 1. If job doesn't exist, create new job.
      * 2. If job exists: a). if job enabled, return error message; b). if job disabled, enable job.
      * @param detector anomaly detector
+     * @param listener Listener to send responses
      */
-    public void startAnomalyDetectorJob(AnomalyDetector detector) {
+    public void startAnomalyDetectorJob(AnomalyDetector detector, ActionListener<AnomalyDetectorJobResponse> listener) {
```
ohltyler, line 123: Just to clarify: this start listener is created & injected throughout the job handler so that whenever the job response is received, there's the extra step of trying to index results and update detector state with a 60s delay, correct?

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: yeah, e could be too unrelated to issues with the detector itself since it could be an issue with cluster, a particular node, etc. errorMessage is the one due to running AnomalyResultAction and the real time cache mostly records that.

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartWorker.java
kaituo: The error is that we cannot find detector config from anomaly-detectors index, not failing to add it to cache after cold start. Is the error log enough?

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorJobActionHandler.java
```diff
@@ -274,8 +339,9 @@ private void onIndexAnomalyDetectorJobResponse(IndexResponse response, AnomalyDe
      * 2.If job exists: a).if job state is disabled, return error message; b).if job state is enabled, disable job.
      *
      * @param detectorId detector identifier
+     * @param listener Listener to send responses
      */
-    public void stopAnomalyDetectorJob(String detectorId) {
+    public void stopAnomalyDetectorJob(String detectorId, ActionListener<AnomalyDetectorJobResponse> listener) {
```
ohltyler, line 344: what is the purpose for having the listener on the stop detector job? currently the task state is already getting updated immediately after this action is finished. But with the new listener, wouldn't this trigger `indexAnomalyResult()` or `indexAnomalyResultException()`?

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: fixed

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: added

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorJobActionHandler.java
kaituo: yes

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: it is a typo. Will change to cluster restart.

src/main/java/org/opensearch/ad/ratelimit/EntityColdStartWorker.java
kaituo: Make sense. Changed to 

```
if (!detectorOptional.isPresent()) {
                        LOG.error(new ParameterizedMessage("fail to load trained model [{}] to cache due to the detector not being found.", modelState.getModelId()));
                        return;
                    }
```

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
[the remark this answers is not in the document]
amitgalitz: Thank you

amitgalitz (dismissed):

ohltyler (approved):
LGTM - I think this will be a very visible improvement for users!

ohltyler (approved):

amitgalitz (approved):

PR #761, opened 2022-12-15
Communication mechanism JS (Part 1) 

vibrantvarun:
### Description
Request Response model to fetch Job details like job index and job type from extensions.

### Issues Resolved
[https://github.com/opensearch-project/opensearch-sdk-java/issues/274](https://github.com/opensearch-project/opensearch-sdk-java/issues/274)

### Companion PR
OpenSearch [https://github.com/opensearch-project/OpenSearch/pull/5586](https://github.com/opensearch-project/OpenSearch/pull/5586)

opensearch-sdk-java [https://github.com/opensearch-project/opensearch-sdk-java/pull/289](https://github.com/opensearch-project/opensearch-sdk-java/pull/289)

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -27,7 +27,6 @@ import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
-import org.opensearch.extensions.JobDetails;
 import org.opensearch.sdk.BaseExtension;
 import org.opensearch.sdk.ExtensionRestHandler;
 import org.opensearch.sdk.ExtensionsRunner;
@@ -112,10 +111,15 @@ public class AnomalyDetectorExtension extends BaseExtension {
         return client;
     }
 
-    // TODO: After AnomalyDetectorJob.java is uncommented, fetch jobIndex from AnomalyDetectorJob
     @Override
-    public JobDetails getJobDetails() {
-        return new JobDetails(AnomalyDetectorPlugin.AD_JOB_TYPE, ".opendistro-anomaly-detector-jobs");
+    public String getJobType() {
+        return AnomalyDetectorPlugin.AD_JOB_TYPE;
+    }
+
+    // TODO After circular dependency of JS is removed and AnomalyDetectorJob is uncommented, import JobIndex from AnomalyDetectorJob.java.
+    @Override
+    public String getJobIndex() {
+        return ".opendistro-anomaly-detector-jobs";
     }
 
     public static void main(String[] args) throws IOException {
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -27,6 +27,7 @@ import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
+import org.opensearch.extensions.JobDetails;
 import org.opensearch.sdk.BaseExtension;
 import org.opensearch.sdk.ExtensionRestHandler;
 import org.opensearch.sdk.ExtensionsRunner;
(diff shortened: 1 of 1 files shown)


dbwiddis (approved):

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -111,6 +112,12 @@ public OpenSearchClient getClient() {
         return client;
     }
 
+    // TODO: After AnomalyDetectorJob.java is uncommented, fetch jobIndex from AnomalyDetectorJob
+    @Override
+    public JobDetails getJobDetails() {
+        return new JobDetails(AnomalyDetectorPlugin.AD_JOB_TYPE, ".opendistro-anomaly-detector-jobs");
```
owaiskazi19, line 118: We should use Job index constant here `ANOMALY_DETECTOR_JOB_INDEX`. 
Also, how does `getJobDetails` is equivalent to which JobScheduler extension point here?
For ADplugin. we have `getJobIndex()`

joshpalis: @owaiskazi19 `JobDetails` is equivalent to both `getJobIndex()` and `getJobType()`. Both these strings are utilized by Job Scheduler to configure it's behavior for `onIndexModule()` and to also configure the `indexToJobProviders` which the `JobSweeper` uses to schedule jobs. I also agree that we should use the `ANOMALY_DETECTOR_JOB_INDEX` constant, @vibrantvarun it should be fine to uncomment this

owaiskazi19: I want to understand more on why are we coupling up `getJobIndex()` and `getJobType()` together in `jobDetails`?Shouldn't it follow the old design or otherwise we have to change these extension points in all the plugins/extensions using JS?

joshpalis: `getJobIndex` and `getJobType` are currently used by Job Scheduler to retrieve implementations of the `JobSchedulerExtension` interface [here](https://github.com/opensearch-project/job-scheduler/blob/7727871496c8c9aceebd415bf8169dea6284c938/src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java#L125). Simply put, these two extension points just provide strings to Job Scheduler.

We're coupling the Job Index and Job Type strings together in one request, rather than separate requests since this is the only initial information we need from extensions in order to configure both the `indicesToListen` set and `indexToJobProviders` hashmap. Separating these into two requests would be unnecessary.  

This wouldn't have any affect on how plugins are using JS, since the plugin architecture for JS remains untouched within `PluginsService`'s `loadExtensionsForPlugin` workflow. As for extensions, the only difference at this point is how we are providing these strings to Job Scheduler. How Job Scheduler uses these strings will remain unchanged.

owaiskazi19: I understand separating them would be unnecessary but that's how the old design is. We don't to bring new extension points with extensibility at this moment before covering the old ones.

vibrantvarun: Okay @owaiskazi19 . 

Following your comment

1) I will be deleting the Job Details object and then decouple both the parameters
2) I will also be making 2 transportService calls now to fetch both the parameters

vibrantvarun: Hi @joshpalis @owaiskazi19 the job index cannot be pulled from AnomalyDetectorJob.java as of now because it still has circular dependency from JS on some other methods written in the file.  If I uncomment only the required code like the Job Index initialization then I am facing an error of jacoco test coverage report. Therefore, for now I am hardcoding the value of job index.

owaiskazi19: We can fix the jacoco test coverage, right? The way you fixed it for ModelTests.

owaiskazi19:
DCO check is failing. Commit using `-s` flag. Thanks

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
vibrantvarun: Yes I did uncomment the code, but there are some dependencies which are imported from JobScheduler like Schedule.java etc. These dependencies cannot be resolved until they are imported from JS native plugin.

owaiskazi19: Makes sense. Let's create a constant in this file then

vibrantvarun: ACK

PR #766, opened 2022-12-20
Add option to run BWC tests in distribution level

zelinh:
### Description
Add option to run BWC tests in distribution level, which is running tests in the test cluster with latest distribution bundle installed and all plugins (included in the latest manifest) present. 

The command to run BWC tests in distribution level would be `./gradlew bwcTestSuite -Dtests.security.manager=false -PcustomDistributionDownloadType=bundle`
If property `customDistributionDownloadType` is not set nor set to `bundle`, the BWC tests would be default to run in plugins level as previously configured. 

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/2870


kaituo (approved):

ohltyler:
Can this be targeted to `main` and backported to `2.x`?

build.gradle
```diff
@@ -384,7 +384,11 @@ task integTestRemote(type: RestIntegTestTask) {
     }
 }
 
-String bwcVersion = "1.1.0.0"
+String bwcMinVersion = "1.1.0.0"
+String bwcBundleVersion = "1.3.2.0"
```
ohltyler, line 388: Is the standard supposed to be the oldest version from the previous major version or the latest? IMO we should only need to persist one or the other rather than both here.

zelinh:
> Can this be targeted to `main` and backported to `2.x`? Or rather, will there be a separate PR with `2.x` BWC versions for `main`?

Yeah currently the distribution build for our 3.0.0 are failing and we don't have a stable bundle to test on. So I'm targeting to 2.5 version for now. I could draft a new PR to onboard on main when we getting close.

build.gradle
zelinh: Hi @ohltyler The reason I use 1.3.2 here is this is the earliest version that we start to support with our current URL pattern. For example, `https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.3.2/latest/linux/x64/tar/dist/opensearch/opensearch-1.3.2-linux-x64.tar.gz`. We introduced an additional path for different distribution types since then. 
I kept the old `1.1` for plugin level testing to minimize the effects of this PR on your original tests but I'm ok to change that to 1.3.2.0 as well.

ohltyler: Got it - I think it's fine to leave as-is. Just to confirm, in general, we want to use the earliest version, right? Like for example, in `main`, we should point to `2.0.0.0` as the BWC version used for upgrading to `3.x`?

ohltyler:
> > Can this be targeted to `main` and backported to `2.x`? Or rather, will there be a separate PR with `2.x` BWC versions for `main`?
> 
> Yeah currently the distribution build for our 3.0.0 are failing and we don't have a stable bundle to test on. So I'm targeting to 2.5 version for now. I could draft a new PR to onboard on main when we getting close.

Got it. I think as long as this is tracked on the infra side this is fine, thanks!

build.gradle
zelinh: Ideally yes I believe. In future we should have a matrix of versions and test each of the versions on upgrading and etc.. For now, i don't see any enforcement on the BWC version for `main`, it also depends on the core repo.

ohltyler (approved):

PR #770, opened 2023-01-05
Fix the discrepancy between Profile API and real time tasks API

kaituo:
### Description
This PR fixes the discrepancy by querying the result index when the total updates is less than 32. We have done similar things in profile API so I refactored reusable code to ProfileUtil. We also cached whether we have queried the result index and won't repeatedly issue the extra query.

Testing done:
1. repeated repro steps in https://github.com/opensearch-project/anomaly-detection/issues/502 and verified the issue has been resolved.
2. added unit tests.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/502

### DIFF
--- a/src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
+++ b/src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
@@ -356,9 +356,9 @@ public class ExecuteADResultResponseRecorder {
                     listener.onFailure(new AnomalyDetectionException(detectorId, "fail to get job"));
                     return;
                 }
-                adTaskCacheManager.markResultIndexQueried(detectorId);
+
                 ProfileUtil
-                    .confirmDetectorInitStatus(
+                    .confirmDetectorRealtimeInitStatus(
                         detectorOptional.get(),
                         jobOptional.get().getEnabledTime().toEpochMilli(),
                         client,
@@ -371,11 +371,13 @@ public class ExecuteADResultResponseRecorder {
                                     // so that the detector won't stay initialized
                                     correctedTotalUpdates = Long.valueOf(rcfMinSamples);
                                 }
+                                adTaskCacheManager.markResultIndexQueried(detectorId);
                                 return correctedTotalUpdates;
                             });
                         }, exception -> {
                             if (ExceptionUtil.isIndexNotAvailable(exception)) {
                                 // anomaly result index is not created yet
+                                adTaskCacheManager.markResultIndexQueried(detectorId);
                                 listener.onResponse(0L);
                             } else {
                                 listener.onFailure(exception);
--- a/src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
+++ b/src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
@@ -22,6 +22,8 @@ import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionListener;
+import org.opensearch.action.update.UpdateResponse;
+import org.opensearch.ad.common.exception.AnomalyDetectionException;
 import org.opensearch.ad.common.exception.EndRunException;
 import org.opensearch.ad.common.exception.ResourceNotFoundException;
 import org.opensearch.ad.constant.CommonErrorMessages;
@@ -32,6 +34,7 @@ import org.opensearch.ad.model.AnomalyResult;
 import org.opensearch.ad.model.DetectorProfileName;
 import org.opensearch.ad.model.FeatureData;
 import org.opensearch.ad.model.IntervalTimeConfiguration;
+import org.opensearch.ad.task.ADTaskCacheManager;
 import org.opensearch.ad.task.ADTaskManager;
 import org.opensearch.ad.transport.AnomalyResultResponse;
 import org.opensearch.ad.transport.ProfileAction;
@@ -40,10 +43,12 @@ import org.opensearch.ad.transport.RCFPollingAction;
 import org.opensearch.ad.transport.RCFPollingRequest;
 import org.opensearch.ad.transport.handler.AnomalyIndexHandler;
 import org.opensearch.ad.util.DiscoveryNodeFilterer;
+import org.opensearch.ad.util.ExceptionUtil;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.commons.authuser.User;
+import org.opensearch.search.SearchHits;
 import org.opensearch.threadpool.ThreadPool;
 
 public class ExecuteADResultResponseRecorder {
@@ -55,6 +60,9 @@ public class ExecuteADResultResponseRecorder {
     private DiscoveryNodeFilterer nodeFilter;
     private ThreadPool threadPool;
     private Client client;
+    private NodeStateManager nodeStateManager;
+    private ADTaskCacheManager adTaskCacheManager;
+    private int rcfMinSamples;
 
     public ExecuteADResultResponseRecorder(
         AnomalyDetectionIndices anomalyDetectionIndices,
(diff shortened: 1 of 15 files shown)


amitgalitz:
Why are we choosing to query the result index when the total updates is less than 32? Why this number?

kaituo:
> > Why are we choosing to query the result index when the total updates is less than 32? Why this number? I was thinking like your comment in the issue is that we should check the result index just when the updates are at 0.
> 
> I guess I see that this is based of minimum samples and its better to be safer rather than sorry and to keep querying logic up to the 32 updates instead of only at 0?

We only call initialization finished when the number of samples reach 32. So I chose 32. If there is any anomaly results, number of samples should be at least 32 and we can make initialization progress to be 100%.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -185,27 +199,66 @@ private void updateLatestRealtimeTask(
...
+                                detectorId,
+                                taskState,
+                                r,
+                                detectorIntervalInMinutes,
+                                error,
+                                listener
+                            ),
+                        e -> {
+                            log.error("Fail to confirm rcf update", e);
+                            adTaskManager
```
amitgalitz, line 239: Are we essentially going with the same `updateLatestRealtimeTaskOnCoordinatingNode` logic regardless if we can confirm RCF update? if we failed on something major like couldn't get detector, wouldn't this cause an issue

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -285,4 +338,51 @@ public void indexAnomalyResultException(
...
+                        jobOptional.get().getEnabledTime().toEpochMilli(),
+                        client,
+                        ActionListener.wrap(searchResponse -> {
+                            ActionListener.completeWith(listener, () -> {
+                                SearchHits hits = searchResponse.getHits();
+                                Long correctedTotalUpdates = rcfTotalUpdates;
+                                if (hits.getTotalHits().value > 0L) {
+                                    // correct the number if we have already had results after job enabling time
+                                    // so that the detector won't stay initialized
+                                    correctedTotalUpdates = Long.valueOf(rcfMinSamples);
```
amitgalitz, line 372: are we just going to manually set number of updates to 32 if there are some anomaly results even if technically it could be higher than 32 cause we have no real way of knowing the real number when cache is cleared?

amitgalitz:
Although nothing is broken right now should we remove the jackson-databind and jackson-annotations dependencies in this PR are do you want to add that into a separate PR if we are backporting this to way earlier OS versions.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: Yes. This is the old cold logic. If we failed on sth major like couldn't get detector, we would have failed earlier and there are logic to end detector run.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: yes, you are right

kaituo:
> Although nothing is broken right now should we remove the jackson-databind and jackson-annotations dependencies in this PR are do you want to add that into a separate PR if we are backporting this to way earlier OS versions.

Will do it in this PR as I don't see core has backported the change to earlier OS versions. Backporting to 2.x will be enough.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
amitgalitz: should we just return on exception here?

kaituo: I try to keep the original workflow even though failing to confirming rcf update. So I don't just return on exception.

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
```diff
@@ -285,4 +338,51 @@ public void indexAnomalyResultException(
...
+            if (!detectorOptional.isPresent()) {
+                listener.onFailure(new AnomalyDetectionException(detectorId, "fail to get detector"));
+                return;
+            }
+            nodeStateManager.getAnomalyDetectorJob(detectorId, ActionListener.wrap(jobOptional -> {
+                if (!jobOptional.isPresent()) {
+                    listener.onFailure(new AnomalyDetectionException(detectorId, "fail to get job"));
+                    return;
+                }
+                adTaskCacheManager.markResultIndexQueried(detectorId);
```
ylwu-amzn, line 359: Should we mark result index queried after `confirmDetectorInitStatus` done?

amitgalitz (approved):

src/main/java/org/opensearch/ad/ProfileUtil.java
```diff
@@ -0,0 +1,57 @@
...
+public class ProfileUtil {
+    /**
+     * Create search request to check if we have at least 1 anomaly score larger than 0 after AD job enabled time
+     * @param detectorId detector id
+     * @param enabledTime the time when AD job is enabled in milliseconds
+     * @return the search request
+     */
+    private static SearchRequest createInittedEverRequest(String detectorId, long enabledTime, String resultIndex) {
+        BoolQueryBuilder filterQuery = new BoolQueryBuilder();
+        filterQuery.filter(QueryBuilders.termQuery(AnomalyResult.DETECTOR_ID_FIELD, detectorId));
```
ylwu-amzn, line 34: Historical analysis result also stored in result index, which has non-null `task_id`. For realtime detection result, we should filter `task_id == null`

src/main/java/org/opensearch/ad/ProfileUtil.java
```diff
@@ -0,0 +1,57 @@
...
+        filterQuery.filter(QueryBuilders.rangeQuery(AnomalyResult.EXECUTION_END_TIME_FIELD).gte(enabledTime));
+        filterQuery.filter(QueryBuilders.rangeQuery(AnomalyResult.ANOMALY_SCORE_FIELD).gt(0));
+
+        SearchSourceBuilder source = new SearchSourceBuilder().query(filterQuery).size(1);
+
+        SearchRequest request = new SearchRequest(CommonName.ANOMALY_RESULT_INDEX_ALIAS);
```
ylwu-amzn, line 40: Should we also consider custom anomaly result index use case?

src/main/java/org/opensearch/ad/ExecuteADResultResponseRecorder.java
kaituo: yes, changed

src/main/java/org/opensearch/ad/ProfileUtil.java
kaituo: good callout. Changed.

src/main/java/org/opensearch/ad/ProfileUtil.java
kaituo: We did consider result index. Please check the new few lines.

amitgalitz (approved):

ylwu-amzn (dismissed):

amitgalitz:
BWC failures might be cause Opensearch bumped to 2.6

kaituo:
Bwc test failed:

> `cluster{::adBwcCluster0}` failed to wait for cluster health yellow after 40 SECONDS
    503 Service Unavailable

latest development: https://github.com/opensearch-project/OpenSearch/issues/5091

kaituo:
> BWC failures might be cause Opensearch bumped to 2.6

ah, that should be the reason.

amitgalitz:
> > BWC failures might be cause Opensearch bumped to 2.6
> 
> ah, that should be the reason.

I'm okay merging this if the other tests pass and bumping main branch bwc to 2.6 on another PR since that doesn't need to be backported to 2.6 and 2.x. Or add it here and change the backport PR manually

kaituo:
> > > BWC failures might be cause Opensearch bumped to 2.6
> > 
> > 
> > ah, that should be the reason.
> 
> I'm okay merging this if the other tests pass and bumping main branch bwc to 2.6 on another PR since that doesn't need to be backported to 2.6 and 2.x. Or add it here and change the backport PR manually

I will do it in another PR. Sean asked me to finish the jackson removal in 2.5 before 2 pm today.

kaituo:
@amitgalitz @ylwu-amzn created a new PR for bwc version bump: https://github.com/opensearch-project/anomaly-detection/pull/775

ylwu-amzn (approved):

PR #775, opened 2023-01-09
bump bwc to 2.6.0

kaituo:
### Description
We need to bump the version because only the latest 2.x will be upgradable to 3.0 when that ships. Since OpenSearch core bumped to 2.6, we need to do that as well. Note after bumping version, bwc still fails to find 2.6.0 core zip.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ buildscript {
         js_resource_folder = "src/test/resources/job-scheduler"
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
         job_scheduler_version = System.getProperty("job_scheduler.version", opensearch_build)
-        bwcVersionShort = "2.5.0"
+        bwcVersionShort = "2.6.0"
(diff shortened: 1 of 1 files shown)


ohltyler:
Seems this should be a requirement of #774 workflow when bumping minor versions. Can you open an issue in [opensearch-build](https://github.com/opensearch-project/opensearch-build) so this may be automatically done in the future? I guess it would need to be a separate PR to change this in `main` branch

kaituo:
> Seems this should be a requirement of #774 workflow when bumping minor versions. Can you open an issue in [opensearch-build](https://github.com/opensearch-project/opensearch-build) so this may be automatically done in the future? I guess it would need to be a separate PR to change this in `main` branch

created an issue in opensearch-build: https://github.com/opensearch-project/opensearch-build/issues/3063

amitgalitz (approved):

ylwu-amzn (approved):

Issue #788, opened 2023-01-26
[BUG] fixing dls/fls logic around numeric aggregations


kaituo:
**What is the bug?**
fixing how we apply dls/fls logic around numeric aggregations

kaituo:
fixed in https://github.com/opensearch-project/anomaly-detection/pull/789

PR #789, opened 2023-01-26
fixing dls/fls logic around numeric aggregations

kaituo:

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/788

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
@@ -57,6 +57,7 @@ import org.opensearch.ad.transport.ProfileAction;
 import org.opensearch.ad.transport.ProfileRequest;
 import org.opensearch.ad.transport.handler.AnomalyIndexHandler;
 import org.opensearch.ad.util.DiscoveryNodeFilterer;
+import org.opensearch.ad.util.SecurityUtil;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.settings.Settings;
@@ -76,7 +77,6 @@ import org.opensearch.jobscheduler.spi.utils.LockService;
 import org.opensearch.threadpool.ThreadPool;
 
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
 
 /**
  * JobScheduler will call AD job runner to get anomaly result periodically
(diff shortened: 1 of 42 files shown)


amitgalitz:
Weird that it says it can't get the zip for job scheduler but even the link in the error message works fine to download job-scheduler-1.3.8.0

kaituo:
> Weird that it says it can't get the zip for job scheduler but even the link in the error message works fine to download job-scheduler-1.3.8.0

Indeed.

amitgalitz (approved):

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -218,28 +218,11 @@ protected void runAdJob(
...
-            // It's possible that user create domain with security disabled, then enable security
-            // after upgrading. This is for BWC, for old detectors which created when security
-            // disabled, the user will be null.
-            user = "";
-            roles = settings.getAsList("", ImmutableList.of("all_access", "AmazonES_all_access"));
-        } else {
-            user = jobParameter.getUser().getName();
-            roles = jobParameter.getUser().getRoles();
-        }
+        User userInfo = SecurityUtil.getUserFromJob(jobParameter, settings);
```
ylwu-amzn, line 221: Is this ready user information stored in AD job? I think we have no way to get user information from security plugin. Just asking to learn more

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -156,10 +156,7 @@
 import org.opensearch.ad.transport.handler.AnomalyIndexHandler;
 import org.opensearch.ad.transport.handler.AnomalyResultBulkIndexHandler;
 import org.opensearch.ad.transport.handler.MultiEntityResultHandler;
-import org.opensearch.ad.util.ClientUtil;
-import org.opensearch.ad.util.DiscoveryNodeFilterer;
-import org.opensearch.ad.util.IndexUtils;
-import org.opensearch.ad.util.Throttler;
+import org.opensearch.ad.util.*;
```
ylwu-amzn, line 159: minor: should we avoid using import *?

ylwu-amzn (approved):

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
kaituo: Yes, it is from AD job.

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
kaituo: Yes, will fix in other PRs since 1.3.8 is due today.

PR #372, opened 2022-02-04
Use admin client for system index updates in integ tests

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description
Fixes 3 tests that were using the default client to perform manual changes to system indices, such as checking if they exist, creating them, deleting them. Note that this is necessary sometimes to put the cluster in a certain state for testing out certain scenarios. To properly perform these actions, we can use the existing admin client.

Advantages of admin client:
- overrides any strict deprecation failures related to direct access to system indices
- overrides any permissions failures related to security-enabled cluster

More context behind these changes:
We were seeing occasional failures when trying to execute these tests against certain cluster configurations, including bundled binary clusters with security enabled. By changing to use the admin client for certain sensitive requests, these failures don't occur. 
### Issues Resolved
Closes #278 
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #353, opened 2022-01-07
Remove Beta label from bug issue template

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description
Removes outdated `Beta` label
 
### Issues Resolved
Closes #342
 
### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #792, opened 2023-01-28
[Feature/extensions] Migrate Create, Update, and Validate actions to extension

dbwiddis:
### Description

Migrates the `IndexAnomalyDetectorTransportAction`, `ValidateAnomalyDetectorTransportAction`, their Rest Handlers, and other supporting classes from plugin to extension.

Testing requires adding a document to an index which requires working around [#5999](https://github.com/opensearch-project/OpenSearch/issues/5999) locally.

See [SDK #372](https://github.com/opensearch-project/opensearch-sdk-java/pull/372) for a summary of the changes that you'll see in the diff.

### Issues Resolved

Fixes [SDK #353](https://github.com/opensearch-project/opensearch-sdk-java/issues/353).


minalsha:
just a small suggestion: is it possible to break this huge PR into small PRs for quic review?

dbwiddis:
> just a small suggestion: is it possible to break this huge PR into small PRs for quic review?

Unfortunately not without having a breaking change.  However, there really aren't as many changes as you might think. 

[SDK #372](https://github.com/opensearch-project/opensearch-sdk-java/pull/372) committed a migration guide which pretty much summarizes the diff here.   I'll be a bit more specific if it helps.  From top down on files:
 - build.gradle: had to suppress some jacoco exclusions due to test coverage and add dependency on rest high level client
 - ADExtension: add path to the new actions, and add getter for the rest client wrapper
 - ADPlugin: (UNUSED, Reference) mostly whitespace changes, setting client and clusterservice to null in objects whose class types have changed
 - Remaining classes mostly change `Client` to `SDKRestClient`, `ClusterService` to `SDKClusterService`, and change the imports for `Request` and `Response` objects.   Exceptions follow:
    - Settings update consumers use a more efficient map approach
    - AnomalyDetectionIndices rollover request required new syntax for createRequest mapping and deleteIndex iteration
    - AbstractAnomalyDetectorSDKAction should actually be deleted, oops, it was from an earlier iteration
    - RestIndex and RestValidate DetectorActions are new format for extension rest actions, mostly self explanatory and indicated in migration doc
    - TransportActions doExecute made public
    - Test, a lot of things suppressed pending completion of createcomponents work next sprint

ryanbogan (approved):

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
```diff
@@ -114,9 +116,15 @@ public SearchFeatureDao(
...
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_ENTITIES_FOR_PREVIEW, it -> this.maxEntitiesForPreview = it);
         this.pageSize = pageSize;
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(PAGE_SIZE, it -> this.pageSize = it);
+        try {
+            Map<Setting<?>, Consumer<?>> settingsUpdateConsumers = new HashMap<>();
+            settingsUpdateConsumers.put(MAX_ENTITIES_FOR_PREVIEW, it -> this.maxEntitiesForPreview = (int) it);
+            settingsUpdateConsumers.put(PAGE_SIZE, it -> this.pageSize = (int) it);
+            clusterService.getClusterSettings().addSettingsUpdateConsumer(settingsUpdateConsumers);
+        } catch (Exception e) {
+            // TODO Handle this
```
owaiskazi19, line 126: [handleAddSettingsUpdateConsumerRequest](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/extensions/AddSettingsUpdateConsumerRequestHandler.java#L60) throws a generic exception here. Can we create an issue to make it more specific so as to handle it here later?

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -238,8 +235,8 @@ public Collection<Object> createComponents(
         */
         // AnomalyDetectionIndices is Injected for IndexAnomalyDetectorTrasnportAction constructor
         this.anomalyDetectionIndices = new AnomalyDetectionIndices(
-            client,
-            clusterService,
+            null, // client,
+            null, // clusterService,
```
owaiskazi19, line 239: For my understanding, this will be change to SDKRestClient after the guice injection work is done for createComponent, right?

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -161,26 +159,26 @@ class IndexState {
...
     ) {
-        this.client = client;
-        this.adminClient = client.admin();
-        this.clusterService = clusterService;
+        this.client = restClient;
+        this.adminClient = restClient;
+        this.clusterService = sdkClusterService;
         this.threadPool = threadPool;
-        this.clusterService.addLocalNodeMasterListener(this);
+        // this.clusterService.addLocalNodeMasterListener(this);
```
owaiskazi19, line 181: We should create an issue to create a transport method for this listener as this class implements `LocalNodeMasterListener`.

src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java
```diff
@@ -53,26 +64,59 @@ public class RestIndexAnomalyDetectorAction extends AbstractAnomalyDetectorActio
 
     private static final String INDEX_ANOMALY_DETECTOR_ACTION = "index_anomaly_detector_action";
     private final Logger logger = LogManager.getLogger(RestIndexAnomalyDetectorAction.class);
-
-    public RestIndexAnomalyDetectorAction(Settings settings, ClusterService clusterService) {
-        super(settings, clusterService);
```
owaiskazi19, line 58: Rather than changing the arguments here of super call, is there any way we can keep the arguments 
for settings -> environmentSettings
clusterService -> sdkClusterService

src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java
```diff
@@ -97,57 +141,70 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
-                }
-                return bytesRestResponse;
-            }
-        };
+    private ExtensionRestResponse indexAnomalyDetectorResponse(ExtensionRestRequest request, IndexAnomalyDetectorResponse response)
+        throws IOException {
+        RestStatus restStatus = RestStatus.CREATED;
+        if (request.method() == RestRequest.Method.PUT) {
+            restStatus = RestStatus.OK;
+        } else {
```
owaiskazi19, line 196: Is this else just for debugging purpose?

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
```diff
@@ -117,33 +150,89 @@ protected BaseRestHandler.RestChannelConsumer prepareRequest(RestRequest request
...
+                DetectorValidationIssue issue = new DetectorValidationIssue(
+                    ADException.getAspect(),
+                    ADException.getType(),
+                    ADException.getMessage()
+                );
+                return new ExtensionRestResponse(
+                    request,
+                    RestStatus.OK,
+                    new ValidateAnomalyDetectorResponse(issue).toXContent(JsonXContent.contentBuilder())
+                );
```
owaiskazi19, line 168: Can we move line 164-168 under `sendAnomalyDetectorValidationParseResponse` method?

src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java
```diff
@@ -664,7 +671,11 @@ protected void validateCategoricalField(String detectorId, boolean indexingDryRu
             listener.onFailure(new IllegalArgumentException(message));
         });
 
-        client.execute(GetFieldMappingsAction.INSTANCE, getMappingsRequest, mappingsListener);
+        // FIXME Need to implement this; does shard level actions on the cluster
+        // https://github.com/opensearch-project/opensearch-sdk-java/issues/361
+        // client.execute(GetFieldMappingsAction.INSTANCE, getMappingsRequest, mappingsListener);
+        // For now just skip and go to the next step:
+        searchAdInputIndices(detectorId, indexingDryRun);
```
owaiskazi19, line 678: Liked the way validation is handle here with granular details in the comments.

src/main/java/org/opensearch/ad/util/ParseUtils.java
```diff
@@ -517,13 +518,16 @@ public static void getDetector(
...
         ActionListener listener,
         Consumer<AnomalyDetector> function,
-        Client client,
-        ClusterService clusterService,
+        SDKRestClient client,
+        SDKClusterService clusterService,
         NamedXContentRegistry xContentRegistry,
         boolean filterByBackendRole
     ) {
+        logger.info("in getDetector()");
```
owaiskazi19, line 526: Debug statement?

src/main/java/org/opensearch/ad/util/ParseUtils.java
```diff
@@ -545,6 +549,7 @@ public static void getDetector(
                         )
                 );
         } else {
+            logger.info("Index not found: {}", AnomalyDetector.ANOMALY_DETECTORS_INDEX);
```
owaiskazi19, line 552: Do we need these log statements?

owaiskazi19 (changes requested):
Looks good overall with few changes and questions! Thanks for working on this @dbwiddis!

src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java
dbwiddis: This `Plugin` class will actually go away.  It's just here for reference until we finish all the extension point implementation.

I'm not actually sure why this has a regular constructor and not an injected one, perhaps it's the primitive int parameter.

A version of this will be created in the `AnomalyDetectorExtension` class with the correct SDK classes, and bound to itself as a component so it can be injected where needed.

src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java
dbwiddis: Yeah, my TODO was kind of thinking it's better to handle these upstream, I just haven't gotten that far to thinking about how.  I'll make an issue.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
dbwiddis: I'm under the impression this isn't needed at all.  But I'm happy to create an issue to determine whether to make it a no-op.

src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java
dbwiddis: Sure, we could use environment settings, but we have to "change the argument" for SDKClusterService anyway.  Both of those are available on the ExtensionsRunner, which we've somewhat started to standardize passing (along with the extension itself) as arguments to the rest handler constructors, to give us access to all the other things.

Once we have create components we may not need all the other things, though.

Definitely something to think about as I integrate create components.

src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java
dbwiddis: Yes and no.  Early on I was getting timeouts on my OpenSearch rest requests, but it was successfully creating the detector, but I didn't have the ID.   But it seems useful to add the detector ID to the logs for the CREATED case.   So I added it for debugging but think it's useful to keep.

src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java
dbwiddis: Uh, lines 164-168 are the replacement for `sendAnomalyDetectorValidationParseResponse` which should probably be deleted.

Unless by "under" you mean, change this to `return sendAnomalyDetectorValidationParseResponse(issue)` and move the 3 lines into that method?  Sure.

src/main/java/org/opensearch/ad/util/ParseUtils.java
dbwiddis: You found my breadcrumbs!

src/main/java/org/opensearch/ad/util/ParseUtils.java
dbwiddis: More breadcrumbs to remove.

vibrantvarun (approved):

owaiskazi19 (approved):

dbwiddis:
> Trying to understand why these many files are in jacoco exclusions?

We've commented out a lot of the unit tests until the functionality is migrated.  Goal is to re-enable them when migration is closer to complete (and we've uncommented/modified the tests)

PR #837, opened 2023-03-07
[Feature/extensions] Migrates Start Detector

joshpalis:
### Description
Migrates the AnomalyDetectorJobTransportAction, RestAnomalyDetectorJobAction,  and other supporting classes from plugin to extension.

### Issues Resolved
 https://github.com/opensearch-project/opensearch-sdk-java/issues/383


src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
```diff
@@ -120,4 +216,5 @@ public List<ReplacedRoute> replacedRoutes() {
                 )
             );
     }
+    */
```
owaiskazi19, line 219: Do we need this anymore since we have added routes already?

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
```diff
@@ -16,60 +16,138 @@
...
+        }
+
         String detectorId = request.param(DETECTOR_ID);
         long seqNo = request.paramAsLong(IF_SEQ_NO, SequenceNumbers.UNASSIGNED_SEQ_NO);
         long primaryTerm = request.paramAsLong(IF_PRIMARY_TERM, SequenceNumbers.UNASSIGNED_PRIMARY_TERM);
-        boolean historical = request.paramAsBoolean("historical", false);
-        String rawPath = request.rawPath();
+        // Passed false until historical analysis workflow is enabled
+        boolean historical = false;
+        String rawPath = request.path();
```
owaiskazi19, line 150: Why not? 
```suggestion
        String rawPath = request.rawPath();
```

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
```diff
@@ -16,60 +16,138 @@
...
+            registerJobDetails();
+        }
+
         String detectorId = request.param(DETECTOR_ID);
         long seqNo = request.paramAsLong(IF_SEQ_NO, SequenceNumbers.UNASSIGNED_SEQ_NO);
         long primaryTerm = request.paramAsLong(IF_PRIMARY_TERM, SequenceNumbers.UNASSIGNED_PRIMARY_TERM);
-        boolean historical = request.paramAsBoolean("historical", false);
-        String rawPath = request.rawPath();
+        // Passed false until historical analysis workflow is enabled
+        boolean historical = false;
```
owaiskazi19, line 149: This should take the default value as `false`
```suggestion
        boolean historical =  request.paramAsBoolean("historical", false);
```

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
```diff
@@ -16,60 +16,138 @@
...
+        LOG.info("Job Details Registered : " + jobDetailsAreRegistered);
+    }
+
+    protected ExtensionRestResponse prepareRequest(RestRequest request) throws IOException {
         if (!EnabledSetting.isADPluginEnabled()) {
             throw new IllegalStateException(CommonErrorMessages.DISABLED_ERR_MSG);
         }
 
+        // Ensure job details are registered with Job Scheduler prior to creating a job
+        if (!jobDetailsAreRegistered) {
```
owaiskazi19, line 141: For my understanding. How did we ensure that job details are registered before in AD plugin?

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -127,7 +125,8 @@ public class ADBatchTaskRunner {
     private final FeatureManager featureManager;
     private final ADCircuitBreakerService adCircuitBreakerService;
     private final ADTaskManager adTaskManager;
-    private final AnomalyResultBulkIndexHandler anomalyResultBulkIndexHandler;
+    // @anomaly-detection - commented until we have support for SDKRestClient.prepareBulk()
```
owaiskazi19, line 128: Do we an issue to track this?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1288,25 +1298,29 @@ private <T> void resetTaskStateAsStopped(
...
-        updateByQueryRequest.setRefresh(true);
+
         String script = String.format(Locale.ROOT, "ctx._source.%s='%s';", STATE_FIELD, ADTaskState.STOPPED.name());
-        updateByQueryRequest.setScript(new Script(script));
+        UpdateByQueryRequest updateByQueryRequest = new UpdateByQueryRequest.Builder().index(DETECTION_STATE_INDEX).query(queryBuilder -> {
+            queryBuilder.match(mb -> mb.field(PARENT_TASK_ID_FIELD).query(vb -> vb.stringValue(detectorTaskId)));
+            queryBuilder.match(mb -> mb.field(TASK_TYPE_FIELD).query(vb -> vb.stringValue(ADTaskType.HISTORICAL_HC_ENTITY.name())));
+            for (String notEndedStates : NOT_ENDED_STATES) {
+                queryBuilder.match(mb -> mb.field(STATE_FIELD).query(vb -> vb.stringValue(notEndedStates)));
+            }
```
owaiskazi19, line 1308: Can we use`matchAll` instead of a for loop here?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1288,25 +1298,29 @@ private <T> void resetTaskStateAsStopped(
...
-        BoolQueryBuilder query = new BoolQueryBuilder();
-        query.filter(new TermQueryBuilder(PARENT_TASK_ID_FIELD, detectorTaskId));
-        query.filter(new TermQueryBuilder(TASK_TYPE_FIELD, ADTaskType.HISTORICAL_HC_ENTITY.name()));
-        query.filter(new TermsQueryBuilder(STATE_FIELD, NOT_ENDED_STATES));
-        updateByQueryRequest.setQuery(query);
-        updateByQueryRequest.setRefresh(true);
+
         String script = String.format(Locale.ROOT, "ctx._source.%s='%s';", STATE_FIELD, ADTaskState.STOPPED.name());
-        updateByQueryRequest.setScript(new Script(script));
+        UpdateByQueryRequest updateByQueryRequest = new UpdateByQueryRequest.Builder().index(DETECTION_STATE_INDEX).query(queryBuilder -> {
```
owaiskazi19, line 1303: Instead of using queryBuilder in the lambda func. Can we create an object of `BoolQueryBuilder` and set query to it?

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1478,21 +1492,23 @@ private void updateLatestFlagOfOldTasksAndCreateNewTask(
...
-        BoolQueryBuilder query = new BoolQueryBuilder();
-        query.filter(new TermQueryBuilder(DETECTOR_ID_FIELD, detector.getDetectorId()));
-        query.filter(new TermQueryBuilder(IS_LATEST_FIELD, true));
-        // make sure we reset all latest task as false when user switch from single entity to HC, vice versa.
-        query.filter(new TermsQueryBuilder(TASK_TYPE_FIELD, taskTypeToString(getADTaskTypes(detectionDateRange, true))));
-        updateByQueryRequest.setQuery(query);
-        updateByQueryRequest.setRefresh(true);
         String script = String.format(Locale.ROOT, "ctx._source.%s=%s;", IS_LATEST_FIELD, false);
-        updateByQueryRequest.setScript(new Script(script));
+        UpdateByQueryRequest updateByQueryRequest = new UpdateByQueryRequest.Builder().index(DETECTION_STATE_INDEX).query(queryBuilder -> {
```
owaiskazi19, line 1496: Same comment as above `BoolQueryBuilder`

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1478,21 +1492,23 @@ private void updateLatestFlagOfOldTasksAndCreateNewTask(
...
+            queryBuilder.match(mb -> mb.field(DETECTOR_ID_FIELD).query(vb -> vb.stringValue(detector.getDetectorId())));
+            queryBuilder.match(mb -> mb.field(IS_LATEST_FIELD).query(vb -> vb.booleanValue(true)));
+            // make sure we reset all latest task as false when user switch from single entity to HC, vice versa.
+            for (String taskTypes : taskTypeToString(getADTaskTypes(detectionDateRange, true))) {
+                queryBuilder.match(mb -> mb.field(TASK_TYPE_FIELD).query(vb -> vb.stringValue(taskTypes)));
```
owaiskazi19, line 1501: Same comment as above for `matchAll`

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1728,18 +1756,33 @@ public void cleanChildTasksAndADResultsOfDeletedTask() {
...
-                DeleteByQueryRequest deleteChildTasksRequest = new DeleteByQueryRequest(DETECTION_STATE_INDEX);
-                deleteChildTasksRequest.setQuery(new TermsQueryBuilder(PARENT_TASK_ID_FIELD, taskId));
+                DeleteByQueryRequest deleteChildTasksRequest = new DeleteByQueryRequest.Builder()
+                    .index(DETECTION_STATE_INDEX)
+                    .query(queryBuilder -> queryBuilder.match(mb -> mb.field(PARENT_TASK_ID_FIELD).query(vb -> vb.stringValue(taskId))))
+                    .build();
+                try {
+                    CompletableFuture<DeleteByQueryResponse> deleteChildTasksResponse = sdkJavaAsyncClient
+                        .deleteByQuery(deleteChildTasksRequest);
+                    DeleteByQueryResponse r = deleteChildTasksResponse.orTimeout(10L, TimeUnit.SECONDS).get();
```
owaiskazi19, line 1775:
```suggestion
                    DeleteByQueryResponse deleteByQueryResponse = deleteChildTasksResponse.orTimeout(10L, TimeUnit.SECONDS).get();
```

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1896,29 +1939,27 @@ public ADTaskCancellationState cancelLocalTaskByDetectorId(String detectorId, St
...
-
-        BoolQueryBuilder query = new BoolQueryBuilder();
-        query.filter(new TermQueryBuilder(DETECTOR_ID_FIELD, detectorId));
+        DeleteByQueryRequest request = new DeleteByQueryRequest.Builder()
+            .index(DETECTION_STATE_INDEX)
+            .query(queryBuilder -> queryBuilder.match(mb -> mb.field(DETECTOR_ID_FIELD).query(vb -> vb.stringValue(detectorId))))
+            .build();
+        try {
+            CompletableFuture<DeleteByQueryResponse> deleteByQueryResponse = sdkJavaAsyncClient.deleteByQuery(request);
+            DeleteByQueryResponse r = deleteByQueryResponse.orTimeout(10L, TimeUnit.SECONDS).get();
```
owaiskazi19, line 1948:
```suggestion
            DeleteByQueryResponse response = deleteByQueryResponse.orTimeout(10L, TimeUnit.SECONDS).get();
```

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1927,18 +1968,19 @@ private void deleteADResultOfDetector(String detectorId) {
...
-                    })
-            );
+        DeleteByQueryRequest deleteADResultsRequest = new DeleteByQueryRequest.Builder()
+            .index(ALL_AD_RESULTS_INDEX_PATTERN)
+            .query(queryBuilder -> queryBuilder.match(mb -> mb.field(DETECTOR_ID_FIELD).query(vb -> vb.stringValue(detectorId))))
+            .build();
+
+        try {
+            CompletableFuture<DeleteByQueryResponse> deleteADResultsResponse = sdkJavaAsyncClient.deleteByQuery(deleteADResultsRequest);
+            DeleteByQueryResponse r = deleteADResultsResponse.orTimeout(10L, TimeUnit.SECONDS).get();
```
owaiskazi19, line 1978:
```suggestion
            DeleteByQueryResponse deleteByQueryResponse = deleteADResultsResponse.orTimeout(10L, TimeUnit.SECONDS).get();
```

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -2912,33 +2954,42 @@ public void resetLatestFlagAsFalse(List<ADTask> adTasks) {
...
+
                 task.setLatest(false);
                 task.setLastUpdateTime(Instant.now());
-                IndexRequest indexRequest = new IndexRequest(DETECTION_STATE_INDEX)
-                    .id(task.getTaskId())
-                    .source(task.toXContent(XContentBuilder.builder(XContentType.JSON.xContent()), XCONTENT_WITH_TYPE));
-                bulkRequest.add(indexRequest);
+
+                BulkOperation op = new BulkOperation.Builder()
+                    .index(b -> b.index(DETECTION_STATE_INDEX).id(task.getTaskId()).document(task))
```
owaiskazi19, line 2965: Another benefit of using JavaClient. No need to parse XContent. We can parse the model object directly!

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -2912,33 +2954,42 @@ public void resetLatestFlagAsFalse(List<ADTask> adTasks) {
...
+        BulkRequest bulkRequest = new BulkRequest.Builder().operations(operations).refresh(Refresh.True).build();
+
+        try {
+            CompletableFuture<BulkResponse> bulkResponse = sdkJavaAsyncClient.bulk(bulkRequest);
+            BulkResponse res = bulkResponse.orTimeout(10L, TimeUnit.SECONDS).get();
+
+            List<BulkResponseItem> bulkItemResponses = res.items();
+            if (bulkItemResponses != null && bulkItemResponses.size() > 0) {
+                for (BulkResponseItem bulkItemResponse : bulkItemResponses) {
+                    if (bulkItemResponse.error() == null) {
```
owaiskazi19, line 2982: Is `error` of Java Client equivalent to `isFailed` for RestClient?

src/main/java/org/opensearch/ad/transport/ADJobRunnerTransportAction.java
```diff
@@ -14,66 +14,65 @@
...
             if (jobParameterDocumentId == null || jobParameterDocumentId.isEmpty()) {
                 listener.onFailure(new IllegalArgumentException("jobParameterDocumentId cannot be empty or null"));
             } else {
+                CompletableFuture<AnomalyDetectorJob> inProgressFuture = new CompletableFuture<>();
                 findById(jobParameterDocumentId, new ActionListener<>() {
                     @Override
                     public void onResponse(AnomalyDetectorJob anomalyDetectorJob) {
-                        scheduledJobParameter = anomalyDetectorJob;
-                        inProgressFuture.complete(scheduledJobParameter);
+                        inProgressFuture.complete(anomalyDetectorJob);
```
owaiskazi19, line 75: Don't we have to set `scheduledJobParameter = anomalyDetectorJob;` here?

src/main/java/org/opensearch/ad/transport/ADJobRunnerTransportAction.java
```diff
@@ -95,41 +107,24 @@ public void onFailure(Exception e) {
...
     }
 
     private void findById(String jobParameterId, ActionListener<AnomalyDetectorJob> listener) {
         GetRequest getRequest = new GetRequest(AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX, jobParameterId);
-
         try {
             client.get(getRequest, ActionListener.wrap(response -> {
                 if (!response.isExists()) {
-                    listener.onResponse(null);
+                    listener.onFailure(null);
```
owaiskazi19, line 122:
```suggestion
                    listener.onResponse(null);
```

build.gradle
```diff
@@ -743,7 +744,32 @@ List<String> jacocoExclusions = [
...
+        'org.opensearch.ad.ratelimit.QueuedRequest',
+        'org.opensearch.ad.ratelimit.CheckpointReadWorker',
+        'org.opensearch.ad.ratelimit.ConcurrentWorker',
+        'org.opensearch.ad.ratelimit.EntityFeatureRequest',
+        'org.opensearch.ad.ratelimit.RateLimitedRequestWorker.RequestQueue',
+        'org.opensearch.ad.stats.StatNames',
+        'org.opensearch.ad.stats.ADStats',
+        'org.opensearch.ad.MaintenanceState',
+        'org.opensearch.ad.AnomalyDetectorExtension.1',
+        'org.opensearch.ad.AnomalyDetectorExtension.1.1'
```
owaiskazi19, line 772: Are the tests commented for these files?

owaiskazi19 (changes requested):
Great work with the migration and integrating job scheduler! Good to see the workflow.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -243,22 +229,24 @@ protected void runAdJob(
         }
         String resultIndex = jobParameter.getResultIndex();
         if (resultIndex == null) {
-            runAnomalyDetectionJob(jobParameter, lockService, lock, detectionStartTime, executionStartTime, detectorId, user, roles);
+            // TODO : https://github.com/opensearch-project/opensearch-sdk-java/issues/626
+            // runAnomalyDetectionJob(jobParameter, lock, detectionStartTime, executionStartTime, detectorId, user, roles);
```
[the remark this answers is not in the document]
joshpalis: This PR focuses on the configuration and facilitation of the job execution and ensures that the required metadata indices are created. The commented method invokes the `AnomalyResultTransportAction`, which validates the configuration of the anomaly detector before invoking the `RCFResultAction`. The `RCFResultTransportAction` does the actual detection work, and will be handled as part of https://github.com/opensearch-project/opensearch-sdk-java/issues/626  and returns the anomaly result as a response.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -685,15 +631,51 @@ private void updateLatestRealtimeTask(
...
+            boolean lockIsReleased = RestStatus.fromCode(releaseLockResponse.getStatusLine().getStatusCode()) == RestStatus.OK
+                ? true
+                : false;
+
+            if (lockIsReleased) {
+                log.info("Released lock for AD job {}", jobParameter.getName());
+            } else {
+                throw new Exception("Failed to release lock");
+            }
+        } catch (Exception exception) {
```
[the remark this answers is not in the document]
joshpalis: Sure ill change this to IOException

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -605,33 +618,32 @@ public void initAnomalyResultIndexDirectly(
...
-    // );
-    // adminClient.indices().create(request, markMappingUpToDate(ADIndex.JOB, actionListener));
-    // } catch (IOException e) {
-    // logger.error("Fail to init AD job index", e);
-    // actionListener.onFailure(e);
-    // }
-    // }
+    public void initAnomalyDetectorJobIndex(ActionListener<CreateIndexResponse> actionListener) {
+        try {
+            CreateIndexRequest request = new CreateIndexRequest(".opendistro-anomaly-detector-jobs")
```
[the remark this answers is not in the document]
joshpalis: I would normally agree however this is how the method was originally. For the sake of reducing the diff size, I think its best to keep it as is

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -436,44 +441,48 @@ public void deleteModelCheckpointByDetectorId(String detectorID) {
...
-            .setAbortOnVersionConflict(false) // when current delete happens, previous might not finish.
-                                              // Retry in this case
-            .setRequestsPerSecond(500); // throttle delete requests
+        DeleteByQueryRequest deleteRequest = new DeleteByQueryRequest.Builder()
+            .index(CommonName.CHECKPOINT_INDEX_NAME)
+            .query(queryBuilder -> queryBuilder.match(mb -> mb.field(DETECTOR_ID).query(vb -> vb.stringValue(detectorID))))
+            .allowNoIndices(true)
+            .ignoreUnavailable(true)
+            .expandWildcards(ExpandWildcard.Open)
+            .conflicts(Conflicts.Proceed)// when current delete happens, previous might not finish.
```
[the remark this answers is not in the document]
joshpalis: the original indices options used for this request was `IndicesOptions.LENIENT_EXPAND_OPEN`, which includes the following 
```
public static final IndicesOptions LENIENT_EXPAND_OPEN = new IndicesOptions(
        EnumSet.of(Option.ALLOW_NO_INDICES, Option.IGNORE_UNAVAILABLE),
        EnumSet.of(WildcardStates.OPEN)
    );
```

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -436,44 +441,48 @@ public void deleteModelCheckpointByDetectorId(String detectorID) {
...
+                                         // Retry in this case
+            .requestsPerSecond(500L)// throttle delete request
+            .build();
         logger.info("Delete checkpoints of detector {}", detectorID);
-        client.execute(DeleteByQueryAction.INSTANCE, deleteRequest, ActionListener.wrap(response -> {
-            if (response.isTimedOut() || !response.getBulkFailures().isEmpty() || !response.getSearchFailures().isEmpty()) {
+        try {
+            CompletableFuture<DeleteByQueryResponse> deleteResponse = sdkJavaAsyncClient.deleteByQuery(deleteRequest);
+            DeleteByQueryResponse response = deleteResponse.orTimeout(10L, TimeUnit.SECONDS).get();
+            if (response.timedOut() || !response.failures().isEmpty()) {
```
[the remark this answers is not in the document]
joshpalis: I had a conversation with Vacha regarding the handling of `DeleteByQueryResponse` failures using the java client. From that conversation, there are no bulk or search failures, only failures of type `BulkIndexByScrollFailure`.

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
joshpalis: I just kept it for reference for the stop detector workflow. But I can remove it and just add the stop detector route handler under `routeHandlers` and comment it out for now.

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
joshpalis: For plugins, job details are registered during the bootstrap of opensearch through the `JobSchedulerExtension` extension points [here](https://github.com/opensearch-project/anomaly-detection/blob/fd2fd5d5f63c9004b9f265e3c1e4475282447da8/src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java#L1012). For extensions, this wont work, since extensions are initialized after the bootstrap of opensearch and can come up and down at anytime (hotswap). Additionally, an extension may have the need to register more than one type of job. 

To support this, the job scheduler communication mechanism has introduced a REST API that an extension may call at any time. The only caveat is that these details must be sent to Job Scheduler prior to the initialization of the corresponding job index, which is why we have sent this request prior to the processing of the start detector request.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
joshpalis: There is no direct equivalent for `isFailed` for the class `BulkIResponseItem`. However `.error()` returns the `ErrorCause` if any, or null. Therefore by checking if the `ErrorCause` is null, we can determine if there was an error.

src/main/java/org/opensearch/ad/transport/ADJobRunnerTransportAction.java
joshpalis: we set it below like this :

```
AnomalyDetectorJob scheduledJobParameter = inProgressFuture
                        .orTimeout(AnomalyDetectorSettings.REQUEST_TIMEOUT.get(settings).getMillis(), TimeUnit.MILLISECONDS)
                        .join();
```

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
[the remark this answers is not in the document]
owaiskazi19: We already have a constant for the index `AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX`. Let's use that. No harm in improving the codebase :)

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
owaiskazi19: Got it. Is this the only place we should check whether the job details are registered?

src/main/java/org/opensearch/ad/task/ADBatchTaskRunner.java
joshpalis: Seems that we do not, I went ahead and created one here : https://github.com/opensearch-project/opensearch-sdk-java/issues/635. Ill add these to the comments

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
```diff
@@ -16,60 +16,138 @@
...
+        XContentBuilder requestBody = JsonXContent.contentBuilder();
+        requestBody.startObject();
+        requestBody.field(GetJobDetailsRequest.JOB_INDEX, AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX);
+        requestBody.field(GetJobDetailsRequest.JOB_TYPE, AnomalyDetectorExtension.AD_JOB_TYPE);
+        requestBody.field(GetJobDetailsRequest.JOB_PARAMETER_ACTION, ADJobParameterAction.class.getName());
+        requestBody.field(GetJobDetailsRequest.JOB_RUNNER_ACTION, ADJobRunnerAction.class.getName());
+        requestBody.field(GetJobDetailsRequest.EXTENSION_UNIQUE_ID, extensionsRunner.getUniqueId());
+        requestBody.endObject();
+
+        Request request = new Request("PUT", String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_job_details"));
```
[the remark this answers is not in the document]
joshpalis: seems that the request method parameter only accepts a string. Ill leave this as is

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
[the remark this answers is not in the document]
joshpalis: sure

src/main/java/org/opensearch/ad/rest/RestAnomalyDetectorJobAction.java
joshpalis: Yes, since Job Scheduler is used only for real time analysis, which is triggered by `startdetector`.

src/main/java/org/opensearch/ad/transport/ADJobRunnerTransportAction.java
owaiskazi19: I get that. What I'm trying to ask is shouldn't it be inside `onResponse`?

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -605,33 +617,32 @@ public void initAnomalyResultIndexDirectly(
...
+                        // hash ring. Auto-expand the number of replicas based on the
+                        // number of data nodes (up to 20) in the cluster so that each node can
+                        // become a coordinating node. This is useful when customers
+                        // scale out their cluster so that we can do adaptive scaling
+                        // accordingly.
+                        // At least 1 replica for fail-over.
+                        .put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, minJobIndexReplicas + "-" + maxJobIndexReplicas)
+                        .put("index.hidden", true)
+                );
+            client.indices().create(request, markMappingUpToDate(ADIndex.JOB, actionListener));
```
owaiskazi19, line 640: I guess we do have AdminClient in SDKRestClient. @dbwiddis ?

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -1072,68 +1082,79 @@ public int getSchemaVersion(ADIndex index) {
...
+            // For example, if we have 2 primary shards, since the max number of shards are maxJobIndexReplicas (20),
+            // we will use 20 / 2 = 10 replicas as the upper bound of replica.
+            int maxExpectedReplicas = Math.max(maxJobIndexReplicas / primaryShardsNumber, minJobIndexReplicas);
+            Settings updatedSettings = Settings
+                .builder()
+                .put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, minJobIndexReplicas + "-" + maxExpectedReplicas)
+                .build();
+            final UpdateSettingsRequest updateSettingsRequest = new UpdateSettingsRequest(ADIndex.JOB.getIndexName())
+                .settings(updatedSettings);
+            client.indices().putSettings(updateSettingsRequest, ActionListener.wrap(response -> {
```
owaiskazi19, line 1141: We are making an admin call here. We do have `admin` in client. Let's make sure we request carefully here and for the rest of the migration.
```suggestion
            client.admin().indices().putSettings(updateSettingsRequest, ActionListener.wrap(response -> {
```

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -102,31 +210,250 @@ public Collection<Object> createComponents(ExtensionsRunner runner) {
...
+        jobRunner.setThreadPool(threadPool);
+        jobRunner.setAnomalyResultHandler(anomalyResultHandler);
+        jobRunner.setSettings(environmentSettings);
+        jobRunner.setAnomalyDetectionIndices(anomalyDetectionIndices);
+        jobRunner.setNodeFilter(nodeFilter);
+        jobRunner.setAdTaskManager(adTaskManager);
+
         return ImmutableList
-            .of(sdkRestClient, anomalyDetectionIndices, jvmService, adCircuitBreakerService, adTaskManager, adTaskCacheManager);
+            .of(
```
dbwiddis, line 435: Just a comment as I haven't checked.  Ideally the order here and in code above defining these matches the AD Plugin, to minimize the diff.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -206,12 +534,48 @@ public SDKRestClient restClient() {
...
+                new ScalingExecutorBuilder(
+                    AD_THREAD_POOL_NAME,
+                    1,
+                    // HCAD can be heavy after supporting 1 million entities.
+                    // Limit to use at most half of the processors.
+                    Math.max(1, OpenSearchExecutors.allocatedProcessors(settings) / 2),
+                    TimeValue.timeValueMinutes(10),
+                    AD_THREAD_POOL_PREFIX + AD_THREAD_POOL_NAME
+                )// ,
```
dbwiddis, line 558: (Style) Would really prefer block comment format-preserving comments here.  (Start with `/*-` and keep the indentation in the middle before the `*/`.

src/main/java/org/opensearch/ad/indices/ADIndex.java
```diff
@@ -34,12 +35,11 @@ public enum ADIndex {
...
-    // JOB(
-    // AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX,
-    // false,
-    // ThrowingSupplierWrapper.throwingSupplierWrapper(AnomalyDetectionIndices::getAnomalyDetectorJobMappings)
-    // ),
+    JOB(
+        AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX,
+        false,
+        ThrowingSupplierWrapper.throwingSupplierWrapper(AnomalyDetectionIndices::getAnomalyDetectorJobMappings)
+    ),
```
dbwiddis, line 42: No complaint here, just pointing to the diff above as a reason we shouldn't be using `//` comments on blocks of code :)

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
dbwiddis: Yeah it's just a self-reference to the same client, but lets us avoid changing the diff.  In other classes I've assigned the `adminClient` parameter to `client.admin()` which is really just `client` but keeps the rest of the code the same. We should probably do so here.

src/main/java/org/opensearch/ad/rest/handler/IndexAnomalyDetectorJobActionHandler.java
```diff
@@ -1,362 +1,361 @@
...
-// new OpenSearchStatusException("Failed to execute stop detector action", RestStatus.INTERNAL_SERVER_ERROR),
-// transportService,
-// listener
-// );
-// }
-// };
-// }
-//
-// }
+/**
```
dbwiddis, line 1: UGH, this diff!!!!

I don't even know what changed!

src/main/java/org/opensearch/ad/transport/ADJobRunnerTransportAction.java
joshpalis: No, we only need to either `complete` or `completeExceptionally` within the `ActionListener`. Once `join` is invoked, that's where we can retrieve the response object

dbwiddis (approved):
LGTM with some observations and an agreement about a rename that you can consider now or later when we re-branch and merge.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
joshpalis: Ill just set adminClient to sdkRestClient.admin()

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
joshpalis: yes I've followed it as closely as I could. There are some components that have not been implemented yet, since I havent needed them yet. But overall, the order in which the components are created and returned is the same as ADPlugin

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
joshpalis: Ill go ahead and fix this

src/main/java/org/opensearch/ad/task/ADTaskManager.java
joshpalis: I had to redo some of the UpdateByQueryRequests and DeleteByQueryRequests to use the correct builders. MatchAll doesnt apply anymore

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -206,12 +534,49 @@ public SDKRestClient restClient() {
...
     }
 
-    // @Override
-    public Map<String, Class<? extends TransportAction<? extends ActionRequest, ? extends ActionResponse>>> getActionsMap() {
-        Map<String, Class<? extends TransportAction<? extends ActionRequest, ? extends ActionResponse>>> map = new HashMap<>();
-        map.put(ADJobParameterAction.NAME, ADJobParameterTransportAction.class);
-        map.put(ADJobRunnerAction.NAME, ADJobRunnerTransportAction.class);
-        return map;
+    /**
+     * createComponents doesn't work for Clock as ES process cannot start
```
owaiskazi19, line 538: Also, if you can raise a PR on the `main` to correct this
```suggestion
     * createComponents doesn't work for Clock as OS process cannot start
```

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -649,7 +595,7 @@ private void indexAnomalyResultException(
             log.error("Failed to index anomaly result for " + detectorId, e);
         } finally {
             if (releaseLock) {
-                releaseLock(jobParameter, lockService, lock);
+                releaseLock(jobParameter, lock);
```
owaiskazi19, line 598: Trying to understand. How are we handling the `lockService`?

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -685,15 +631,51 @@ private void updateLatestRealtimeTask(
...
+    private void releaseLock(AnomalyDetectorJob jobParameter, LockModel lock) {
+
+        Request releaseLockRequest = new Request(
+            "PUT",
+            String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_release_lock")
+        );
+        releaseLockRequest.addParameter(LockModel.LOCK_ID, lock.getLockId());
+
+        try {
+            Response releaseLockResponse = client.performRequest(releaseLockRequest);
```
owaiskazi19, line 667: Is `performRequest` async?

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
```diff
@@ -436,44 +441,48 @@ public void deleteModelCheckpointByDetectorId(String detectorID) {
...
-            }
         } else {
-            logger.warn(SEARCH_FAILURE_LOG_MSG + " {}", detectorID);
-            for (ScrollableHitSource.SearchFailure searchFailure : response.getSearchFailures()) {
-                logger.warn(searchFailure);
+            if (!response.failures().isEmpty()) {
+                logger.warn(BULK_FAILURE_LOG_MSG + " {}", detectorID);
+                for (BulkIndexByScrollFailure bulkFailure : response.failures()) {
+                    logger.warn(bulkFailure);
+                }
```
owaiskazi19, line 485: This is a little confusing to me. We have `BulkIndexByScrollFailure` but old code has `ScrollableHitSource` which is use for the scrollable results. @saratvemulapalli need some inputs here to verify the changes in this file

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1677,31 +1720,38 @@ protected <T> void deleteTaskDocs(
...
                     logger.info("Old AD tasks deleted for detector {}", detectorId);
-                    BulkItemResponse[] bulkItemResponses = res.getItems();
-                    if (bulkItemResponses != null && bulkItemResponses.length > 0) {
-                        for (BulkItemResponse bulkItemResponse : bulkItemResponses) {
-                            if (!bulkItemResponse.isFailed()) {
-                                logger.debug("Add detector task into cache. Task id: {}", bulkItemResponse.getId());
+                    List<BulkResponseItem> bulkItemResponses = res.items();
+                    if (bulkItemResponses != null && bulkItemResponses.size() > 0) {
+                        for (BulkResponseItem bulkItemResponse : bulkItemResponses) {
+                            if (bulkItemResponse.error() == null) {
```
owaiskazi19, line 1739: Nit: Verify once. Isn't the condition the other way round?

owaiskazi19 (commented):
Almost there. Few more nitpicks and questions

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
joshpalis: The `lockService` isn't directly available for extensions since extensions won't be installed within OpenSearch. To support acquiring/releasing a lock for an extension, we have exposed two REST APIs on job scheduler to acquire a lock (returns an object of type LockModel) and to release a lock (which updates the metadata lock index entry on job scheduler to allow a different job to run).

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
joshpalis: `performRequest ` is a blocking call. There is currently no async mechanism to send a rest call via the SDKRestClient.

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
minalsha: @joshpalis  extensions maynot be installed within OpenSearch right now since we are working on running extensions out of process. But in near future, extensions can run in-proc within OpenSearch. We may want to think about this for near future.

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -306,9 +318,9 @@ public boolean doesAnomalyDetectorIndexExist() {
      *
      * @return true if anomaly detector job index exists
      */
-    // public boolean doesAnomalyDetectorJobIndexExist() {
-    // return clusterService.state().getRoutingTable().hasIndex(AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX);
-    // }
+    public boolean doesAnomalyDetectorJobIndexExist() {
+        return clusterService.state().getRoutingTable().hasIndex(AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX);
```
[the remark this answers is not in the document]
joshpalis: seems this is coming from spotless, I am unable to prevent this

src/main/java/org/opensearch/ad/task/ADTaskManager.java
```diff
@@ -1069,51 +1084,51 @@ private <T> void resetLatestDetectorTaskState(
...
+                    listener.onFailure(e);
+                }
+            } else {
+                logger.debug("AD job is not found, reset realtime task as stopped for detector {}", detectorId);
+                resetTaskStateAsStopped(adTask, function, transportService, listener);
+            }
+        }, e -> {
+            logger.error("Fail to get AD realtime job for detector " + detectorId, e);
+            listener.onFailure(e);
+        }));
```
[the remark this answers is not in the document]
joshpalis: this is coming from spotless

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
joshpalis: @VachaShah just wanted to get your input on this. Please correct me if I am wrong, but as we discussed before, the `DeleteByQueryResponse` failures for opensearch-java's `OpenSearchAsyncClient` must be read from the list of `BulkIndexByScrollFailure` and there are no corresponding types for the HighLevelRestClient's `SearchFailure` and `BulkFailure`.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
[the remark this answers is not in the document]
owaiskazi19: Ah, spotless!

src/main/java/org/opensearch/ad/transport/ADJobRunnerTransportAction.java
owaiskazi19: This is an edit in a new file and not the old plugin code. Fine with this change!

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
joshpalis: @minalsha Even if extensions run in-process, extensions are still initialized after the bootstrap of OpenSearch, which is when Job Scheduler class loads the plugin implementations of `ScheduledJobRunners`. This is when the `lockService` is provided for plugins. For extensions, we still woulnd't be able to access the lock service directly, that communication will have to be facilitated through the REST APIs that we exposed on Job Scheduler to acquire and release the lock

minalsha: sounds good. Thanks for explaining it.

src/main/java/org/opensearch/ad/task/ADTaskManager.java
joshpalis: Correct, let me change this

src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java
owaiskazi19: @joshpalis let's add async in SDK and use it here. This can be done in the next PR.

owaiskazi19 (approved):
Looking for some inputs from @VachaShah! If any code change is required @joshpalis you can handle it in the next PR!

src/main/java/org/opensearch/ad/ml/CheckpointDao.java
VachaShah: Yes for `DeleteByQueryResponse` in the java-client, the failures need to be read from the list of `BulkByIndexScrollFailures`.

PR #838, opened 2023-03-14
[Feature/Extensions] Use parameterized SDKClusterSettings in tests

dbwiddis:
### Description

Updates tests which needed SDKClusterSettings instantiated with a list of settings following merge of [SDK #559](https://github.com/opensearch-project/opensearch-sdk-java/pull/559).

### Issues Resolved
Fixes [SDK #503](https://github.com/opensearch-project/opensearch-sdk-java/issues/503)

### DIFF
--- a/src/test/java/org/opensearch/ad/indices/CustomIndexTests.java
+++ b/src/test/java/org/opensearch/ad/indices/CustomIndexTests.java
@@ -15,10 +15,9 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 
 import org.opensearch.Version;
@@ -33,14 +32,17 @@ import org.opensearch.cluster.ClusterState;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.metadata.MappingMetadata;
 import org.opensearch.cluster.metadata.Metadata;
-import org.opensearch.cluster.service.ClusterService;
-import org.opensearch.common.settings.ClusterSettings;
+import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.sdk.Extension;
+import org.opensearch.sdk.ExtensionsRunner;
+import org.opensearch.sdk.SDKClusterService;
+import org.opensearch.sdk.SDKClusterService.SDKClusterSettings;
 
 public class CustomIndexTests extends AbstractADTest {
     AnomalyDetectionIndices adIndices;
     Client client;
-    ClusterService clusterService;
+    SDKClusterService clusterService;
     DiscoveryNodeFilterer nodeFilter;
     ClusterState clusterState;
     String customIndexName;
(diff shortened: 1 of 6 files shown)


src/test/java/org/opensearch/ad/indices/CustomIndexTests.java
```diff
@@ -52,31 +54,28 @@ public void setUp() throws Exception {
...
+                AnomalyDetectorSettings.AD_RESULT_HISTORY_MAX_DOCS_PER_SHARD,
+                AnomalyDetectorSettings.AD_RESULT_HISTORY_ROLLOVER_PERIOD,
+                AnomalyDetectorSettings.AD_RESULT_HISTORY_RETENTION_PERIOD,
+                AnomalyDetectorSettings.MAX_PRIMARY_SHARDS
+            );
+        ExtensionsRunner mockRunner = mock(ExtensionsRunner.class);
+        Extension mockExtension = mock(Extension.class);
+        when(mockRunner.getEnvironmentSettings()).thenReturn(settings);
+        when(mockRunner.getExtension()).thenReturn(mockExtension);
+        when(mockExtension.getSettings()).thenReturn(settingsList);
```
owaiskazi19, line 77: Thinking it out loud. It's definitely good to have more tests and checks but since we are focusing on migration. Should we add other checks except the one below?

dbwiddis: Right now I'm trying to restore exactly the same test functionality that already exists.  Not sure what needs to be added that you wouldn't want to also add to the main project.

owaiskazi19: Let me be more specific here. I meant to ask do we need to test these because to reduce the diff in terms of migration these tests are extra addition :

```
 List<Setting<?>> settingsList = List.of(AnomalyDetectorSettings.MAX_ENTITIES_FOR_PREVIEW, AnomalyDetectorSettings.PAGE_SIZE);
        ExtensionsRunner mockRunner = mock(ExtensionsRunner.class);
        Extension mockExtension = mock(Extension.class);
        when(mockRunner.getEnvironmentSettings()).thenReturn(settings);
        when(mockRunner.getExtension()).thenReturn(mockExtension);
        when(mockExtension.getSettings()).thenReturn(settingsList);

dbwiddis: I think I understand.  Here's the dilemma:
 - I implemented the `SDKClusterSettings` as an inner class, so it would have access to the `SDKClusterService` instance field of ExtensionsRunner.
 - The reason for this is that the Cluster Settings object needs methods involving transport (settings update consumers, etc.) that require the ExtensionsRunner or other objects from it.
 - This means we can't just instantiate it like a regular class: that would require it to be a static (nested) class.  But instantiating it like above would not allow it to have access to the Extensions Runner.
 - There are workarounds to all of the above, but generally speaking I think minimizing the diff in the src/main/java filepath is worth it for a few boilerplate copy/paste mocking workarounds in a small number of test classes (we could even create a separate single mock somewhere to minimize repetition if it gets out of hand).

I welcome you to propose a better way to do this.  Believe me I tried many different options and settled on this one, but TMTOWTDI.
 - So matching the existing constructor taking arguments that do not include
 - to match the existing test (and reduce the diff

owaiskazi19: I liked the idea of creating a separate single mock if migration changes are creating more diff. We can think about this later.

owaiskazi19 (approved):

PR #843, opened 2023-03-20
[Feature/extensions] Update initialization of SDK Clients

dbwiddis:
### Description

Updates the `AnomalyDetectorExtension` class:
 - BaseExtension no longer implements `ActionExtension` so that was added, otherwise the `@Override` annotation on `getExtensionRestHandlers()` was broken.
 - Fixes initialization of RestClient and AsyncClient to always use the same (initialized) SDKClient instance
 - Removes the ExtensionSettings parameter, which will be removed in [SDK #579](https://github.com/opensearch-project/opensearch-sdk-java/pull/579)

### Issues Resolved

Part of [SDK #556](https://github.com/opensearch-project/opensearch-sdk-java/pull/579) but fixes things broken from previous merges.

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -190,7 +190,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
 
     // TODO: replace or override client object on BaseExtension
     // https://github.com/opensearch-project/opensearch-sdk-java/issues/160
-    public OpenSearchAsyncClient getClient() {
+    public OpenSearchAsyncClient javaAsyncClient() {
         return this.sdkJavaAsyncClient;
     }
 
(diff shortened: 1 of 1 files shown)


dbwiddis:
Gradle check failing until [SDK #579](https://github.com/opensearch-project/opensearch-sdk-java/pull/579/files) is merged.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -179,22 +182,22 @@ public List<NamedXContentRegistry.Entry> getNamedXContent() {
...
-                getExtensionSettings().getOpensearchAddress(),
-                Integer.parseInt(getExtensionSettings().getOpensearchPort())
-            );
+        OpenSearchAsyncClient client = runner.getSdkClient().initializeJavaAsyncClient();
         return client;
     }
 
+    // TODO: replace or override client object on BaseExtension
+    // https://github.com/opensearch-project/opensearch-sdk-java/issues/160
+    public OpenSearchAsyncClient getClient() {
```
owaiskazi19, line 193: Since we have the getter for `sdkRestClient` as `restClient`. We can keep this method name as `javaClient`

dbwiddis: how about `javaAsyncClient()`?

PR #850, opened 2023-03-30
Giving admin priority over backendrole filtering

### LINKED ISSUE #842: [BUG] Prioritize admin access over backend roles
**What is the bug?**
Currently when security is enabled and backend role filtering is also enabled we prioritize a matching backend role over administrator access. So if the admin doesn't have a matching backend role to the detector the detector isn't accessible to them which goes against the standard that admin should have access to everything.

On a security enabled cluster, enable backend role filtering for AD. Create a detector with user1 that has backend role A. Make sure admin user doesn't have backend role B, and the detector wont be shown on the detector list.

**What is the expected behavior?**
All detectors regardless of backend role should be shown on the detector list for the admin user.

**Solution**
We should check if the user is an admin first and if they are we can ignore backend role matching. This should also be added to documentation.

amitgalitz:
### Description
Prioritizing admin role over backend role filtering for all AD operations.

Validated manually that admins can now perform all AD APIs on given detector even if the backend roles don't match. Also if an admin updates a detector, it won't rewrite the backend role.

Security tests pass against a 2.6 cluster, 3.0 distribution has other issues that are being solved by other plugins.

### Issues Resolved

resolves #842 

### DIFF
--- a/src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
+++ b/src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
@@ -247,7 +247,6 @@ public class SecureADRestIT extends AnomalyDetectorRestTestCase {
             ),
             null
         );
-        enableFilterBy();
         // User client has admin all access, and has "opensearch" backend role so client should be able to update detector
         // But the detector's backend role should not be replaced as client's backend roles (all_access).
         Response response = updateAnomalyDetector(aliceDetector.getDetectorId(), newDetector, client());
--- a/src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
+++ b/src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
@@ -54,8 +54,11 @@ public class SecureADRestIT extends AnomalyDetectorRestTestCase {
     RestClient goatClient;
     String lionUser = "lion";
     RestClient lionClient;
+    String tigerUser = "tiger";
+    RestClient tigerClient;
     private String indexAllAccessRole = "index_all_access";
     private String indexSearchAccessRole = "index_all_search";
+    private String adminAllAccessRole = "all_access";
 
     @Before
     public void setupSecureTests() throws IOException {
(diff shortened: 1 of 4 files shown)


kaituo:
For "Also if an admin updates a detector, it won't rewrite the backend role. ", is it because we have forbidden rewriting user when updating a detector here: https://github.com/opensearch-project/anomaly-detection/pull/126 ?

amitgalitz:
> Overall looks good to me. Waiting for your investigation on security tests.
> 
> Also, for
> 
> "For 2.6 with all plugins there and a version that is already released I am getting this error when trying to create a user:
> 
> {"status":"FORBIDDEN","message":"No permission to access REST API: User admin with Security roles [own_index] does not have any role privileged for admin access. No client TLS certificate found in request"}"
> 
> Did you mean you cannot create a user in 2.6+ cluster?

During the setup, establishing the "tigerClient" as admin led to all other clients failing and not being able to create other users. I realized I can just use the client() as admin and I also added a method to confirm client() indeed has the "all_access" role to help reading code in the future.

amitgalitz:
> For "Also if an admin updates a detector, it won't rewrite the backend role. ", is it because we have forbidden rewriting user when updating a detector here: #126 ?

That is my assumption, I saw this specifically https://github.com/opensearch-project/anomaly-detection/pull/126/files#diff-16fb3468e4277c88b68a142f897f888e2505820c75a22f9166cf783a0b807111R164. I didn't specifically confirm this is where it happens outside of the manual and integration testings

amitgalitz:
@kaituo security tests pass on 2.6 cluster. 3.0 cluster giving errors cause of distributions failures from other plugins.

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -186,6 +197,63 @@ public void testGetApiFilterByEnabled() throws IOException {
...
+            aliceDetector.getCategoryField(),
+            new User(
+                randomAlphaOfLength(5),
+                ImmutableList.of("odfe", randomAlphaOfLength(5)),
+                ImmutableList.of(randomAlphaOfLength(5)),
+                ImmutableList.of(randomAlphaOfLength(5))
+            ),
+            null
+        );
+        enableFilterBy();
```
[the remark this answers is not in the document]
amitgalitz: Yes, I removed this

kaituo (approved):

src/test/java/org/opensearch/ad/util/ParseUtilsTests.java
```diff
@@ -287,4 +288,29 @@ public void testGetFeatureFieldNames() throws IOException {
...
+
+    public void testIsAdminBackendRoleIsAllAccess() {
+        String backendRole1 = "all_access";
+        User user1 = new User(
+            randomAlphaOfLength(5),
+            ImmutableList.of(backendRole1),
+            ImmutableList.of(randomAlphaOfLength(5)),
+            ImmutableList.of(randomAlphaOfLength(5))
+        );
+        assertFalse(isAdmin(user1));
```
jackiehanyang, line 310: curiosity question: a bit confused on this test case. If we're just checking if the role contains `all_access` keyword to decide if the role is admin, why would this test case fail? Cause in this test case, the backendRole1 seems also contains `all_access`?

> `user.getRoles().contains("all_access")`

jackiehanyang (approved):

src/test/java/org/opensearch/ad/util/ParseUtilsTests.java
amitgalitz: This test case is failing because all_access is added as a backend role and not as a role. This is to make sure user can't just be given backend role name of `all_access` and bypass our `isAdmin()` check which is only if they actually have the role itself of `all_access`. Its a little confusing terminology but backend_role are more  Arbitrary strings that you can give to a user to either easily map roles or to identify the group they are in so for example we can use it to filter which detectors they can see or not.

PR #855, opened 2023-04-05
[Feature/extensions] Integrated custom result index, injected guice for Create, Validate and Get Detector

owaiskazi19:
### Description
This PR:
- [ ] Creates custom result index with Create Detector
- [ ] Used `client.execute` for Create, Validate and Get Detector
- [ ] Added new test file `GetDetectorTransportActionTests`

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/651


src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
```diff
@@ -58,17 +57,13 @@ public class RestGetAnomalyDetectorAction extends BaseExtensionRestHandler {
     private static final String GET_ANOMALY_DETECTOR_ACTION = "get_anomaly_detector";
     private static final Logger logger = LogManager.getLogger(RestGetAnomalyDetectorAction.class);
     private Settings settings;
-    private TransportService transportService;
     private SDKRestClient client;
-    private SDKClusterService clusterService;
     private ExtensionsRunner extensionsRunner;
 
     public RestGetAnomalyDetectorAction(ExtensionsRunner extensionsRunner, SDKRestClient client) {
         this.extensionsRunner = extensionsRunner;
```
joshpalis, line 64: Since we're not using the extensionsRunner anywhere else besides the constructor to retrieve the settings, no need to keep it as a class field.

src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
```diff
@@ -105,15 +100,13 @@ protected ExtensionRestResponse prepareRequest(RestRequest request) throws IOExc
         String detectorId = request.param(DETECTOR_ID);
         String typesStr = request.param(TYPE);
 
-        String rawPath = request.path();
-        // FIXME handle this
-        // Passed false until job scheduler is integrated
-        boolean returnJob = false;
-        boolean returnTask = false;
-        boolean all = false;
+        String rawPath = request.rawPath();
```
joshpalis, line 103: seems this was originally the `request.path()` instead of the `request.rawPath()`. Any particular reason for this change?

owaiskazi19: Because previously we were using `ExtensionRestRequest`. 
`RestRequest` has [request.rawPath()](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java#L64)

src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java
owaiskazi19: @joshpalis doesn't have the write access. Resolving the comments on his behalf.

joshpalis (approved):

dbwiddis (approved):

PR #857, opened 2023-04-06
[Feature/extensions] Get detector stats api

vibrantvarun:
### Description
Get Detector Stats API
The api's Provides information about how the plugin is performing.

### Issues Resolved
[https://github.com/opensearch-project/opensearch-sdk-java/issues/382](https://github.com/opensearch-project/opensearch-sdk-java/issues/382)

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -165,7 +165,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
                 new RestGetAnomalyDetectorAction(extensionsRunner(), restClient()),
                 new RestAnomalyDetectorJobAction(extensionsRunner(), restClient()),
                 new RestDeleteAnomalyDetectorAction(extensionsRunner(), restClient()),
-                new RestStatsAnomalyDetectorAction(extensionsRunner(), restClient(),adStats,nodeFilter)
+                new RestStatsAnomalyDetectorAction(extensionsRunner(), restClient(), adStats, nodeFilter)
             );
     }
 
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -140,6 +140,8 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
     private static Gson gson;
     // package private for testing
     GenericObjectPool<LinkedBuffer> serializeRCFBufferPool;
+    private ADStats adStats;
+    private DiscoveryNodeFilterer nodeFilter;
 
     static {
         SpecialPermission.check();
@@ -161,7 +163,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
                 new RestGetAnomalyDetectorAction(extensionsRunner(), restClient()),
                 new RestAnomalyDetectorJobAction(extensionsRunner(), restClient()),
                 new RestDeleteAnomalyDetectorAction(extensionsRunner(), restClient()),
-                new RestStatsAnomalyDetectorAction(extensionsRunner(), restClient())
+                new RestStatsAnomalyDetectorAction(extensionsRunner(), restClient(),adStats,nodeFilter)
             );
     }
 
@@ -183,7 +185,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
         Throttler throttler = new Throttler(getClock());
         ClientUtil clientUtil = new ClientUtil(environmentSettings, restClient(), throttler);
         IndexUtils indexUtils = new IndexUtils(restClient(), clientUtil, sdkClusterService, indexNameExpressionResolver, javaAsyncClient());
-        DiscoveryNodeFilterer nodeFilter = new DiscoveryNodeFilterer(sdkClusterService);
+        nodeFilter = new DiscoveryNodeFilterer(sdkClusterService);
         AnomalyDetectionIndices anomalyDetectionIndices = new AnomalyDetectionIndices(
             sdkRestClient,
             sdkJavaAsyncClient,
@@ -409,7 +411,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
             .put(StatNames.MODEL_COUNT.getName(), new ADStat<>(false, new ModelsOnNodeCountSupplier(modelManager, cacheProvider)))
             .build();
 
-        ADStats adStats = new ADStats(stats);
+        adStats = new ADStats(stats);
 
         ADTaskCacheManager adTaskCacheManager = new ADTaskCacheManager(environmentSettings, sdkClusterService, memoryTracker);
         ADTaskManager adTaskManager = new ADTaskManager(
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -51,11 +51,11 @@ import org.opensearch.ad.model.AnomalyDetectorJob;
 import org.opensearch.ad.model.AnomalyResult;
 import org.opensearch.ad.model.DetectorInternalState;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
-import org.opensearch.ad.rest.RestIndexAnomalyDetectorAction;
-import org.opensearch.ad.rest.RestValidateAnomalyDetectorAction;
-import org.opensearch.ad.rest.RestGetAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestAnomalyDetectorJobAction;
+import org.opensearch.ad.rest.RestGetAnomalyDetectorAction;
+import org.opensearch.ad.rest.RestIndexAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestStatsAnomalyDetectorAction;
+import org.opensearch.ad.rest.RestValidateAnomalyDetectorAction;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.ad.stats.ADStat;
@@ -68,14 +68,7 @@ import org.opensearch.ad.stats.suppliers.ModelsOnNodeSupplier;
 import org.opensearch.ad.stats.suppliers.SettableSupplier;
 import org.opensearch.ad.task.ADTaskCacheManager;
 import org.opensearch.ad.task.ADTaskManager;
-import org.opensearch.ad.transport.ADJobParameterAction;
-import org.opensearch.ad.transport.ADJobRunnerAction;
-import org.opensearch.ad.transport.AnomalyDetectorJobAction;
-import org.opensearch.ad.transport.StatsAnomalyDetectorAction;
-import org.opensearch.ad.transport.ADJobParameterTransportAction;
-import org.opensearch.ad.transport.ADJobRunnerTransportAction;
-import org.opensearch.ad.transport.AnomalyDetectorJobTransportAction;
-import org.opensearch.ad.transport.StatsAnomalyDetectorTransportAction;
+import org.opensearch.ad.transport.*;
 import org.opensearch.ad.transport.handler.AnomalyIndexHandler;
 import org.opensearch.ad.util.ClientUtil;
 import org.opensearch.ad.util.DiscoveryNodeFilterer;
@@ -153,7 +146,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
                 new RestValidateAnomalyDetectorAction(extensionsRunner(), restClient(), javaAsyncClient()),
                 new RestGetAnomalyDetectorAction(extensionsRunner(), restClient()),
                 new RestAnomalyDetectorJobAction(extensionsRunner(), restClient()),
-                    new RestStatsAnomalyDetectorAction(extensionsRunner(),restClient(),adStats,nodeFilter)
+                new RestStatsAnomalyDetectorAction(extensionsRunner(), restClient(), adStats, nodeFilter)
 
             );
     }
@@ -177,7 +170,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
 
         Throttler throttler = new Throttler(getClock());
         ClientUtil clientUtil = new ClientUtil(environmentSettings, restClient(), throttler);
-        IndexUtils indexUtils = new IndexUtils(restClient(), clientUtil, sdkClusterService, indexNameExpressionResolver);
+        IndexUtils indexUtils = new IndexUtils(restClient(), clientUtil, sdkClusterService, indexNameExpressionResolver, javaAsyncClient());
         nodeFilter = new DiscoveryNodeFilterer(sdkClusterService);
         AnomalyDetectionIndices anomalyDetectionIndices = new AnomalyDetectionIndices(
             sdkRestClient,
@@ -583,7 +576,8 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
                 new ActionHandler<>(ADJobRunnerAction.INSTANCE, ADJobRunnerTransportAction.class),
                 new ActionHandler<>(ADJobParameterAction.INSTANCE, ADJobParameterTransportAction.class),
                 new ActionHandler<>(AnomalyDetectorJobAction.INSTANCE, AnomalyDetectorJobTransportAction.class),
-                new ActionHandler<>(StatsAnomalyDetectorAction.INSTANCE, StatsAnomalyDetectorTransportAction.class)
+                new ActionHandler<>(StatsAnomalyDetectorAction.INSTANCE, StatsAnomalyDetectorTransportAction.class),
+                new ActionHandler<>(ADStatsNodesAction.INSTANCE, ADStatsNodesTransportAction.class)
                 // TODO : Register AnomalyResultAction/TransportAction here :
                 // https://github.com/opensearch-project/opensearch-sdk-java/issues/626
             );
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -51,10 +51,11 @@ import org.opensearch.ad.model.AnomalyDetectorJob;
 import org.opensearch.ad.model.AnomalyResult;
 import org.opensearch.ad.model.DetectorInternalState;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
-import org.opensearch.ad.rest.RestAnomalyDetectorJobAction;
-import org.opensearch.ad.rest.RestGetAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestIndexAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestValidateAnomalyDetectorAction;
+import org.opensearch.ad.rest.RestGetAnomalyDetectorAction;
+import org.opensearch.ad.rest.RestAnomalyDetectorJobAction;
+import org.opensearch.ad.rest.RestStatsAnomalyDetectorAction;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.ad.stats.ADStat;
@@ -68,11 +69,13 @@ import org.opensearch.ad.stats.suppliers.SettableSupplier;
 import org.opensearch.ad.task.ADTaskCacheManager;
 import org.opensearch.ad.task.ADTaskManager;
 import org.opensearch.ad.transport.ADJobParameterAction;
-import org.opensearch.ad.transport.ADJobParameterTransportAction;
 import org.opensearch.ad.transport.ADJobRunnerAction;
-import org.opensearch.ad.transport.ADJobRunnerTransportAction;
 import org.opensearch.ad.transport.AnomalyDetectorJobAction;
+import org.opensearch.ad.transport.StatsAnomalyDetectorAction;
+import org.opensearch.ad.transport.ADJobParameterTransportAction;
+import org.opensearch.ad.transport.ADJobRunnerTransportAction;
 import org.opensearch.ad.transport.AnomalyDetectorJobTransportAction;
+import org.opensearch.ad.transport.StatsAnomalyDetectorTransportAction;
 import org.opensearch.ad.transport.handler.AnomalyIndexHandler;
 import org.opensearch.ad.util.ClientUtil;
 import org.opensearch.ad.util.DiscoveryNodeFilterer;
@@ -127,6 +130,10 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
     // package private for testing
     GenericObjectPool<LinkedBuffer> serializeRCFBufferPool;
 
+    private ADStats adStats;
+
+    private DiscoveryNodeFilterer nodeFilter;
+
     static {
         SpecialPermission.check();
         // gson intialization requires "java.lang.RuntimePermission" "accessDeclaredMembers" to
@@ -145,7 +152,9 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
                 new RestIndexAnomalyDetectorAction(extensionsRunner(), restClient(), javaAsyncClient()),
                 new RestValidateAnomalyDetectorAction(extensionsRunner(), restClient(), javaAsyncClient()),
                 new RestGetAnomalyDetectorAction(extensionsRunner(), restClient()),
-                new RestAnomalyDetectorJobAction(extensionsRunner(), restClient())
+                new RestAnomalyDetectorJobAction(extensionsRunner(), restClient()),
+                    new RestStatsAnomalyDetectorAction(extensionsRunner(),restClient(),adStats,nodeFilter)
+
             );
     }
 
@@ -169,7 +178,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
         Throttler throttler = new Throttler(getClock());
         ClientUtil clientUtil = new ClientUtil(environmentSettings, restClient(), throttler);
         IndexUtils indexUtils = new IndexUtils(restClient(), clientUtil, sdkClusterService, indexNameExpressionResolver);
-        DiscoveryNodeFilterer nodeFilter = new DiscoveryNodeFilterer(sdkClusterService);
+        nodeFilter = new DiscoveryNodeFilterer(sdkClusterService);
         AnomalyDetectionIndices anomalyDetectionIndices = new AnomalyDetectionIndices(
             sdkRestClient,
             sdkJavaAsyncClient,
@@ -395,7 +404,7 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
             .put(StatNames.MODEL_COUNT.getName(), new ADStat<>(false, new ModelsOnNodeCountSupplier(modelManager, cacheProvider)))
             .build();
 
-        ADStats adStats = new ADStats(stats);
+        adStats = new ADStats(stats);
 
         ADTaskCacheManager adTaskCacheManager = new ADTaskCacheManager(environmentSettings, sdkClusterService, memoryTracker);
         ADTaskManager adTaskManager = new ADTaskManager(
@@ -573,7 +582,8 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
             .asList(
                 new ActionHandler<>(ADJobRunnerAction.INSTANCE, ADJobRunnerTransportAction.class),
                 new ActionHandler<>(ADJobParameterAction.INSTANCE, ADJobParameterTransportAction.class),
-                new ActionHandler<>(AnomalyDetectorJobAction.INSTANCE, AnomalyDetectorJobTransportAction.class)
+                new ActionHandler<>(AnomalyDetectorJobAction.INSTANCE, AnomalyDetectorJobTransportAction.class),
+                new ActionHandler<>(StatsAnomalyDetectorAction.INSTANCE, StatsAnomalyDetectorTransportAction.class)
                 // TODO : Register AnomalyResultAction/TransportAction here :
                 // https://github.com/opensearch-project/opensearch-sdk-java/issues/626
             );
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -102,12 +102,12 @@ import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.monitor.jvm.JvmInfo;
 import org.opensearch.monitor.jvm.JvmService;
 import org.opensearch.sdk.BaseExtension;
-import org.opensearch.sdk.ExtensionRestHandler;
 import org.opensearch.sdk.ExtensionsRunner;
 import org.opensearch.sdk.SDKClient.SDKRestClient;
 import org.opensearch.sdk.SDKClusterService;
 import org.opensearch.sdk.SDKNamedXContentRegistry;
 import org.opensearch.sdk.api.ActionExtension;
+import org.opensearch.sdk.rest.ExtensionRestHandler;
 import org.opensearch.threadpool.ExecutorBuilder;
 import org.opensearch.threadpool.ScalingExecutorBuilder;
 import org.opensearch.threadpool.ThreadPool;
@@ -129,10 +129,12 @@ public class AnomalyDetectorExtension extends BaseExtension implements ActionExt
 
     private static final String EXTENSION_SETTINGS_PATH = "/ad-extension.yml";
 
+    public static final String LEGACY_AD_BASE = "/_opendistro/_anomaly_detection";
+    public static final String LEGACY_OPENDISTRO_AD_BASE_URI = LEGACY_AD_BASE + "/detectors";
     public static final String AD_BASE_DETECTORS_URI = "/detectors";
-    public static final String AD_JOB_TYPE = "opendistro_anomaly_detector";
     public static final String AD_THREAD_POOL_PREFIX = "opensearch.ad.";
     public static final String AD_THREAD_POOL_NAME = "ad-threadpool";
+    public static final String AD_JOB_TYPE = "opendistro_anomaly_detector";
 
     @Deprecated
     private SDKRestClient sdkRestClient;
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
@@ -52,6 +52,7 @@ import org.opensearch.ad.model.AnomalyResult;
 import org.opensearch.ad.model.DetectorInternalState;
 import org.opensearch.ad.ratelimit.CheckpointWriteWorker;
 import org.opensearch.ad.rest.RestAnomalyDetectorJobAction;
+import org.opensearch.ad.rest.RestDeleteAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestGetAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestIndexAnomalyDetectorAction;
 import org.opensearch.ad.rest.RestStatsAnomalyDetectorAction;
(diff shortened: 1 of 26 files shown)


src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -11,64 +11,150 @@
...
+import org.opensearch.sdk.SDKClusterService;
 
 import com.google.common.collect.ImmutableList;
 
 /**
- * RestStatsAnomalyDetectorAction consists of the REST handler to get the stats from the anomaly detector plugin.
+ * RestStatsAnomalyDetectorAction consists of the REST handler to get the stats from the anomaly detector extension.
  */
-public class RestStatsAnomalyDetectorAction extends BaseRestHandler {
+public class RestStatsAnomalyDetectorAction extends AbstractAnomalyDetectorAction {
```
joshpalis, line 60: why extend the `AbstractAnomalyDetectorAction` instead of `BaseExtensionRestHandler`

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -145,7 +145,9 @@ public List<ExtensionRestHandler> getExtensionRestHandlers() {
                 new RestIndexAnomalyDetectorAction(extensionsRunner(), restClient(), javaAsyncClient()),
                 new RestValidateAnomalyDetectorAction(extensionsRunner(), restClient(), javaAsyncClient()),
                 new RestGetAnomalyDetectorAction(extensionsRunner(), restClient()),
-                new RestAnomalyDetectorJobAction(extensionsRunner(), restClient())
+                new RestAnomalyDetectorJobAction(extensionsRunner(), restClient()),
+                new RestStatsAnomalyDetectorAction(extensionsRunner(), restClient(), adStats, nodeFilter)
```
dbwiddis, line 149: ADStats and DiscoveryNodeFilterer objects are already returned via `createComponents()` so they are available by injection in the action.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -395,7 +397,7 @@ public PooledObject<LinkedBuffer> wrap(LinkedBuffer obj) {
             .put(StatNames.MODEL_COUNT.getName(), new ADStat<>(false, new ModelsOnNodeCountSupplier(modelManager, cacheProvider)))
             .build();
 
-        ADStats adStats = new ADStats(stats);
+        adStats = new ADStats(stats);
```
dbwiddis, line 400: Per previous comment, you can leave this as a local variable that you inject, no need to make it an instance variable.

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
dbwiddis: Looks like it gets that base class by inheritance, and adds access to several settings in the process that will get updated via settings update consumers.

However, later on in this class we're just fetching the setting directly and not taking advantage of the inheritance.

So either way is good but we should be consistent!

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -11,92 +11,75 @@
...
+
+import com.google.inject.Inject;
 
 /**
  *  ADStatsNodesTransportAction contains the logic to extract the stats from the nodes
  */
-public class ADStatsNodesTransportAction extends
-    TransportNodesAction<ADStatsRequest, ADStatsNodesResponse, ADStatsNodeRequest, ADStatsNodeResponse> {
-
+public class ADStatsNodesTransportAction extends TransportAction<ADStatsNodeRequest, ADStatsNodesResponse> {
```
dbwiddis, line 38: What are we losing by removing this inheritance vs. replacing it with an SDK equivalent?   This approach will require all the classes which inherit from this same superclass to have to make the same changes, while if you update (or replace) the superclass, then that's a one-time change that all the subclasses can use.

src/main/java/org/opensearch/ad/util/IndexUtils.java
```diff
@@ -116,17 +126,27 @@ public String getIndexHealthStatus(String indexOrAliasName) throws IllegalArgume
...
-        return response.map(r -> r.getIndex(indexName).getPrimaries().docs.getCount()).orElse(-1L);
+        IndicesStatsRequest indicesStatsRequest = new IndicesStatsRequest.Builder().build();
+
+        CompletableFuture<IndicesStatsResponse> indicesStatsFutureResponse = openSearchAsyncClient
+            ._transport()
+            .performRequestAsync(indicesStatsRequest, IndicesStatsRequest._ENDPOINT, TransportOptions.builder().build());
+        // Optional<IndicesStatsResponse> response = clientUtil.timedRequest(indicesStatsRequest, logger, client.admin().indices()::stats);
+
+        IndicesStatsResponse indicesStatsResponse;
+        Long numberOfDocumentsInIndex = -1L;
```
dbwiddis, line 141: Why used a boxed `long` here?  I see the method return type is a `Long` but that's normally used if you want to return `null` (and may have originally had that logic before JDK8 and `Optional`).  And perhaps that's why this method is now deprecated.  Right now we're returning either 0 or -1 in error conditions, so we can save the boxing until the final return.

src/main/java/org/opensearch/ad/util/IndexUtils.java
```diff
@@ -116,17 +126,27 @@ public String getIndexHealthStatus(String indexOrAliasName) throws IllegalArgume
      * @return The number of documents in an index. 0 is returned if the index does not exist. -1 is returned if the
      * request fails.
      */
-    /*  @anomaly-detection - commented until we have support for SDKRestClient.stats() https://github.com/opensearch-project/opensearch-sdk-java/issues/620
-    @Deprecated
```
dbwiddis, line 120: Do we need to migrate a deprecated method?

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
vibrantvarun: Okay. I am replacing it with BaseExtensionRestHandler to maintain the consistency

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
vibrantvarun: ACk

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: I have to dig into this. I will post my findings here.

src/main/java/org/opensearch/ad/util/IndexUtils.java
vibrantvarun: yes it required to be uncommented because the method is been used in ADStatsTests

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
```diff
@@ -67,12 +68,7 @@
 import org.opensearch.ad.stats.suppliers.SettableSupplier;
 import org.opensearch.ad.task.ADTaskCacheManager;
 import org.opensearch.ad.task.ADTaskManager;
-import org.opensearch.ad.transport.ADJobParameterAction;
-import org.opensearch.ad.transport.ADJobParameterTransportAction;
-import org.opensearch.ad.transport.ADJobRunnerAction;
-import org.opensearch.ad.transport.ADJobRunnerTransportAction;
-import org.opensearch.ad.transport.AnomalyDetectorJobAction;
-import org.opensearch.ad.transport.AnomalyDetectorJobTransportAction;
+import org.opensearch.ad.transport.*;
```
owaiskazi19, line 71: Eh. Wildcard import

src/main/java/org/opensearch/ad/util/IndexUtils.java
vibrantvarun: I have used a boxed long here because if I don't use this then I have to add an extra check here and then store the value of the count in some other variable.

`return indicesStatsResponse.indices().get(indexName).primaries().docs().count()!=null? indicesStatsResponse.indices().get(indexName).primaries().docs().count(): -1`

src/main/java/org/opensearch/ad/util/RestHandlerUtils.java
```diff
@@ -15,12 +15,13 @@
 import static org.opensearch.rest.RestStatus.INTERNAL_SERVER_ERROR;
 
 import java.io.IOException;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
owaiskazi19, line 18: Wildcard import

src/main/java/org/opensearch/ad/util/RestHandlerUtils.java
```diff
@@ -230,4 +236,49 @@ public static boolean isProperExceptionToReturn(Throwable e) {
     private static String coalesceToEmpty(@Nullable String s) {
         return s == null ? "" : s;
     }
+
+    public static final String unrecognized(RestRequest request, Set<String> invalids, Set<String> candidates, String detail) {
```
owaiskazi19, line 240: This method was a part of `BaseRestHandler` to reduce the diff and maintain the consistency we should add it in `BaseExtensionRestHandler` if not present already and pull it from there.

src/test/java/org/opensearch/ad/transport/handler/AbstractIndexHandlerTest.java
```diff
@@ -73,6 +74,9 @@ enum IndexCreation {
     @Mock
     protected IndexNameExpressionResolver indexNameResolver;
 
+    @Mock
+    protected OpenSearchAsyncClient openSearchAsyncClient;
```
owaiskazi19, line 78: Here and above. I know it doesn't align with the variable naming but the wrapper is actually from java client which we created in SDK and for better understanding.
```suggestion
    protected OpenSearchAsyncClient javaAsyncClient;
```

src/main/java/org/opensearch/ad/transport/StatsAnomalyDetectorTransportAction.java
```diff
@@ -115,33 +120,33 @@ private ActionListener<ADStatsResponse> getRestStatsListener(ActionListener<Stat
...
             TermsAggregationBuilder termsAgg = AggregationBuilders.terms(DETECTOR_TYPE_AGG).field(AnomalyDetector.DETECTOR_TYPE_FIELD);
             SearchRequest request = new SearchRequest()
                 .indices(AnomalyDetector.ANOMALY_DETECTORS_INDEX)
                 .source(new SearchSourceBuilder().aggregation(termsAgg).size(0).trackTotalHits(true));
 
-            client.search(request, ActionListener.wrap(r -> {
-                StringTerms aggregation = r.getAggregations().get(DETECTOR_TYPE_AGG);
-                List<StringTerms.Bucket> buckets = aggregation.getBuckets();
+            sdkRestClient.search(request, ActionListener.wrap(r -> {
+                ParsedStringTerms aggregation = r.getAggregations().get(DETECTOR_TYPE_AGG);
```
owaiskazi19, line 144: Why are we changing the type here from `StringTerms` to `ParsedStringTerms`?

src/main/java/org/opensearch/ad/transport/StatsAnomalyDetectorTransportAction.java
```diff
@@ -193,16 +198,17 @@ private Map<String, Object> getClusterStatsMap(ADStatsRequest adStatsRequest) {
...
      */
     private void getNodeStats(
-        Client client,
+        SDKRestClient sdkRestClient,
         MultiResponsesDelegateActionListener<ADStatsResponse> listener,
         ADStatsRequest adStatsRequest
     ) {
-        client.execute(ADStatsNodesAction.INSTANCE, adStatsRequest, ActionListener.wrap(adStatsResponse -> {
+        ADStatsNodeRequest adStatsNodeRequest = new ADStatsNodeRequest(adStatsRequest);
+        sdkRestClient.execute(ADStatsNodesAction.INSTANCE, adStatsNodeRequest, ActionListener.wrap(adStatsResponse -> {
```
owaiskazi19, line 211: Isn't the request here should be `ADStatsRequest`? Why are we making a new one?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -11,92 +11,75 @@
...
-        List<FailedNodeException> failures
-    ) {
-        return new ADStatsNodesResponse(clusterService.getClusterName(), responses, failures);
+    public ADStatsNodesResponse newResponse(List<ADStatsNodeResponse> responses, List<FailedNodeException> failures) {
+        return new ADStatsNodesResponse(sdkClusterService.state().getClusterName(), responses, failures);
     }
 
-    @Override
-    protected ADStatsNodeRequest newNodeRequest(ADStatsRequest request) {
+    public ADStatsNodeRequest newNodeRequest(ADStatsRequest request) {
```
owaiskazi19, line 75: Why public? Are we using this method in tests?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -11,92 +11,75 @@
...
         return new ADStatsNodeRequest(request);
     }
 
-    @Override
-    protected ADStatsNodeResponse newNodeResponse(StreamInput in) throws IOException {
+    public ADStatsNodeResponse newNodeResponse(StreamInput in) throws IOException {
```
owaiskazi19, line 79: Same as above?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -11,92 +11,75 @@
...
     }
 
-    @Override
-    protected ADStatsNodesResponse newResponse(
-        ADStatsRequest request,
-        List<ADStatsNodeResponse> responses,
-        List<FailedNodeException> failures
-    ) {
-        return new ADStatsNodesResponse(clusterService.getClusterName(), responses, failures);
+    public ADStatsNodesResponse newResponse(List<ADStatsNodeResponse> responses, List<FailedNodeException> failures) {
```
owaiskazi19, line 71: Let's not change the method arguments here because the tests would be affected. Old method had `ADStatsRequest` as one of the param

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -11,92 +11,75 @@
...
 
+import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_GET_NODE_STATS;
+import static org.opensearch.ad.util.RestHandlerUtils.wrapRestActionListener;
+
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
owaiskazi19, line 18: Wildcard

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -11,64 +11,150 @@
...
+                StatsAnomalyDetectorAction.INSTANCE,
+                adStatsRequest,
+                ActionListener.wrap(response -> statsFutureResponse.complete(response), ex -> statsFutureResponse.completeExceptionally(ex))
+            );
+
+        StatsAnomalyDetectorResponse statsAnomalyDetectorResponse = statsFutureResponse
+            .orTimeout(AnomalyDetectorSettings.REQUEST_TIMEOUT.get(settings).getMillis(), TimeUnit.MILLISECONDS)
+            .join();
+
+        logger.info("Stats Response is  **********************  " + statsAnomalyDetectorResponse.toString());
```
owaiskazi19, line 152: We don't have to log the response here. Leftover?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -126,6 +109,29 @@ private ADStatsNodeResponse createADStatsNodeResponse(ADStatsRequest adStatsRequ
...
+            } else {
+                logger.info("Could not process get Node stats API due to exception", e);
+            }
+            listener.onFailure(e);
+        }
+        List<ADStatsNodeResponse> responses = new ArrayList<>();
+        if (adStatsNodeResponse != null) {
+            responses.add(adStatsNodeResponse);
+        }
+        listener.onResponse(newResponse(responses, failures));
```
owaiskazi19, line 135: Based on [this](https://github.com/opensearch-project/anomaly-detection/pull/857/files#r1160094000) we might won't need `doExecute` here.

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
```diff
@@ -13,14 +13,15 @@
...
+import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
-import org.opensearch.transport.TransportRequest;
 
 /**
  *  ADStatsNodeRequest to get a nodes stat
  */
-public class ADStatsNodeRequest extends TransportRequest {
+public class ADStatsNodeRequest extends ActionRequest {
```
owaiskazi19, line 24: Why are we changing the request here?

owaiskazi19 (changes requested):
Let's try to minimize the diff for migration by handling the new methods in SDK. This PR adds a lot of new changes. As part of migration we should focus on the above.

Also, get stats has these 2 APIs.

`GET _plugins/_anomaly_detection/stats`
`GET _plugins/_anomaly_detection/<nodeId>/stats`.

Can you add checklist in your PR description that you tested both and add the response of the APIs in the issue on SDK repo [like](https://github.com/opensearch-project/opensearch-sdk-java/issues/380#issuecomment-1500612305)?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: @dbwiddis

When ADStatsNodesTransportAction is used in

1) Plugin Architecture: It extends TransportNodesAction. The transportNodesAction Accepts 4 parameters 
(BaseNodesRequest, NodesResponse, NodeRequest, NodeResponse). It can perform Async operation on multiple nodes. 
P.S.:      NodeRequest extends TransportRequest

2) Extension Architecture: It extends TransportAction. The transport action accepts 2 parameters (NodeRequest and NodesResponse). It can only perform request on only 1 node. Therefore in order to provide the functionality similar to TransportNodesAction I can open up a new issue.

As the TransportNodesAction counterpart for extensions needs more understanding and overall scope is beyond GetDetectorStats API.

What would you suggest? Shall I open a new issue?

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
vibrantvarun: Because as per the changes made by dan in TransportRequest, it now accepts the request which extends the ActionRequest and response which extends ActionResponse

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -11,64 +11,119 @@
...
     }
 
-    @Override
     public String getName() {
         return STATS_ANOMALY_DETECTOR_ACTION;
     }
 
     @Override
-    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+    public List<RouteHandler> routeHandlers() {
```
dbwiddis, line 82: This implementation works for now and no need to hold up this PR.  But heads up that with [SDK #665](https://github.com/opensearch-project/opensearch-sdk-java/issues/665) we'll be able to go back to the replaced routs format of existing rest handler classes.  So stay tuned!

src/main/java/org/opensearch/ad/util/IndexUtils.java
dbwiddis: Got it, was thinking of if this was a new method we shouldn't do this and if it's a deprecated method we should call the replacement, but this is really just migration work on a poorly written (and now deprecated) method :)

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
dbwiddis: `ActionRequest` extends `TransportRequest` so this doesn't break anything but does add some functionality relying on `ActionRequest` signature. Is this being triggered remotely from JS, etc. and that's why we need the `ActionRequest`?

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -11,64 +11,119 @@
...
+
+        StatsAnomalyDetectorResponse statsAnomalyDetectorResponse = statsFutureResponse
+            .orTimeout(AnomalyDetectorSettings.REQUEST_TIMEOUT.get(settings).getMillis(), TimeUnit.MILLISECONDS)
+            .join();
+
+        XContentBuilder statsAnomalyDetectorResponseBuilder = statsAnomalyDetectorResponse
+            .toXContent(JsonXContent.contentBuilder(), ToXContent.EMPTY_PARAMS);
+        ExtensionRestResponse response = new ExtensionRestResponse(request, RestStatus.OK, statsAnomalyDetectorResponseBuilder);
+
+        return response;
```
dbwiddis, line 126: (Style) Why not just `return new ExtensionRestResponse(...)` here instead of creating a new object and returning it on the next line?  Yeah, the compiler optimizes this out, just reads better to me. :)

src/main/java/org/opensearch/ad/constant/CommonErrorMessages.java
```diff
@@ -94,6 +94,8 @@ public static String getTooManyCategoricalFieldErr(int limit) {
     public static String FAIL_TO_GET_STATS = "Fail to get stats";
     public static String FAIL_TO_SEARCH = "Fail to search";
 
+    public static String FAIL_TO_GET_NODE_STATS = "Fail to get node stats";
```
dbwiddis, line 97: Nit 1. I know you're coping from the previous error messages on lines 85-93, but this should be "failed" rather than "fail".  (As should all the rest.) 

Nit 2. All the others are together without whitespace in between them, this one isn't special.  Save that CR byte from the file size!

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
dbwiddis: Yeah, this is probably worth a new issue.  Summing it up:
 - `TransportNodesAction` is on OpenSearch meant for plugins, so we can't/shouldn't change it for extensions, but
 - There are some methods that add some functionality to the inheritance from `TransportAction`, specifically the `newResponse` method and the abstract methods.

I'm not sure we need a `SDKTransportNodesAction` but we should at least look into whether it makes sense.

dbwiddis (commented):
LGTM generally.  Will review again after @owaiskazi19's comments are addressed.

src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java
vibrantvarun: It is a rest action so I cannot inject the required dependency via guice. If it would have been a transport Action then I could have achieved the dependency injection.

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: As discussed, I will be opening a new issue for this.

src/main/java/org/opensearch/ad/transport/StatsAnomalyDetectorTransportAction.java
vibrantvarun: I am implementing this change due to type cast error. As we are using SDKRestClient so the aggregation is now returning ParsedStringTerms instead of StringTerms. So I have to make this change.

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
vibrantvarun: yes. Sorry it was left by mistake. Will remove it in the new iteration.

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: As of now, we need doExecute method as We have to cater all the aspects needed to implement the above referenced work.

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
vibrantvarun: No this request is not being triggered from JS

src/main/java/org/opensearch/ad/transport/StatsAnomalyDetectorTransportAction.java
owaiskazi19: Yep. As discussed we will be going in with this change. Can you add an entry in plugin migration doc to address this change? You can put it under a new title "Change data types for RestClient"

owaiskazi19: Opened the PR to address the above: https://github.com/opensearch-project/opensearch-sdk-java/pull/673

vibrantvarun:
curl -X GET "localhost:9200/_extensions/_ad-extension-1/stats/"
```
{
  "anomaly_detectors_index_status": "non-existent",
  "anomaly_detection_state_status": "non-existent",
  "single_entity_detector_count": 0,
  "detector_count": 0,
  "multi_entity_detector_count": 0,
  "anomaly_detection_job_index_status": "non-existent",
  "models_checkpoint_index_status": "non-existent",
  "anomaly_results_index_status": "non-existent",
  "nodes": {
    "ad-extension-1": {
      "ad_execute_request_count": 0,
      "models": [
        
      ],
      "ad_canceled_batch_task_count": 0,
      "ad_hc_execute_request_count": 0,
      "ad_hc_execute_failure_count": 0,
      "model_count": 0,
      "ad_execute_failure_count": 0,
      "ad_batch_task_failure_count": 0,
      "ad_total_batch_task_execution_count": 0,
      "ad_executing_batch_task_count": 0
    }
  }
}
```

vibrantvarun:
curl -X GET "localhost:9200/_extensions/_ad-extension-1/stats/ad_hc_execute_failure_count"
 ```
 {
  "nodes": {
    "ad-extension-1": {
      "ad_hc_execute_failure_count": 0
    }
  }
}
 ```

vibrantvarun:
curl -X GET "localhost:9200/_extensions/_ad-extension-1/_ad-extension-1/stats/ad_hc_execute_failure_count"

```
{
  "nodes": {
    "ad-extension-1": {
      "ad_hc_execute_failure_count": 0
    }
  }
}

```

joshpalis:
@vibrantvarun I am curious to see the behavior once a detector is created and started. You can refer to [this guide](https://github.com/opensearch-project/opensearch-sdk-java/issues/682) on how to generate test data and how to create/run a detector

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
joshpalis: Same question as Owais, if its not necessary to change then lets revert this change

dbwiddis (approved):
LGTM. I see some comments from @owaiskazi19 that should be answered.

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
vibrantvarun: As suggested by @dbwiddis I will open a new issue where we will cater the blast radius of TransportNodesAction and will do the required work there.

vibrantvarun:
Response of detector stats after triggering create detector api

```
{
  "anomaly_detectors_index_status": "yellow",
  "anomaly_detection_state_status": "non-existent",
  "single_entity_detector_count": 1,
  "detector_count": 1,
  "multi_entity_detector_count": 0,
  "anomaly_detection_job_index_status": "non-existent",
  "models_checkpoint_index_status": "non-existent",
  "anomaly_results_index_status": "non-existent",
  "nodes": {
    "ad-extension-1": {
      "ad_execute_request_count": 0,
      "models": [
        
      ],
      "ad_canceled_batch_task_count": 0,
      "ad_hc_execute_request_count": 0,
      "ad_hc_execute_failure_count": 0,
      "model_count": 0,
      "ad_execute_failure_count": 0,
      "ad_batch_task_failure_count": 0,
      "ad_total_batch_task_execution_count": 0,
      "ad_executing_batch_task_count": 0
    }
  }
}


```

joshpalis (approved):

owaiskazi19:
@vibrantvarun can you add the response and API in the issue like I mentioned above?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: No we are not using it in test class as of now. But this method is used in the method below **doExecute**

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: No this method also we are not using anywhere. But it will be used in integtest in future.

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
owaiskazi19: Did you get any error if you keep the request as `TransportRequest`?

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
owaiskazi19: Let's add a TODO and link the issue here.

src/main/java/org/opensearch/ad/util/RestHandlerUtils.java
```diff
@@ -81,6 +81,10 @@ public final class RestHandlerUtils {
     public static final String RESULTS = "results";
     public static final String TOP_ANOMALIES = "_topAnomalies";
     public static final String VALIDATE = "_validate";
+
+    public static final String NODE_ID = "nodeId";
+
+    public static final String STAT = "stat";
```
owaiskazi19, line 87: We generally add the constant for rest handler URI but it's fine to add for request param as well. Can you remove the extra lines here?

src/test/java/org/opensearch/ad/stats/ADStatsTests.java
```diff
@@ -71,16 +109,22 @@ public void setup() {
...
+        ExtensionsRunner mockRunner = mock(ExtensionsRunner.class);
+        Extension mockExtension = mock(Extension.class);
+        when(mockRunner.getEnvironmentSettings()).thenReturn(settings);
+        when(mockRunner.getExtension()).thenReturn(mockExtension);
+        when(mockExtension.getSettings()).thenReturn(settingsList);
+        SDKClusterService sdkClusterService = new SDKClusterService(mockRunner);
+        // SDKClusterService.SDKClusterSettings clusterSettings = sdkClusterService.new SDKClusterSettings(
+        // Settings.EMPTY, Collections.unmodifiableSet(new HashSet<>(Arrays.asList(MAX_MODEL_SIZE_PER_NODE)))
+        // );
+        // when(sdkClusterService.getClusterSettings()).thenReturn(clusterSettings);
```
owaiskazi19, line 122: Do we need this commented code here anymore?

src/test/java/org/opensearch/ad/transport/handler/AbstractIndexHandlerTest.java
```diff
@@ -73,6 +74,9 @@ enum IndexCreation {
     @Mock
     protected IndexNameExpressionResolver indexNameResolver;
 
+    @Mock
+    protected OpenSearchAsyncClient javaAsyncClient;
```
owaiskazi19, line 78: Can you mock the client in `setUp` like:

`sdkJavaAsyncClient = mock(OpenSearchAsyncClient.class);`

owaiskazi19 (changes requested):
Almost there. Plus the comments above

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
owaiskazi19: Why do we need to make it as `public` then? `protected` should do

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
owaiskazi19: Let's keep it as `protected` then

src/test/java/org/opensearch/ad/transport/handler/AbstractIndexHandlerTest.java
vibrantvarun: ACk

src/test/java/org/opensearch/ad/stats/ADStatsTests.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/transport/ADStatsNodeRequest.java
vibrantvarun: yes I got an error, TransportAction only accepts the requests which extends ActionRequest.

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: https://github.com/opensearch-project/opensearch-sdk-java/issues/683

vibrantvarun: Done added in PR as well

src/main/java/org/opensearch/ad/transport/StatsAnomalyDetectorTransportAction.java
vibrantvarun: ADStatsNodeRequest is created because:
1) The nodeOperationMethod in [ADStatsNodesTransportAction](https://github.com/joshpalis/anomaly-detection/blob/fd547014fdde5114bbc9c8e49fe7aaa37eb6e793/src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java#L100) take ADStatsNodeRequest in as argument. 
2) We have called nodeOperationMethod from doExecute  method.  I don't wanted to make any change in existing methods in order to maintain the sanity of the work. So I created ADStatsNodeRequest in the above method and passed it in the transportAction call.

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
vibrantvarun: gotcha

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
owaiskazi19: Let's change the request to `ADStatsRequest`

src/main/java/org/opensearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -79,10 +154,10 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
      */
     private ADStatsRequest getRequest(RestRequest request) {
         // parse the nodes the user wants to query the stats for
-        String nodesIdsStr = request.param("nodeId");
+        String nodesIdsStr = request.param(NODE_ID);
         Set<String> validStats = adStats.getStats().keySet();
 
-        ADStatsRequest adStatsRequest = null;
+        ADStatsRequest adStatsRequest;
```
owaiskazi19, line 160: You can initialize with `null`

src/main/java/org/opensearch/ad/transport/ADStatsNodesTransportAction.java
```diff
@@ -12,93 +12,82 @@
...
 import org.opensearch.monitor.jvm.JvmService;
-import org.opensearch.threadpool.ThreadPool;
-import org.opensearch.transport.TransportService;
+import org.opensearch.sdk.SDKClusterService;
+import org.opensearch.tasks.Task;
+import org.opensearch.tasks.TaskManager;
 
+import com.google.inject.Inject;
+
+// TODO: https://github.com/opensearch-project/opensearch-sdk-java/issues/683 (multi node support for extensions)
```
owaiskazi19, line 36: Can you add more details on why multi node support is needed here?

owaiskazi19 (commented):
Looks good. 2 small suggestions

src/test/java/org/opensearch/ad/transport/handler/AbstractIndexHandlerTest.java
```diff
@@ -12,8 +12,7 @@
 package org.opensearch.ad.transport.handler;
 
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
owaiskazi19, line 15: Wildcard import :(

src/test/java/org/opensearch/ad/transport/handler/AbstractIndexHandlerTest.java
```diff
@@ -73,6 +73,9 @@ enum IndexCreation {
     @Mock
     protected IndexNameExpressionResolver indexNameResolver;
 
+    @Mock
+    protected static OpenSearchAsyncClient sdkJavaAsyncClient;
```
owaiskazi19, line 77: You are mocking it already below. No need to mock here. Just move this after line 60 and say:
```
protected  OpenSearchAsyncClient sdkJavaAsyncClient;

owaiskazi19 (approved):
LGTM! Thanks for addressing the feedback

PR #890, opened 2023-05-03
Execute Detector

vibrantvarun:
### Description
Execute Detector

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
+++ b/src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
@@ -11,80 +11,106 @@
 
 package org.opensearch.ad.rest;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import org.opensearch.action.ActionListener;
+import org.opensearch.ad.AnomalyDetectorExtension;
+import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.REQUEST_TIMEOUT;
+import org.opensearch.ad.transport.AnomalyResultResponse;
 import static org.opensearch.ad.util.RestHandlerUtils.DETECTOR_ID;
 import static org.opensearch.ad.util.RestHandlerUtils.RUN;
+import org.opensearch.common.settings.Settings;
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
-
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
-
 import org.apache.commons.lang.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.opensearch.ad.AnomalyDetectorPlugin;
 import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.model.AnomalyDetectorExecutionInput;
 import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.ad.transport.AnomalyResultAction;
 import org.opensearch.ad.transport.AnomalyResultRequest;
-import org.opensearch.client.node.NodeClient;
-import org.opensearch.cluster.service.ClusterService;
-import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.common.xcontent.json.JsonXContent;
+import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.core.xcontent.XContentParser;
-import org.opensearch.rest.BaseRestHandler;
-import org.opensearch.rest.BytesRestResponse;
+import org.opensearch.extensions.rest.ExtensionRestResponse;
 import org.opensearch.rest.RestRequest;
 import org.opensearch.rest.RestStatus;
-import org.opensearch.rest.action.RestToXContentListener;
-
 import com.google.common.collect.ImmutableList;
+import org.opensearch.sdk.ExtensionsRunner;
+import org.opensearch.sdk.SDKClient.SDKRestClient;
+import org.opensearch.sdk.rest.BaseExtensionRestHandler;
+import org.opensearch.sdk.rest.ReplacedRouteHandler;
 
 /**
  * This class consists of the REST handler to handle request to detect data.
  */
-public class RestExecuteAnomalyDetectorAction extends BaseRestHandler {
+public class RestExecuteAnomalyDetectorAction extends BaseExtensionRestHandler {
 
     public static final String DETECT_DATA_ACTION = "execute_anomaly_detector";
     // TODO: apply timeout config
     private volatile TimeValue requestTimeout;
-
+    private SDKRestClient sdkRestClient;
+    private Settings settings;
     private final Logger logger = LogManager.getLogger(RestExecuteAnomalyDetectorAction.class);
 
-    public RestExecuteAnomalyDetectorAction(Settings settings, ClusterService clusterService) {
-        this.requestTimeout = REQUEST_TIMEOUT.get(settings);
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(REQUEST_TIMEOUT, it -> requestTimeout = it);
+    public RestExecuteAnomalyDetectorAction(ExtensionsRunner extensionsRunner, SDKRestClient sdkRestClient) {
+        this.requestTimeout = REQUEST_TIMEOUT.get(extensionsRunner.getEnvironmentSettings());
+        this.sdkRestClient=sdkRestClient;
+        extensionsRunner.getSdkClusterService().getClusterSettings().addSettingsUpdateConsumer(REQUEST_TIMEOUT, it -> requestTimeout = it);
+        this.settings=extensionsRunner.getEnvironmentSettings();
     }
 
-    @Override
     public String getName() {
         return DETECT_DATA_ACTION;
     }
 
-    @Override
-    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+    protected ExtensionRestResponse prepareRequest(RestRequest request) throws IOException {
         if (!EnabledSetting.isADPluginEnabled()) {
             throw new IllegalStateException(CommonErrorMessages.DISABLED_ERR_MSG);
         }
         AnomalyDetectorExecutionInput input = getAnomalyDetectorExecutionInput(request);
-        return channel -> {
-            String rawPath = request.rawPath();
-            String error = validateAdExecutionInput(input);
-            if (StringUtils.isNotBlank(error)) {
-                channel.sendResponse(new BytesRestResponse(RestStatus.BAD_REQUEST, error));
-                return;
-            }
 
-            AnomalyResultRequest getRequest = new AnomalyResultRequest(
+        String rawPath = request.rawPath();
+        String error = validateAdExecutionInput(input);
+
+        if (StringUtils.isNotBlank(error)) {
+            return getExecuteDetectorResponse(RestStatus.BAD_REQUEST,request,null,error);
+        }
+
+        CompletableFuture<AnomalyResultResponse> futureResponse = new CompletableFuture<>();
+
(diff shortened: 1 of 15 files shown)


src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -115,21 +146,36 @@ private String validateAdExecutionInput(AnomalyDetectorExecutionInput input) {
...
     }
+
+    private ExtensionRestResponse getExecuteDetectorResponse(
+        RestStatus restStatus,
+        RestRequest request,
+        AnomalyResultResponse response,
+        String content
+    ) throws IOException {
+        ExtensionRestResponse extensionRestResponse;
+        if (content.isEmpty()) {
```
joshpalis, line 170: just curious, what's the point of checking if the content is empty if we're only using this method once and passing "" for the content.

src/main/java/org/opensearch/ad/transport/handler/AnomalyResultBulkIndexHandler.java
```diff
@@ -42,15 +42,24 @@ public class AnomalyResultBulkIndexHandler extends AnomalyIndexHandler<AnomalyRe
...
-        super(client, settings, threadPool, ANOMALY_RESULT_INDEX_ALIAS, anomalyDetectionIndices, clientUtil, indexUtils, clusterService);
+        super(
+            sdkRestClient,
+            settings,
+            threadPool,
+            ANOMALY_RESULT_INDEX_ALIAS,
+            anomalyDetectionIndices,
+            clientUtil,
+            indexUtils,
+            clusterService
```
joshpalis, line 61: missed a change to `sdkClusterService` here

src/main/java/org/opensearch/ad/transport/handler/MultiEntityResultHandler.java
```diff
@@ -73,7 +73,7 @@ public MultiEntityResultHandler(
      * @param listener callback after flushing
      */
     public void flush(ADResultBulkRequest currentBulkRequest, ActionListener<ADResultBulkResponse> listener) {
-        if (indexUtils.checkIndicesBlocked(clusterService.state(), ClusterBlockLevel.WRITE, this.indexName)) {
+        if (indexUtils.checkIndicesBlocked(sdkClusterService.state(), ClusterBlockLevel.WRITE, this.indexName)) {
```
joshpalis, line 76: seems you have missed the clusterService name change in the constructor for the `MultiEntityResultHandler` class, curious as to how this passed the build

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
vibrantvarun: The reason I have added a content parameter is because as per the previous code written for plugins, if we find the error in validating request then it throws a error after setting it in ByteResponse. So from wherever this api would have been called different type of error handling would be done. I have passed "" placeholder for content when the api execution is done without any error.

src/main/java/org/opensearch/ad/transport/handler/AnomalyResultBulkIndexHandler.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/transport/handler/MultiEntityResultHandler.java
vibrantvarun: ACk

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
joshpalis: Ah I missed that first invocation of the method after checking for errors. LGTM

joshpalis (approved):

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
owaiskazi19: We don't need an extra check here for content. If there's any error in the response the code execution would go to the catch block [here](https://github.com/opensearch-project/anomaly-detection/pull/890/files#diff-de43e12dd14419a5047d6b20decd5f99683bf094c7d0331f0cebbe08a74b92f8R116). Also, we need to handle the error by checking the correct exception. Passing `""` is not the correct way to do so. I would say let's handle the exact error throwing in the future not for this API but for all the other we have so far and create an issue for the same.

saratvemulapalli (approved):
Lots of changes are renaming the client variable and cluster service. 
For future I would have dedicated PRs for refactoring and Execute Detector. 

Are there tests (unit) which we can uncomment for execute detector?

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
vibrantvarun: So reduce the diff in implementation and send the right error to the user, I have added a piece of [code](https://github.com/opensearch-project/anomaly-detection/pull/890/files#diff-de43e12dd14419a5047d6b20decd5f99683bf094c7d0331f0cebbe08a74b92f8R86) here.  This code snippet replicate the [earlier implementation](https://github.com/opensearch-project/anomaly-detection/pull/890/files#diff-de43e12dd14419a5047d6b20decd5f99683bf094c7d0331f0cebbe08a74b92f8L76) in returning the response. 

I am returning "" [here](https://github.com/opensearch-project/anomaly-detection/pull/890/files#diff-de43e12dd14419a5047d6b20decd5f99683bf094c7d0331f0cebbe08a74b92f8R108)  because I have created a common method [below](https://github.com/opensearch-project/anomaly-detection/pull/890/files#diff-de43e12dd14419a5047d6b20decd5f99683bf094c7d0331f0cebbe08a74b92f8R163) by following dry principle as I am using the same piece of code at 2 places. That is reason why I have put a check of content.isEmpty in the method.

owaiskazi19: You can check for 
1. the RestStatus if it's OK or a BAD_REQUEST
2. response since you are already passing `null` for error. 
I would prefer the check for RestStatus.  For the successful response you can pass the content as `response.toXContent(JsonXContent.contentBuilder(), ToXContent.EMPTY_PARAMS)` instead of "".

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -115,21 +146,36 @@ private String validateAdExecutionInput(AnomalyDetectorExecutionInput input) {
...
+            extensionRestResponse = new ExtensionRestResponse(
```
owaiskazi19, line 171: You can just return from here. We can avoid having a separate variable here and in the else too.

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
vibrantvarun: cool

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -19,73 +19,104 @@
...
     private final Logger logger = LogManager.getLogger(RestExecuteAnomalyDetectorAction.class);
 
-    public RestExecuteAnomalyDetectorAction(Settings settings, ClusterService clusterService) {
-        this.requestTimeout = REQUEST_TIMEOUT.get(settings);
-        clusterService.getClusterSettings().addSettingsUpdateConsumer(REQUEST_TIMEOUT, it -> requestTimeout = it);
+    public RestExecuteAnomalyDetectorAction(ExtensionsRunner extensionsRunner, SDKRestClient sdkRestClient) {
+        this.requestTimeout = REQUEST_TIMEOUT.get(extensionsRunner.getEnvironmentSettings());
+        this.sdkRestClient = sdkRestClient;
+        extensionsRunner.getSdkClusterService().getClusterSettings().addSettingsUpdateConsumer(REQUEST_TIMEOUT, it -> requestTimeout = it);
+        this.settings = extensionsRunner.getEnvironmentSettings();
```
owaiskazi19, line 69: You can define this at the top and use it for 
`this.requestTimeout = REQUEST_TIMEOUT.get(settings);`

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -19,73 +19,104 @@
...
+            .execute(
+                AnomalyResultAction.INSTANCE,
+                anomalyResultRequest,
+                ActionListener.wrap(r -> futureResponse.complete(r), e -> futureResponse.completeExceptionally(e))
             );
-            client.execute(AnomalyResultAction.INSTANCE, getRequest, new RestToXContentListener<>(channel));
-        };
+
+        AnomalyResultResponse anomalyResultResponse = futureResponse
+            .orTimeout(AnomalyDetectorSettings.REQUEST_TIMEOUT.get(settings).getMillis(), TimeUnit.MILLISECONDS)
```
owaiskazi19, line 105: Since we have `this.requestTimeout` already defined in the constructor. Fetch mills directly from the variable:
```suggestion
            .orTimeout(this.requestTimeout.getMillis(), TimeUnit.MILLISECONDS)
```

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
vibrantvarun: Changes pushed

vibrantvarun:
@saratvemulapalli no test cases can be uncomment or unignored because it needs opensearch test suite conifgured for extensions and hashring support.

src/main/java/org/opensearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -19,73 +19,102 @@
...
+                ActionListener.wrap(r -> futureResponse.complete(r), e -> futureResponse.completeExceptionally(e))
             );
-            client.execute(AnomalyResultAction.INSTANCE, getRequest, new RestToXContentListener<>(channel));
-        };
+
+        AnomalyResultResponse anomalyResultResponse = futureResponse
+            .orTimeout(this.requestTimeout.getMillis(), TimeUnit.MILLISECONDS)
+            .join();
+
+        return getExecuteDetectorResponse(RestStatus.OK, request, anomalyResultResponse, anomalyResultResponse.toString());
```
owaiskazi19, line 106: Creating a new method is not helping here because you are passing an extra param `anomalyResultResponse.toString()` which is not required. Passing `null` is also not a good option. I would say remove `getExecuteDetectorResponse` method and return `ExtensionRestResponse` directly from here and same for the error case above as well.

owaiskazi19: Or if you want to keep the method pass `null` for `content` here.

vibrantvarun: Umm, good idea. I was trying to keep consistency with other Rest actions so I kept like that. I previously had the same idea. But then thought I should play around much with the code as It should look similar with other rest actions.

vibrantvarun: Anyways I will implement what you said above, as I completely agree with your opinion

vibrantvarun: Okay so now I have pushed the changes.

owaiskazi19 (approved):

PR #891, opened 2023-05-03
[Extensions] Added tests for PriorityCache, ADTaskProfileTransportActionTests and increased code coverage

owaiskazi19:
### Description
First PR to add unit tests back for AD Extension.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/723

### DIFF
--- a/src/main/java/org/opensearch/ad/settings/EnabledSetting.java
+++ b/src/main/java/org/opensearch/ad/settings/EnabledSetting.java
@@ -41,8 +41,6 @@ public class EnabledSetting extends AbstractSetting {
 
     public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED = "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";
 
-    public static final String DOOR_KEEPER_IN_CACHE_ENABLED = "plugins.anomaly_detection.door_keeper_in_cache.enabled";;
-
     public static final Map<String, Setting<?>> settings = unmodifiableMap(new HashMap<String, Setting<?>>() {
         {
             Setting LegacyADPluginEnabledSetting = Setting
--- a/src/main/java/org/opensearch/ad/settings/EnabledSetting.java
+++ b/src/main/java/org/opensearch/ad/settings/EnabledSetting.java
@@ -39,8 +39,9 @@ public class EnabledSetting extends AbstractSetting {
 
     public static final String LEGACY_OPENDISTRO_AD_BREAKER_ENABLED = "opendistro.anomaly_detection.breaker.enabled";
 
-    public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED =
-        "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";;
+    public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED = "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";
+
+    public static final String DOOR_KEEPER_IN_CACHE_ENABLED = "plugins.anomaly_detection.door_keeper_in_cache.enabled";;
 
     public static final Map<String, Setting<?>> settings = unmodifiableMap(new HashMap<String, Setting<?>>() {
         {
--- a/src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
+++ b/src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
@@ -52,7 +52,6 @@ import org.opensearch.ad.ml.ModelState;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.Entity;
 import org.opensearch.ad.settings.AnomalyDetectorSettings;
-import org.opensearch.ad.settings.EnabledSetting;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.ByteSizeValue;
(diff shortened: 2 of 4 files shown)


src/main/java/org/opensearch/ad/settings/EnabledSetting.java
```diff
@@ -39,8 +39,9 @@ public class EnabledSetting extends AbstractSetting {
 
     public static final String LEGACY_OPENDISTRO_AD_BREAKER_ENABLED = "opendistro.anomaly_detection.breaker.enabled";
 
-    public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED =
-        "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";;
+    public static final String INTERPOLATION_IN_HCAD_COLD_START_ENABLED = "plugins.anomaly_detection.hcad_cold_start_interpolation.enabled";
+
+    public static final String DOOR_KEEPER_IN_CACHE_ENABLED = "plugins.anomaly_detection.door_keeper_in_cache.enabled";;
```
saratvemulapalli, line 44: Is this new (merge from main)? 
I dont see it being used in the PR.

src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
```diff
@@ -36,24 +89,22 @@ public void setUp() throws Exception {
...
-            Collections
-                .unmodifiableSet(
-                    new HashSet<>(
-                        Arrays
-                            .asList(
-                                AnomalyDetectorSettings.DEDICATED_CACHE_SIZE,
-                                AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE,
-                                AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE,
-                                AnomalyDetectorSettings.CHECKPOINT_TTL,
-                                AnomalyDetectorSettings.CHECKPOINT_SAVING_FREQ
```
saratvemulapalli, line 51: Curious why these settings are not tested in the new changes?

src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
```diff
@@ -656,24 +696,4 @@ public void testGetNoPriorityUpdate() {
...
-
-        assertTrue(null != entityCache.get(entity2.getModelId(detectorId).get(), detector));
-
-        entityCache.removeEntityModel(detectorId, entity2.getModelId(detectorId).get());
-
-        assertTrue(null == entityCache.get(entity2.getModelId(detectorId).get(), detector));
-
-        verify(checkpoint, times(1)).deleteModelCheckpoint(eq(entity2.getModelId(detectorId).get()), any());
-        verify(checkpointWriteQueue, never()).write(any(), anyBoolean(), any());
-    }
```
saratvemulapalli, line 677: Why are these changes removed, I dont understand what they do (yet) but shouldn't we test them?

saratvemulapalli (commented):
Great to see unit tests working for AD Extension. 

Can we start to uncomment code coverage checks in `build.gradle` for these files?

src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
owaiskazi19: These settings were added to `main` after we created `feature/extensions` [here](https://github.com/opensearch-project/anomaly-detection/pull/620/files#diff-00c99f214d22b6de9c5e592fc22fed35b542a8afc7f6b9d23c93e4efd0c2a5c0R103). Probably this test was fetched from `main` because it doesn't incline with the code changes of `feature/extensions`.

src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
owaiskazi19: Same as above

src/main/java/org/opensearch/ad/settings/EnabledSetting.java
owaiskazi19: Removed. This was coming from `main`

dbwiddis (approved):
Approving with the caveat that I think we're trying to catch up as much as we can with our current branch, but at some point will need to re-fork and that will catch up with other additions since then.

src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
owaiskazi19: These settings on feature/extensions are of the type [Duration](https://github.com/opensearch-project/anomaly-detection/blob/feature/extensions/src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java#L183) whereas based on the above PR the type got changed to [Setting](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/settings/AnomalyDetectorSettings.java#L196). Changing the type would require to fetch more code changes from `main` which we will be doing anyway with the re-fork.
The reason of the conflict here is the code on feature/extenions was from the time when we forked but probably while commenting the code for these tests, we pulled the code from main for the tests.

src/test/java/org/opensearch/ad/caching/PriorityCacheTests.java
```diff
@@ -20,7 +72,7 @@ public class PriorityCacheTests extends AbstractCacheTest {
     CheckpointDao checkpoint;
     ModelManager modelManager;
 
-    ClusterService clusterService;
+    SDKClusterService clusterService;
```
vibrantvarun, line 75: clusterService -> sdkClusterService

joshpalis (approved):

Issue #866, opened 2023-04-15
[FEATURE] refactor common settings, names, etc.


kaituo:
**Is your feature request related to a problem?**
refactor common settings, names, etc. used in both forecasting and AD

PR #902, opened 2023-05-12
Code Refactoring for CommonMessages

kaituo:
### Description
In this pull request, I have refactored the code related to shared log or error messages in both AD and forecasting modules to CommonMessages. Additionally, the previously used CommonErrorMessages has been renamed to ADCommonMessages. For the Forecasting module, I have introduced new names in ForecastCommonMessages.

Testing done:
* gradle build

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/866

### DIFF
--- a/src/main/java/org/opensearch/ad/AnomalyDetectorProfileRunner.java
+++ b/src/main/java/org/opensearch/ad/AnomalyDetectorProfileRunner.java
@@ -11,11 +11,11 @@
 
 package org.opensearch.ad;
 
-import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG;
-import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_PARSE_DETECTOR_MSG;
+import static org.opensearch.ad.constant.ADCommonMessages.FAIL_TO_PARSE_DETECTOR_MSG;
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.rest.RestStatus.BAD_REQUEST;
 import static org.opensearch.rest.RestStatus.INTERNAL_SERVER_ERROR;
+import static org.opensearch.timeseries.constant.CommonMessages.FAIL_TO_FIND_CONFIG_MSG;
 
 import java.util.List;
 import java.util.Map;
@@ -33,8 +33,8 @@ import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.ad.common.exception.NotSerializedADExceptionName;
 import org.opensearch.ad.common.exception.ResourceNotFoundException;
+import org.opensearch.ad.constant.ADCommonMessages;
 import org.opensearch.ad.constant.ADCommonName;
-import org.opensearch.ad.constant.CommonErrorMessages;
 import org.opensearch.ad.model.ADTaskType;
 import org.opensearch.ad.model.AnomalyDetector;
 import org.opensearch.ad.model.AnomalyDetectorJob;
@@ -109,7 +109,7 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
 
     public void profile(String detectorId, ActionListener<DetectorProfile> listener, Set<DetectorProfileName> profilesToCollect) {
         if (profilesToCollect.isEmpty()) {
-            listener.onFailure(new IllegalArgumentException(CommonErrorMessages.EMPTY_PROFILES_COLLECT));
+            listener.onFailure(new IllegalArgumentException(ADCommonMessages.EMPTY_PROFILES_COLLECT));
             return;
         }
         calculateTotalResponsesToWait(detectorId, profilesToCollect, listener);
@@ -136,11 +136,11 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                     listener.onFailure(new OpenSearchStatusException(FAIL_TO_PARSE_DETECTOR_MSG + detectorId, BAD_REQUEST));
                 }
             } else {
-                listener.onFailure(new OpenSearchStatusException(FAIL_TO_FIND_DETECTOR_MSG + detectorId, BAD_REQUEST));
+                listener.onFailure(new OpenSearchStatusException(FAIL_TO_FIND_CONFIG_MSG + detectorId, BAD_REQUEST));
             }
         }, exception -> {
-            logger.error(FAIL_TO_FIND_DETECTOR_MSG + detectorId, exception);
-            listener.onFailure(new OpenSearchStatusException(FAIL_TO_FIND_DETECTOR_MSG + detectorId, INTERNAL_SERVER_ERROR));
+            logger.error(FAIL_TO_FIND_CONFIG_MSG + detectorId, exception);
+            listener.onFailure(new OpenSearchStatusException(FAIL_TO_FIND_CONFIG_MSG + detectorId, INTERNAL_SERVER_ERROR));
         }));
     }
 
@@ -207,7 +207,7 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                         new MultiResponsesDelegateActionListener<DetectorProfile>(
                             listener,
                             totalResponsesToWait,
-                            CommonErrorMessages.FAIL_FETCH_ERR_MSG + detectorId,
+                            ADCommonMessages.FAIL_FETCH_ERR_MSG + detectorId,
                             false
                         );
                     if (profilesToCollect.contains(DetectorProfileName.ERROR)) {
@@ -266,7 +266,7 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                     }
 
                 } catch (Exception e) {
-                    logger.error(CommonErrorMessages.FAIL_TO_GET_PROFILE_MSG, e);
+                    logger.error(ADCommonMessages.FAIL_TO_GET_PROFILE_MSG, e);
                     listener.onFailure(e);
                 }
             } else {
@@ -277,7 +277,7 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                 logger.info(exception.getMessage());
                 onGetDetectorForPrepare(detectorId, listener, profilesToCollect);
             } else {
-                logger.error(CommonErrorMessages.FAIL_TO_GET_PROFILE_MSG + detectorId);
+                logger.error(ADCommonMessages.FAIL_TO_GET_PROFILE_MSG + detectorId);
                 listener.onFailure(exception);
             }
         }));
@@ -304,7 +304,7 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                     DetectorProfile profile = profileBuilder.totalEntities(value).build();
                     listener.onResponse(profile);
                 }, searchException -> {
-                    logger.warn(CommonErrorMessages.FAIL_TO_GET_TOTAL_ENTITIES + detector.getDetectorId());
+                    logger.warn(ADCommonMessages.FAIL_TO_GET_TOTAL_ENTITIES + detector.getDetectorId());
                     listener.onFailure(searchException);
                 });
                 // using the original context in listener as user roles have no permissions for internal operations like fetching a
@@ -353,7 +353,7 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                     DetectorProfile profile = profileBuilder.totalEntities(Long.valueOf(compositeAgg.getBuckets().size())).build();
                     listener.onResponse(profile);
                 }, searchException -> {
-                    logger.warn(CommonErrorMessages.FAIL_TO_GET_TOTAL_ENTITIES + detector.getDetectorId());
+                    logger.warn(ADCommonMessages.FAIL_TO_GET_TOTAL_ENTITIES + detector.getDetectorId());
                     listener.onFailure(searchException);
                 });
                 // using the original context in listener as user roles have no permissions for internal operations like fetching a
--- a/src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
+++ b/src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
@@ -45,7 +45,6 @@ public class ForecastCommonMessages {
     // ======================================
     // Used for custom forecast result index
     // ======================================
-    public static String CAN_NOT_FIND_RESULT_INDEX = "Can't find result index ";
     public static String INVALID_RESULT_INDEX_PREFIX = "Result index must start with " + CUSTOM_RESULT_INDEX_PREFIX;
     public static String INVALID_CHAR_IN_RESULT_INDEX_NAME =
         "Result index name has invalid character. Valid characters are a-z, 0-9, -(hyphen) and _(underscore)";
--- a/src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
+++ b/src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
@@ -1,0 +1,62 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.forecast.constant;
+
+import static org.opensearch.forecast.constant.ForecastCommonName.CUSTOM_RESULT_INDEX_PREFIX;
+
+public class ForecastCommonMessages {
+    // ======================================
+    // Validation message
+    // ======================================
+    public static String INVALID_FORECAST_INTERVAL = "Forecast interval must be a positive integer";
+    public static String NULL_FORECAST_INTERVAL = "Forecast interval should be set";
+    public static String INVALID_FORECASTER_NAME =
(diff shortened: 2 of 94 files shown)


src/main/java/org/opensearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -11,11 +11,11 @@
 
 package org.opensearch.ad;
 
-import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG;
-import static org.opensearch.ad.constant.CommonErrorMessages.FAIL_TO_PARSE_DETECTOR_MSG;
+import static org.opensearch.ad.constant.ADCommonMessages.FAIL_TO_PARSE_DETECTOR_MSG;
```
sudiptoguha, line 14: What if there is an artifact which is not a detector? Say a clustering, or a density/hot-spot measurement? Maybe have "FAIL_TO_PARSE_ARTEFACT_MSG" ?

src/main/java/org/opensearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -11,11 +11,11 @@
...
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.rest.RestStatus.BAD_REQUEST;
 import static org.opensearch.rest.RestStatus.INTERNAL_SERVER_ERROR;
+import static org.opensearch.timeseries.constant.CommonMessages.FAIL_TO_FIND_CONFIG_MSG;
```
sudiptoguha, line 18: Good name.

src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
```diff
@@ -0,0 +1,62 @@
...
+    // Resource constraints
+    // ======================================
+    public static final String DISABLED_ERR_MSG = "Forecast functionality is disabled. To enable update plugins.forecast.enabled to true";
+
+    // ======================================
+    // RESTful API
+    // ======================================
+    public static String FAIL_TO_CREATE_FORECASTER = "Fail to create forecaster";
+    public static String FAIL_TO_UPDATE_FORECASTER = "Fail to update forecaster";
+    public static String FAIL_TO_FIND_FORECASTER_MSG = "Can't find forecaster with id: ";
```
sudiptoguha, line 35: Nit: maybe use "Can not..." -- if OpenSearch is translated using wrappers, one can see the English abbreviations being one more point of worry.

src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
```diff
@@ -0,0 +1,62 @@
...
+
+    // ======================================
+    // Resource constraints
+    // ======================================
+    public static final String DISABLED_ERR_MSG = "Forecast functionality is disabled. To enable update plugins.forecast.enabled to true";
+
+    // ======================================
+    // RESTful API
+    // ======================================
+    public static String FAIL_TO_CREATE_FORECASTER = "Fail to create forecaster";
```
sudiptoguha, line 33: Nit: "Failed"

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+
+package org.opensearch.timeseries.constant;
+
+import java.util.Locale;
+
+public class CommonMessages {
+    // ======================================
+    // Validation message
+    // ======================================
+    public static String NEGATIVE_TIME_CONFIGURATION = "should be non-negative";
```
sudiptoguha, line 20: Nit: Maybe say "Time <something> should be ..." --- that something is likely of interest.

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static String INVALID_RESULT_INDEX_MAPPING = "Result index mapping is not correct for index: ";
+    public static String EMPTY_NAME = "name should be set";
+    public static String NULL_TIME_FIELD = "Time field should be set";
+    public static String EMPTY_INDICES = "Indices should be set";
+
+    public static String getTooManyCategoricalFieldErr(int limit) {
+        return String.format(Locale.ROOT, CommonMessages.TOO_MANY_CATEGORICAL_FIELD_ERR_MSG_FORMAT, limit);
+    }
+
+    public static final String TOO_MANY_CATEGORICAL_FIELD_ERR_MSG_FORMAT = "We can have only %d categorical field/s.";
```
sudiptoguha, line 30: Not a nit. "Currently we only support up to %d categorical ..." -- may be even have "in order to bound system resource consumption." That would explain that there is no magic stopping this from happening. This should empower folks who have pressing needs to voice it.

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static String CAN_NOT_FIND_RESULT_INDEX = "Can't find result index ";
+    public static String INVALID_CHAR_IN_RESULT_INDEX_NAME =
+        "Result index name has invalid character. Valid characters are a-z, 0-9, -(hyphen) and _(underscore)";
+    public static String FAIL_TO_VALIDATE = "fail to validate";
+    public static String INVALID_TIMESTAMP = "Timestamp field: (%s) must be of type date";
```
sudiptoguha, line 35: Is there a default/preferred date format? Something that would avoid conversion back and forth ...

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public static String getTooManyCategoricalFieldErr(int limit) {
+        return String.format(Locale.ROOT, CommonMessages.TOO_MANY_CATEGORICAL_FIELD_ERR_MSG_FORMAT, limit);
+    }
+
+    public static final String TOO_MANY_CATEGORICAL_FIELD_ERR_MSG_FORMAT = "We can have only %d categorical field/s.";
+    public static String CAN_NOT_FIND_RESULT_INDEX = "Can't find result index ";
+    public static String INVALID_CHAR_IN_RESULT_INDEX_NAME =
+        "Result index name has invalid character. Valid characters are a-z, 0-9, -(hyphen) and _(underscore)";
+    public static String FAIL_TO_VALIDATE = "fail to validate";
```
sudiptoguha, line 34: "Validation failed" or "failed to validate"

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static String INVALID_TIMESTAMP = "Timestamp field: (%s) must be of type date";
+    public static String NON_EXISTENT_TIMESTAMP = "Timestamp field: (%s) is not found in index mapping";
+    public static String INVALID_NAME = "Valid characters for name are a-z, A-Z, 0-9, -(hyphen), _(underscore) and .(period)";
+    // change this error message to make it compatible with old version's integration(nexus) test
+    public static String FAIL_TO_FIND_CONFIG_MSG = "Can't find config with id: ";
```
sudiptoguha, line 39: config file? config entry? What is the object type?

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String CAN_NOT_CHANGE_CATEGORY_FIELD = "Can't change category field";
+    public static final String CAN_NOT_CHANGE_CUSTOM_RESULT_INDEX = "Can't change custom result index";
+    public static final String CATEGORICAL_FIELD_TYPE_ERR_MSG = "A categorical field must be of type keyword or ip.";
+    // Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction
+    public static final String FEATURE_INVALID_MSG_PREFIX = "Feature has an invalid query";
+    public static final String FEATURE_WITH_EMPTY_DATA_MSG = FEATURE_INVALID_MSG_PREFIX + " returning empty aggregated data: ";
+    public static final String FEATURE_WITH_INVALID_QUERY_MSG = FEATURE_INVALID_MSG_PREFIX + " causing a runtime exception: ";
```
sudiptoguha, line 46: "caused a runtime" ... but it can be "causing. .." if there are ongoing retries.

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String UNKNOWN_SEARCH_QUERY_EXCEPTION_MSG =
+        "Feature has an unknown exception caught while executing the feature query: ";
+
+    // ======================================
+    // Index message
+    // ======================================
+    public static final String CREATE_INDEX_NOT_ACKNOWLEDGED = "Create index %S not acknowledged";
```
sudiptoguha, line 53: acknowledged by which entity?

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String SUCCESS_SAVING_RESULT_MSG = "Result saved successfully.";
+    public static final String CANNOT_SAVE_RESULT_ERR_MSG = "Cannot save results due to write block.";
+
+    // ======================================
+    // Resource constraints
+    // ======================================
+    public static final String MEMORY_CIRCUIT_BROKEN_ERR_MSG = "AD memory circuit is broken.";
```
sudiptoguha, line 60: "Memory consumption is too high."? But I see the other error below -- what else is involved in a circuit breaker which can be shared with an user?

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+
+    // ======================================
+    // Transport
+    // ======================================
+    public static final String INVALID_TIMESTAMP_ERR_MSG = "timestamp is invalid";
+
+    // ======================================
+    // transport/restful client
+    // ======================================
+    public static final String WAIT_ERR_MSG = "Exception in waiting for result";
```
sudiptoguha, line 70: More details may help an user ...

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String ALL_FEATURES_DISABLED_ERR_MSG =
+        "Having trouble querying data because all of your features have been disabled.";
```
sudiptoguha, line 72: shorten to "because all features have been disabled"

src/main/java/org/opensearch/ad/AnomalyDetectorProfileRunner.java
kaituo: I have already did similar things. But when I try to bring the change to current PR, I will need other dependency changes. To limit the scope of the PR, I used a temporary workaround like the above.

src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
kaituo: yes, changed

src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
kaituo: changed

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: This message is partial. One use case:

```
if (interval < 0) {
            throw new IllegalArgumentException(
                String.format(Locale.ROOT, "Interval %s %s", interval, CommonMessages.NEGATIVE_TIME_CONFIGURATION)
            );
        }
```

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: changed to "failed to validate"

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: I tried to call a detector or forecaster configuration config. It is a document inside the config index and used to access specific detector/forecaster configuration.

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: Frontend code used this message in its test: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/710e3c4ddadf3689b7f014b5df88f2a82b8db999/public/pages/ReviewAndCreate/containers/__tests__/ReviewAndCreate.test.tsx#L195-L198

Also, there is a comment above "// Modifying message for FEATURE below may break the parseADValidationException method of ValidateAnomalyDetectorTransportAction". Can I keep it as it is for now?

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: by OpenSearch core. Changed to 

```
public static final String CREATE_INDEX_NOT_ACKNOWLEDGED = "Create index %S not acknowledged by OpenSearch core";
```

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: just as you said memory circuit broken means memory consumption is too high.  I changed to

```
public static final String MEMORY_CIRCUIT_BROKEN_ERR_MSG = "AD memory circuit is open due to excessive memory consumption.";
```

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: Exception will be printed after this message like 

```
catch (InterruptedException e1) {
            LOG.error(CommonMessages.WAIT_ERR_MSG);
            throw new IllegalStateException(e1);
        }
```

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: the msg has been hard coded in the frontend: https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/710e3c4ddadf3689b7f014b5df88f2a82b8db999/public/pages/DetectorDetail/utils/constants.ts#L57. Changing it will break frontend. Can we leave it as it is?

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
sudiptoguha: Yes - there is no need to break other artefacts!

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
sudiptoguha: This error is very close to the error in line 82. Is it the case that this error is for continually running jobs and line 82 is for the validator? Or the other way round? Then the two similar errors make sense.

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
sudiptoguha: thanks

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
sudiptoguha: ! yes, let's not break other artifacts.

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: The memory circuit breaker error means the total memory usage exceeded limit, while the error in line 82 means AD memory usage exceeded limit.

sudiptoguha: "The memory circuit breaker error means the total memory usage exceeded limit" --> why not just say that in the message? (It's fine if the error token has the same name as now.) If the user has larger machines/cluster then the total memory usage would be less of a problem and may solve their problem if this error is triggered?

src/main/java/org/opensearch/forecast/constant/ForecastCommonMessages.java
```diff
@@ -0,0 +1,62 @@
...
+    // ======================================
+    // Security
+    // ======================================
+    public static String NO_PERMISSION_TO_ACCESS_FORECASTER = "User does not have permissions to access forecaster: ";
+    public static String FAIL_TO_GET_USER_INFO = "Unable to get user information from forecaster ";
+
+    // ======================================
+    // Used for custom forecast result index
+    // ======================================
+    public static String CAN_NOT_FIND_RESULT_INDEX = "Can't find result index ";
```
jackiehanyang, line 48: is this the full message? or it should be something like `Can't find result index %s`

src/main/java/org/opensearch/timeseries/constant/CommonMessages.java
kaituo: Ok. Changed to

```
public static final String MEMORY_CIRCUIT_BROKEN_ERR_MSG = "The total OpenSearch memory usage exceeds our threshold, opening the AD memory circuit. ";
```

sudiptoguha (dismissed):

sudiptoguha (approved):

jackiehanyang (approved):

PR #919, opened 2023-06-02
[Extensions] Improves performance by replacing cluster state calls for index existence with exists()/aliasExists() API calls

joshpalis:
### Description
Companion SDK PR : https://github.com/opensearch-project/opensearch-sdk-java/pull/802 (SDK PR must be merged first)

Replaces the following cluster state calls to check if an index/alias exists with the `SDKIndicesClient` `exist()` and `existsAlias()` methods :
- `state().getRoutingTable().hasIndex(indexName)`
- `state().metadata().hasIndex(indexName)`
- `state().metadata().indices().containsKey(indexName)`
- `state().metadata().hasAlias(alias)` 

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/674

### DIFF
--- a/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
+++ b/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
@@ -67,6 +67,7 @@ import org.opensearch.ad.settings.AnomalyDetectorSettings;
 import org.opensearch.ad.util.DiscoveryNodeFilterer;
 import org.opensearch.client.indices.CreateIndexRequest;
 import org.opensearch.client.indices.CreateIndexResponse;
+import org.opensearch.client.indices.GetIndexRequest;
 import org.opensearch.client.indices.PutMappingRequest;
 import org.opensearch.client.indices.rollover.RolloverRequest;
 import org.opensearch.client.opensearch.OpenSearchAsyncClient;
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -304,13 +305,57 @@ public static String getCheckpointMappings() throws IOException {
...
+        GetIndexRequest getindexRequest = new GetIndexRequest(indexName);
+
+        CompletableFuture<Boolean> existsFuture = new CompletableFuture<>();
+        sdkRestClient.indices().exists(getindexRequest, ActionListener.wrap(response -> { existsFuture.complete(response); }, exception -> {
+            existsFuture.completeExceptionally(exception);
+        }));
+
+        Boolean existsResponse = existsFuture
+            .orTimeout(AnomalyDetectorSettings.REQUEST_TIMEOUT.get(environmentSettings).getMillis(), TimeUnit.MILLISECONDS)
+            .join();
```
dbwiddis, line 324: Do we need any exception handling here, even if it's just catching and rethrowing with some more information?  OK if the answer is no, just wondering what the user experience will be.

(Repeat comment for most of these calls.)

src/main/java/org/opensearch/ad/util/IndexUtils.java
```diff
@@ -113,9 +116,9 @@ public IndexUtils(
      * @throws IllegalArgumentException Thrown when an alias is passed in that points to more than one index
      */
     public String getIndexHealthStatus(String indexOrAliasName) throws IllegalArgumentException {
-        if (!clusterService.state().getRoutingTable().hasIndex(indexOrAliasName)) {
+        if (!indexExists(indexOrAliasName)) {
             // Check if the index is actually an alias
-            if (clusterService.state().metadata().hasAlias(indexOrAliasName)) {
+            if (aliasExists(indexOrAliasName)) {
```
dbwiddis, line 121: These two calls *might* be able to be combined; instead of checking existence, you could "get" the index and then something in the index metadata indicates whether it's an alias.... I think.   Not for this PR but maybe consider that a TODO.

dbwiddis (approved):
LGTM with questions.

joshpalis:
Checks are failing since SDK main has failed to publish to sonatype. See the error [here](https://github.com/opensearch-project/opensearch-sdk-java/actions/runs/5159989449/jobs/9295652323)

owaiskazi19 (approved):
Except @dbwiddis comments. Rest LGTM

src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java
joshpalis: I dont think its necessary as the only way this request fails is by timing out, since even if the index/alias does not exist, it would follow the happy path (for lack of a better word). There's no other information to provide.

PR #924, opened 2023-06-07
[Extensions] Improves performance by replacing cluster state calls to retrieve cluster name

joshpalis:
### Description
SDK Companion PR : https://github.com/opensearch-project/opensearch-sdk-java/pull/818 (SDK PR must be merged first)

Retrieves cluster name from the SDKClusterService rather than from the cluster state
See comment [here](https://github.com/opensearch-project/opensearch-sdk-java/issues/674#issuecomment-1581621293) for more details

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/674

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/ADCancelTaskTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/ADCancelTaskTransportAction.java
@@ -25,6 +25,9 @@ import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.TransportAction;
 import org.opensearch.ad.task.ADTaskCancellationState;
 import org.opensearch.ad.task.ADTaskManager;
+import org.opensearch.cluster.ClusterName;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.sdk.ExtensionsRunner;
 import org.opensearch.sdk.SDKClusterService;
 import org.opensearch.tasks.Task;
 import org.opensearch.tasks.TaskManager;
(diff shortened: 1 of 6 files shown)


dbwiddis (approved):

src/main/java/org/opensearch/ad/transport/ADCancelTaskTransportAction.java
```diff
@@ -35,25 +37,32 @@ public class ADCancelTaskTransportAction extends TransportAction<ADCancelTaskReq
...
     }
 
     protected ADCancelTaskResponse newResponse(
         ADCancelTaskRequest request,
         List<ADCancelTaskNodeResponse> responses,
         List<FailedNodeException> failures
     ) {
-        return new ADCancelTaskResponse(clusterService.state().getClusterName(), responses, failures);
+        return new ADCancelTaskResponse(
+            new ClusterName(extensionsRunner.getEnvironmentSettings().get("cluster.name")),
```
owaiskazi19, line 62: Instead of creating a new object every time. Can we have it in createComponent and pass along?

joshpalis: I've explored this option but `createComponents` is invoked during `ExtensionsRunner` initialization, therefore we would be unable to retrieve the cluster name from the environment settings until after extension initialization since environment settings are set at that point.

owaiskazi19: Can you check if we can store the cluster name as a global variable somewhere, the reason being cluster name would remain same?

joshpalis: Sure, I think the best place to store this would be in the `SDKClusterService` itself. I'll add a cluster name variable to the `SDKClusterService` and set this during extension initialization as part of `updateSDKClusterService()`, which currently only updates the cluster settings. This way, we can just pull this information from the cluster service and avoid having to pass the extension runner to these transport actions.

owaiskazi19 (approved):
Looks much cleaner now

PR #925, opened 2023-06-12
Refactor: Reorganize code for Forecasting and AnomalyDetector packages

kaituo:
### Description
This PR includes several refactorings to increase code reuse and improve readability:

1. Entity, FeatureData, and RestHandlerUtils have been moved from the 'ad' package to the 'timeseries' package, enabling their use within the Forecasting code.
2. A new class, DataByFeatureId, has been created to hold shared code previously located in FeatureData. FeatureData now inherits from DataByFeatureId. The standalone usage of DataByFeatureId will be detailed in subsequent PRs.
3. Renamed checkAnomalyDetectorFeaturesSyntax to checkFeaturesSyntax in RestHandlerUtils to make it more generic and usable for Forecasting.
4. Constants AGG_NAME_MAX_TIME, AGG_NAME_MIN_TIME, DATE_HISTOGRAM, and FEATURE_AGGS have been moved from ADCommonName to CommonName to make them accessible for Forecasting.
5. A new method, parseConfig, has been added to the Config class. This method can parse configuration for either AnomalyDetector or Forecaster based on input.

Testing:
- The changes have been validated with a successful gradle build.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/866


src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -559,4 +562,14 @@ protected void checkAndThrowValidationErrors(ValidationAspect validationAspect)
             throw new TimeSeriesException(CommonMessages.FAIL_TO_VALIDATE);
         }
     }
+
+    public static Config parseConfig(Class<? extends Config> configClass, XContentParser parser) throws IOException {
```
amitgalitz, line 566: I see this hasn't been used yet, I assume you'll use it later on?

kaituo: yes

src/main/java/org/opensearch/timeseries/util/RestHandlerUtils.java
```diff
@@ -130,18 +135,18 @@ public static XContentParser createXContentParserFromRegistry(NamedXContentRegis
...
         if (features != null) {
-            if (features.size() > maxAnomalyFeatures) {
-                return "Can't create more than " + maxAnomalyFeatures + " anomaly features";
+            if (features.size() > maxFeatures) {
+                return "Can't create more than " + maxFeatures + " features";
             }
-            return validateFeaturesConfig(anomalyDetector.getFeatureAttributes());
+            return validateFeaturesConfig(config.getFeatureAttributes());
         }
         return null;
```
amitgalitz, line 151: Also in the method `validateFeaturesConfig` in this file we have this line: `errorMsgBuilder.append("Detector has duplicate feature names: ");` We should probably change to either "There are duplicate feature names"/"Given configurations have duplicate feature names" or exact the class being used (Forecaster/detector) for the given messaging. In short term wont really be an issue since max feature number will be one for forecasting

src/main/java/org/opensearch/timeseries/util/RestHandlerUtils.java
```diff
@@ -130,18 +135,18 @@ public static XContentParser createXContentParserFromRegistry(NamedXContentRegis
...
     }
```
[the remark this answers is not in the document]
kaituo: Yes. Changed comment on line 200 like

```
/**
     * Wrap action listener to avoid return verbose error message and wrong 500 error to user.
     * Suggestion for exception handling in timeseries analysis (e.g., AD and Forecast):
     * 1. If the error is caused by wrong input, throw IllegalArgumentException exception.
     * 2. For other errors, please use TimeSeriesException or its subclass, or use
     *    OpenSearchStatusException.
     *
     * TODO: tune this function for wrapped exception, return root exception error message
     *
     * @param actionListener action listener
     * @param generalErrorMessage general error message
     * @param <T> action listener response type
     * @return wrapped action listener
     */
    public static <T> ActionListener wrapRestActionListener(ActionListener<T> actionListener, String generalErrorMessage) {
        return ActionListener.<T>wrap(r -> { actionListener.onResponse(r); }, e -> {
            logger.error("Wrap exception before sending back to user", e);
            Throwable cause = Throwables.getRootCause(e);
            if (isProperExceptionToReturn(e)) {
                actionListener.onFailure(e);
            } else if (isProperExceptionToReturn(cause)) {
                actionListener.onFailure((Exception) cause);
            } else {
                RestStatus status = isBadRequest(e) ? BAD_REQUEST : INTERNAL_SERVER_ERROR;
                String errorMessage = generalErrorMessage;
                if (isBadRequest(e) || e instanceof TimeSeriesException) {
                    errorMessage = e.getMessage();
                } else if (cause != null && (isBadRequest(cause) || cause instanceof TimeSeriesException)) {
                    errorMessage = cause.getMessage();
                }
                actionListener.onFailure(new OpenSearchStatusException(errorMessage, status));
            }
        });
    }
```

src/main/java/org/opensearch/timeseries/model/Entity.java
```diff
@@ -9,7 +9,7 @@
  * GitHub history for details.
  */
 
-package org.opensearch.ad.model;
+package org.opensearch.timeseries.model;
```
amitgalitz, line 12: nit: we use detectorId in some method input name which should change now to configID or timeSeriesID. Also some mentions of detector in comments (these comments though also refer to bwc between versions that are only really relatable to AD and are helpful as they are though so I don't mind to either keep as is (e.g. getModelId) and maybe just mentioning nuance when applied to new forecaster. Can also be dealt with when we get to the point forecasting code is calling these methods.

kaituo: I fixed comments and parameters of getModelId. I also modified getModelId to not consider bwc when this is used in forecasting.

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -559,4 +562,14 @@ protected void checkAndThrowValidationErrors(ValidationAspect validationAspect)
...
+        if (configClass == AnomalyDetector.class) {
+            return AnomalyDetector.parse(parser);
+        } else if (configClass == Forecaster.class) {
+            return Forecaster.parse(parser);
+        } else {
+            throw new IllegalArgumentException("Unsupported config type");
```
owaiskazi19, line 572:
```suggestion
            throw new IllegalArgumentException("Unsupported config type. Supported config types are [AnomalyDetector, Forecaster]");
```

owaiskazi19 (commented):
Minor comments. Overall LGTM!

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -559,4 +562,14 @@ protected void checkAndThrowValidationErrors(ValidationAspect validationAspect)
             throw new TimeSeriesException(CommonMessages.FAIL_TO_VALIDATE);
         }
     }
+
+    public static Config parseConfig(Class<? extends Config> configClass, XContentParser parser) throws IOException {
+        if (configClass == AnomalyDetector.class) {
```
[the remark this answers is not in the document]
kaituo: In order to apply the 'instanceof' operator, we require an existing instance of the object. However, in this case, the caller doesn't possess an instance yet. The primary objective of this method is, in fact, to generate such an instance.

src/main/java/org/opensearch/timeseries/model/Config.java
kaituo: Changed.

src/main/java/org/opensearch/timeseries/model/Config.java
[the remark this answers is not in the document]
owaiskazi19: Can you use `equals` here instead of `==`?

owaiskazi19:
Should we wait for nightly build for CI to pass?

src/main/java/org/opensearch/timeseries/model/Config.java
owaiskazi19: Will you be pushing all the changes together?

kaituo: yes

kaituo:
> Should we wait for updated nightly build for CI to pass?

We don't need to. The CI has been failing due to that for more than 1 week now.

src/main/java/org/opensearch/timeseries/model/Config.java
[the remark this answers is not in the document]
kaituo: In Java, == is used for reference equality, meaning it checks whether two references point to the exact same object. When comparing classes, you want to know if configClass is exactly AnomalyDetector.class or Forecaster.class, not if it is an instance that is equal to those classes in some way.

On the other hand, equals() is used for logical equality and must be implemented by the class. It checks whether two objects should be considered "equal" based on their internal state, not their identity. However, equals() isn't meaningful when comparing Class objects, because there isn't a logical way to say two distinct classes are "equal."

So in summary, your current code is correct and there's no need to replace == with equals().

kaituo:
Bwc test failed:

    cluster{::adBwcCluster0} failed to wait for cluster health yellow after 40 SECONDS
    503 Service Unavailable

latest development: https://github.com/opensearch-project/OpenSearch/issues/5091

owaiskazi19 (approved):

PR #942, opened 2023-07-06
[Extensions] Renames AD extension routes to NamedRoutes

DarshitChanpura:
### Description
As part of an ongoing effort to introduce authorization in REST layer, a concept of NamedRoute was introduced in core. A NamedRoute provides cluster admins an ability to grant/revoke access to specific endpoints in a cluster. This is introduced as an optional feature for plugins, and a mandatory feature for extensions since they don't interact with the cluster at transport layer and so the only way to authorize these requests would be at REST layer. To enforce all routes to be NamedRoute, a change was made in SDK: https://github.com/opensearch-project/opensearch-sdk-java/pull/827 to enforce declaring new routes as NamedRoute.

This PR introduces changes that conform to the changes introduced in SDK.

### Example
Consider this url for example: `POST /detectors/{detectorId}/profile`
This route will now be registered with a route named: `ad:detectors/profile` considering `routeNamePrefix` value in settings.yml is set to `ad`.
Consider this role:

```yml
ad_role:
  reserved: true
  cluster_permissions:
    - 'ad:detectors/profile'
```

And says user X is mapped to this role. Now there are two possible use cases:

1. Without Security: No change.
2. With Security:
      a. Once https://github.com/opensearch-project/security/pull/2956 is merged and released for 2.9, cluster admins can now permission these routes as if they were granting cluster_permission to a transport action. This will be evaluated at REST layer inside security plugin. For the given example, user X will be evaluated against the route `/detectors/{detectorId}/profile` which is mapped to name `ad:detectors/profile`, and since user X is mapped to a role `ad_role` which contains this permission, the user will be successfully evaluated and the request chain will proceed. On other hand, if user was not mapped to this role or if there was a typo mapping the role, the evaluation would fail with 401.

    b. If https://github.com/opensearch-project/security/pull/2956 does not get released in 2.9 due to internal reviews not being complete on time, then there will not be any change in the way authorization works. It will be business as usual.

### Issues Resolved
- Related to https://github.com/opensearch-project/security/issues/2590


DarshitChanpura:
CI will be unblocked once https://github.com/opensearch-project/opensearch-sdk-java/pull/827 is merged and an artifact is available.

src/main/java/org/opensearch/ad/rest/RestPreviewAnomalyDetectorAction.java
```diff
@@ -11,8 +11,9 @@
 
 package org.opensearch.ad.rest;
 
-import static org.opensearch.ad.util.RestHandlerUtils.PREVIEW;
+import static org.opensearch.ad.util.RestHandlerUtils.*;
```
dbwiddis, line 14: Don't use star imports

src/main/java/org/opensearch/ad/rest/RestDeleteAnomalyResultsAction.java
```diff
@@ -112,8 +115,16 @@ protected ExtensionRestResponse prepareRequest(RestRequest request) throws IOExc
...
-    public List<RouteHandler> routeHandlers() {
+    public List<NamedRoute> routes() {
+        String path = AnomalyDetectorExtension.AD_BASE_DETECTORS_URI + "/results";
         return ImmutableList
-            .of(new RouteHandler(RestRequest.Method.DELETE, AnomalyDetectorExtension.AD_BASE_DETECTORS_URI + "/results", handleRequest));
+            .of(
+                new NamedRoute.Builder()
+                    .method(DELETE)
+                    .path(path)
+                    .uniqueName(routePrefix("detector/results"))
```
dbwiddis, line 125: This (and most of the others) are "detectors".

src/main/java/org/opensearch/ad/rest/AbstractSearchAction.java
```diff
@@ -44,10 +46,7 @@
 import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.extensions.rest.ExtensionRestResponse;
-import org.opensearch.rest.BytesRestResponse;
-import org.opensearch.rest.RestChannel;
-import org.opensearch.rest.RestRequest;
-import org.opensearch.rest.RestStatus;
+import org.opensearch.rest.*;
```
dbwiddis, line 49: No *

dbwiddis (changes requested):
LGTM except we don't use star imports (if spotless isn't failing we should make it do so!) and the base URI is "detectors" so the named route should match.

src/main/java/org/opensearch/ad/rest/RestPreviewAnomalyDetectorAction.java
DarshitChanpura: the pain of intellij and its extra smartness :D. I will fix this.

DarshitChanpura:
@dbwiddis The terms "detector/" and "detectors/" seem to be used interchangeably. I've updated it to reflect uri for base detectors, but it could be standardized if needed.

DarshitChanpura:
CI will pass once https://github.com/opensearch-project/opensearch-sdk-java/pull/868 is merged and an artifact is available.

owaiskazi19:
@DarshitChanpura can you provide an example of the new URl for ad-extension for let's say create detector?

src/main/java/org/opensearch/ad/rest/RestDeleteAnomalyDetectorAction.java
```diff
@@ -105,7 +106,7 @@ public List<ReplacedRouteHandler> replacedRouteHandlers() {
             );
     }
 
-    private Function<RestRequest, ExtensionRestResponse> handleRequest = (request) -> {
+    private Function<RestRequest, RestResponse> handleRequest = (request) -> {
```
owaiskazi19, line 109: Trying to understand why are we changing the response type from `ExtensionRestResponse` to `RestResponse`?
If this is coming from NameRoute then we should change the return type of `prepareRequest` to `RestResponse` as well?

DarshitChanpura: NamedRoute is generally available to both plugins and extensions, and hence RestResponse instead of ExtensionRestResponse (which inherits RestReponse transitively). Which instance of `prepareRequest` are you referring to?

owaiskazi19: On line 71

DarshitChanpura:
> @DarshitChanpura can you provide an example of the new URl for ad-extension for let's say create detector? Let's add it in the PR description.

Update the description

src/main/java/org/opensearch/ad/rest/RestDeleteAnomalyDetectorAction.java
DarshitChanpura: prepareRequest should still return ExtensionRestResponse as it is one of the concrete implementations of RestResponse and will be interpreted as such when send the response back to user. However, the only use case I see for ExtensionRestResponse is for communication between OS and extension to mark that the content was consumed and a list of consumed content. 

@dbwiddis had some thoughts on whether we should keep ExtensionRestResponse or just use BytesRestReponse class

owaiskazi19 (approved):
LGTM! Thanks for addressing comments

DarshitChanpura:
Closing this until the work is resumed.

Issue #936, opened 2023-06-29
Compatibility with segment replication


dreamer-89:
### Summary
With 2.9.0 release, there are lot of enhancements going in for segment replication[1][2] feature (went GA in 2.7.0), we need to ensure different plugins are compatible with current state of this feature. Previously, we ran [tests](https://github.com/opensearch-project/OpenSearch/issues/6602) on plugin repos to verify this compatibility but want plugin owners to be aware of these changes so that required updates (if any) can be made. With `2.10.0` release, [remote store feature](https://github.com/opensearch-project/OpenSearch/issues/2992) is going GA which internally uses SEGMENT replication strategy only i.e. it enforces all indices to use `SEGMENT` replication strategy. So, it is important to validate plugins are compatible with segment replication feature. 

#### 1. Refresh policy behavior
1. RefreshPolicy.IMMEDIATE will only refresh primary shards but not replica shards immediately. Instead post refresh, primary will start  a round of segment replication to update the replica shard copies leading to eventual consistency.
2. RefreshPolicy.WAIT_UNTIL ensures the indexing operation is searchable in your cluster i.e. RAW (Read after write guarantee). With segment replication, this guarantee is not promised due to delay in replica shared updates from asynchronous background refreshes.

####  2. Refresh lag on replicas
With segment replication, there is inherent delay in documents to be searchable on replica shard copies. This is due to the fact that replica shard copies over data (segment) files from primary. Thus, compared to document replication, there will be on average increase in amount of time the replica shards are consistent with primaries.

#### 3. System/hidden indices support
With https://github.com/opensearch-project/OpenSearch/pull/8200, system and hidden indices are now supported with `SEGMENT` replication strategy. We need to ensure there are no bottlenecks which prevents system/hidden indices with segment replication.

### Next steps
With segment replication strong reads are not guaranteed. Thus, if the plugin needs strong reads guarantees specially as alternative to change in behavior of refresh policy and lag on replicas (point 1 and 2 above), we need to update search requests to target primary shard only. With https://github.com/opensearch-project/OpenSearch/pull/7375, core now supports primary shards only based search. Please follow [documentation](https://opensearch.org/docs/latest/api-reference/search/#the-preference-query-parameter) for examples and details

### Open questions
In case of any questions or issues, please post it in [core](https://github.com/opensearch-project/OpenSearch/issues/8211) issue

### Reference
[1] [Design](https://github.com/opensearch-project/OpenSearch/issues/2229)

[...]

PR #948, opened 2023-07-11
Enforce DOCUMENT Replication for AD Indices and Adjust Primary Shards

kaituo:
### Description
In this PR, we temporarily enforce DOCUMENT replication for AD indices. This change is necessary due to the current limitation of SegRep, which doesn't support Get/MultiGet by ID. This measure will be in place until SegRep adds support for these operations.

This adjustment aligns with the modification made in the referenced PR: https://github.com/opensearch-project/job-scheduler/pull/417

Additionally, this PR increases the number of primary shards for forecasting indices from 10 to 20, recognizing the higher write intensity of these indices. For instance, when the horizon is 24, we are expected to save 25 documents: 24 for each forecast and 1 for the actual value. In contrast, for AD, we save one document per actual value.

Testing done:
1. gradle build passed

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/936

### DIFF
--- a/src/main/java/org/opensearch/forecast/indices/ForecastIndexManagement.java
+++ b/src/main/java/org/opensearch/forecast/indices/ForecastIndexManagement.java
@@ -204,8 +204,8 @@ public class ForecastIndexManagement extends IndexManagement<ForecastIndex> {
         } catch (IOException e) {
             throw new EndRunException("", "Cannot find checkpoint mapping file", true);
         }
-        // AD indices need RAW (e.g., we want users to be able to consume AD results as soon as possible and send out an alert if anomalies
-        // found).
+        // forecast indices need RAW (e.g., we want users to be able to consume forecast results as soon as
+        // possible and send out an alert if a threshold is breached).
         Settings replicationSettings = Settings.builder().put(SETTING_REPLICATION_TYPE, DOCUMENT.name()).build();
         CreateIndexRequest request = new CreateIndexRequest(ForecastCommonName.FORECAST_CHECKPOINT_INDEX_NAME, replicationSettings)
             .mapping(mapping, XContentType.JSON);
--- a/src/main/java/org/opensearch/forecast/indices/ForecastIndexManagement.java
+++ b/src/main/java/org/opensearch/forecast/indices/ForecastIndexManagement.java
@@ -11,6 +11,7 @@
 
 package org.opensearch.forecast.indices;
 
+import static org.opensearch.cluster.metadata.IndexMetadata.SETTING_REPLICATION_TYPE;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/timeseries/indices/IndexManagement.java
```diff
@@ -477,7 +482,11 @@ public static String getJobMappings() throws IOException {
      */
     public void initJobIndex(ActionListener<CreateIndexResponse> actionListener) {
         try {
-            CreateIndexRequest request = new CreateIndexRequest(CommonName.JOB_INDEX).mapping(getJobMappings(), XContentType.JSON);
+            // AD indices need RAW (e.g., we want users to be able to consume AD results as soon as possible and send out an alert if
+            // anomalies found).
+            Settings replicationSettings = Settings.builder().put(SETTING_REPLICATION_TYPE, DOCUMENT.name()).build();
```
sudiptoguha, line 487: This looks good and is a positive change. Would this cause existing detectors to have any issues?

sudiptoguha (dismissed):

src/main/java/org/opensearch/timeseries/indices/IndexManagement.java
kaituo: It should not. Segment replication should not be enabled for existing indices.

dbwiddis (dismissed):
Changes LGTM.  Consider adding a link to the SegRep issue tracking Get by ID so that future-me and future-you know when we can/should undo these temporary fixes.

amitgalitz (approved):
Thanks for making this change!

kaituo:
> Changes LGTM. Consider adding a link to the SegRep issue tracking Get by ID so that future-me and future-you know when we can/should undo these temporary fixes.

Both of us have a link in the get by ID issue: https://github.com/opensearch-project/OpenSearch/issues/8536 
Is that ok?

src/main/java/org/opensearch/forecast/indices/ForecastIndexManagement.java
```diff
@@ -201,7 +204,10 @@ public void initCheckpointIndex(ActionListener<CreateIndexResponse> actionListen
         } catch (IOException e) {
             throw new EndRunException("", "Cannot find checkpoint mapping file", true);
         }
-        CreateIndexRequest request = new CreateIndexRequest(ForecastCommonName.FORECAST_CHECKPOINT_INDEX_NAME)
+        // AD indices need RAW (e.g., we want users to be able to consume AD results as soon as possible and send out an alert if anomalies
```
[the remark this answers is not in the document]
kaituo: good catch. Fixed.

amitgalitz (approved):

sudiptoguha (approved):

PR #971, opened 2023-08-08
Fix build with latest OpenSearch

smortex:
OpenSearch recently merged some code that cause buid failure for this
module, as spotted in #585.

This change the import to match the new one after
https://github.com/opensearch-project/OpenSearch/pull/9103

PR #972, opened 2023-08-09
Fix build with latest OpenSearch (for real this time)

smortex:
This is a follow-up to #971, where commit 2839bc2c6f470639b68901086c80cb2152d96558 only fixed _some_ but not _all_ of the build errors (it allowed `./gradlew run` to pass but `./gradlew build` was failing).

Fix the build errors when we run `./gradlew build`.


smortex:
Looks better (than #971).

As far as I understand, the BWC tests depends on the latest build succeeding but they currently fail.  As a consequence, the tests prerequisites are not fulfill and the test are not run.

kaituo (approved):

sudiptoguha (approved):

Issue #978, opened 2023-08-14
[FEATURE]bump rcf version


kaituo:
**Is your feature request related to a problem?**
added support for automatic Anomaly Detection (AD), the 'Alert Once' option introduced in RCF 3.8, and forecasting.

**What solution would you like?**
bump rcf

PR #979, opened 2023-08-14
Update RCF to v3.8 and Enable Auto AD with 'Alert Once' Option

kaituo:
### Description
This PR added support for automatic Anomaly Detection (AD) and the 'Alert Once' option introduced in RCF 3.8.

Testing done:

1. Deserialization Test:
* Verified model deserialization from 3.0-rc3.
* Ensured consistent scoring using the rc3 checkpoint and rc3 dependency on identical data.

2. Backward Compatibility Test:
* Executed a mixed cluster with versions 2.10 and 3.0.
* Validated that older detectors still produce results without throwing any exceptions in a blue-green simulation scenario.

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/978


src/main/java/org/opensearch/ad/ml/ModelManager.java
```diff
@@ -532,6 +533,10 @@ private void trainModelForStep(
             .boundingBoxCacheFraction(TimeSeriesSettings.REAL_TIME_BOUNDING_BOX_CACHE_RATIO)
             .shingleSize(detector.getShingleSize())
             .anomalyRate(1 - thresholdMinPvalue)
+            .transformMethod(TransformMethod.NORMALIZE)
+            .alertOnce(true)
+            .autoAdjust(true)
```
sudiptoguha, line 538: AlertOnce and AutoAdjust can interfere. It's ok to have both true -- will have fewer anomalies. If any of these are changed to false then we should consider  changing the other (in particular, if autoAdjust is false then do we want alertOnce -- unclear, may be may be not)

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
```diff
@@ -256,6 +257,10 @@ private ThresholdedRandomCutForest createTRCF() {
             .precision(Precision.FLOAT_32)
             .randomSeed(seed)
             .boundingBoxCacheFraction(0)
+            .transformMethod(TransformMethod.NORMALIZE)
+            .alertOnce(true)
+            .autoAdjust(true)
+            .internalShinglingEnabled(false)
```
sudiptoguha, line 263: Is this for existing models? Single Streams? Creating external shingled models will take up more memory (and more queries)

sudiptoguha (approved):
Mostly ok - some questions which may need clarifications.

src/main/java/org/opensearch/ad/ml/ModelManager.java
kaituo: are you saying there are dependencies between these two options? How to figure out whether to set one false if the other one is false?

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
kaituo: we only enabled internal shingling for real-time HC detectors. Other detectors all use external shingling for historical reasons.

sudiptoguha: Makes sense. Thx.

src/main/java/org/opensearch/ad/ml/ModelManager.java
sudiptoguha: Well, "alertOnce" is a directive :) That is, if there is a "run" or continuous anomalies then we start alert once and stop (till some internal checks indicates that at least one normal point was seen). In the autoAdjust case, we would learn and try to suppress anomalies of both the observed value and expected value are in some noise range. So if there is a level shift followed quickly by another spike/dip then autoAdjust has some chance of catching it. But all of these are based on unsupervised estimations -- so it's hard to guarantee behavior. In general starting from (false,false) autoAdjust = true should reduce anomalies; and alertOnce = true would reduce even further. But there may be applications where users prefer a lot of continuous anomalies (for example, to detect a level shift!).

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
```diff
@@ -1098,4 +1100,138 @@ public void testShouldSave() {
...
+        URI uri = ClassLoader.getSystemResource("org/opensearch/ad/ml/rcf_3_0_rc3_single_stream.json").toURI();
+        String filePath = Paths.get(uri).toString();
+        String json = Files.readString(Paths.get(filePath), Charset.defaultCharset());
+        // For the parsing of .toTrcf to work I had to manually change "\u003d" in code back to =.
+        // In the byte array it doesn't seem like this is an issue but whenever reading the byte array response into a file it
+        // converts "=" to "\u003d" https://groups.google.com/g/google-gson/c/JDHUo9DWyyM?pli=1
+        // I also needed to bypass the trcf as it wasn't being read as a key value but instead part of the string
+        Map map = gson.fromJson(json, Map.class);
+        String model = (String) ((Map) ((Map) ((ArrayList) ((Map) map.get("hits")).get("hits")).get(0)).get("_source")).get("modelV2");
+        // model = model.split(":")[1].substring(1);
```
amitgalitz, line 1120: nit: previously i think I added this to deal with "trcf" was this not needed any more in this case?

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
```diff
@@ -1098,4 +1100,138 @@ public void testShouldSave() {
...
+        // In the byte array it doesn't seem like this is an issue but whenever reading the byte array response into a file it
+        // converts "=" to "\u003d" https://groups.google.com/g/google-gson/c/JDHUo9DWyyM?pli=1
+        // I also needed to bypass the trcf as it wasn't being read as a key value but instead part of the string
+        Map map = gson.fromJson(json, Map.class);
+        String model = (String) ((Map) ((Map) ((ArrayList) ((Map) map.get("hits")).get("hits")).get(0)).get("_source")).get("modelV2");
+        model = model.split(":")[1];
+        model = model.substring(1, model.length() - 2);
+        // model = Base64.getEncoder().encodeToString(org.apache.commons.codec.binary.Base64.decodeBase64(model));
+        // Simulate JSON parsing by replacing Unicode escape sequence with the actual character
+        model = unescapeJavaString(model);
```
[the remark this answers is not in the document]
kaituo: Maybe related to how I record those Json response file. Without doing it, we experience model corruption exception.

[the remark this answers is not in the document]
amitgalitz: I said the difference in recording the json response makes sense, I'm not 100% sure if we are talking about the same thing, but I used the lines above to remove the last = charachter in the response

amitgalitz (approved):

src/test/java/org/opensearch/ad/ml/CheckpointDaoTests.java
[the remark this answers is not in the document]
kaituo: yeah, different checkpoint might need different conversion. In my single stream checkpoint, I don't need to do anything.

sudiptoguha (approved):

Issue #989, opened 2023-08-21
Compatibility with segment replication


dreamer-89:
Coming from https://github.com/opensearch-project/anomaly-detection/issues/936#issuecomment-1631313721 and with core now supporting real time reads for segment replication enabled indices (https://github.com/opensearch-project/OpenSearch/issues/8536), we can revert https://github.com/opensearch-project/anomaly-detection/pull/948 which enforces DOCUMENT replication strategy. 

CC @kaituo @ohltyler 

Request owners to label this issue with `2.10.0` release.

kaituo:
will respond later.

jackiehanyang:
revert change pr created - https://github.com/opensearch-project/anomaly-detection/pull/999

@dreamer-89 do you have any instruction on how to test this to ensure the core side change is working well with AD?

dreamer-89:
> revert change pr created - #999
> 
> @dreamer-89 do you have any instruction on how to test this to ensure the core side change is working well with AD?

Thanks @jackiehanyang for working on this change. I commented on revert PR #999

kaituo:
I've conducted a comprehensive review of our backend and would like to share my insights on two key components:

1. State Index Query:

* The real-time performance is satisfactory, especially since it primarily supports the profile API. The RAW feature is not required.
* For historical detectors, a change may or may not be necessary. Interestingly, ml-commons has no reported issues and historical detectors have implementation similar to ml-commons. You can refer to [Issue #1023](https://github.com/opensearch-project/ml-commons/issues/1023) which Yaliang has reviewed and signed off on. Should any modifications be required, TaskManager.getAndExecuteOnLatestTasks should prioritize primary shards to locate the most recent task.

2. Result Index Query from Alerting:

As indicated in [Issue #974](https://github.com/opensearch-project/alerting/issues/974), there's a slight delay of a few seconds. However, given that we operate with a minute-level delay, I believe this won't pose any significant concerns.

In light of the above, I am inclined to approve Pull Request #999 . This should allow us to finalize the seg replication compatibility.

dreamer-89:
Thanks @kaituo for the summary and explanation. Do we need anything else other than merging PR #999 ? If not, do you mind merging #999 and closing this issue.

kaituo:
We do need another approver.

dreamer-89:
@kaituo @jackiehanyang : As PR #999 is merged and backported. Do you mind closing this issue if nothing else is pending.

[...]

PR #999, opened 2023-08-28
Revert "Enforce DOCUMENT Replication for AD Indices and Adjust Primar…

jackiehanyang:
This reverts commit bc1649922ef39619424fc16c64af5836e48a4a12.

### Description
Per request in issue - https://github.com/opensearch-project/anomaly-detection/issues/989, reverting this change as core now supporting real time reads for segment replication enabled indices (https://github.com/opensearch-project/OpenSearch/issues/8536)

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/989

### DIFF
--- a/src/main/java/org/opensearch/ad/indices/ADIndexManagement.java
+++ b/src/main/java/org/opensearch/ad/indices/ADIndexManagement.java
@@ -19,8 +19,6 @@ import static org.opensearch.ad.settings.AnomalyDetectorSettings.AD_RESULT_HISTO
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.ANOMALY_DETECTION_STATE_INDEX_MAPPING_FILE;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.ANOMALY_RESULTS_INDEX_MAPPING_FILE;
 import static org.opensearch.ad.settings.AnomalyDetectorSettings.CHECKPOINT_INDEX_MAPPING_FILE;
-import static org.opensearch.cluster.metadata.IndexMetadata.SETTING_REPLICATION_TYPE;
-import static org.opensearch.indices.replication.common.ReplicationType.DOCUMENT;
 
 import java.io.IOException;
 import java.util.EnumMap;
@@ -66,7 +64,7 @@ public class ADIndexManagement extends IndexManagement<ADIndex> {
      * @param settings       OS cluster setting
      * @param nodeFilter     Used to filter eligible nodes to host AD indices
      * @param maxUpdateRunningTimes max number of retries to update index mapping and setting
-     * @throws IOException when failing to get mapping file
+     * @throws IOException
      */
     public ADIndexManagement(
         Client client,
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/ad/indices/ADIndexManagement.java
```diff
@@ -197,10 +195,7 @@ public void initDefaultResultIndexDirectly(ActionListener<CreateIndexResponse> a
     @Override
     public void initStateIndex(ActionListener<CreateIndexResponse> actionListener) {
         try {
-            // AD indices need RAW (e.g., we want users to be able to consume AD results as soon as possible and send out an alert if
-            // anomalies found).
-            Settings replicationSettings = Settings.builder().put(SETTING_REPLICATION_TYPE, DOCUMENT.name()).build();
-            CreateIndexRequest request = new CreateIndexRequest(ADCommonName.DETECTION_STATE_INDEX, replicationSettings)
+            CreateIndexRequest request = new CreateIndexRequest(ADCommonName.DETECTION_STATE_INDEX)
```
dreamer-89, line 198: @jackiehanyang : Thanks for working on this change.

With https://github.com/opensearch-project/OpenSearch/issues/8536, core now supports realtime reads with [get](https://opensearch.org/docs/latest/api-reference/document-apis/get-documents/)/[mget](https://opensearch.org/docs/latest/api-reference/document-apis/multi-get/) APIs for `SEGMENT` replication. Thus, please ensure search request needing RAW, are using get/mget APIs.

kaituo: @dreamer-89 What do you mean by using get/mget APIs for search request? We do have some search requests needing RAW.

Also, for realtime get/mget, do we need any changes in AD to use the new code in core?

dreamer-89: @kaituo : I meant if your plugin relies on get/mget APIs for data retrieval, then these APIs should automatically return latest data because this is handled on core. We recently added this support for SEGMENT replication enabled indices via https://github.com/opensearch-project/OpenSearch/issues/8536. Thus, as long as you are only using these APIs for RAW, no change is needed. 

>Also, for realtime get/mget, do we need any changes in AD to use the new code in core?

No

Also, please note DOCUMENT replication guarantees RAW on the write path by using `IMMEDIATE` or `WAIT_UNTIL` refresh policy, which ensure all replica shards are updated with same indexing operation. With SEGMENT replication this guarantee does not holds true. Thus, if you are using IMMEDIATE, WAIT_UNTIL refresh policy, then for follow up search requests (where you need RAW), you need to use either get/mget APIs or use _primary preference parameter. The former is preferred as later puts extra load on primary shard.

CC @mch2

kaituo: @dreamer-89 How do I replace search request with get/mget APIs?

kaituo (approved):

owaiskazi19 (approved):

dreamer-89:
@jackiehanyang @kaituo @owaiskazi19 : Can you please help with backport, review and merge

dreamer-89:
@jackiehanyang @kaituo @owaiskazi19 : Ping for backport

PR #1050, opened 2023-09-13
Fix AnomalyDetectorRestApiIT.testUpdateAnomalyDetectorNameToNew

kaituo:
### Description
Previously, the test checked if the new detector and the parsed new detector had different last update times. This was not the intended behavior. The correct check should be between the old and new detector's last update time. This discrepancy was mostly masked by serialization and deserialization differences, but surfaced in JDK 8 on Windows where the timestamps matched.

Changes:
- Updated test to compare old vs. new detector's last update time.

Testing:
- Ran `gradle build`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -602,10 +602,13 @@ List<String> jacocoExclusions = [
         // https://github.com/opensearch-project/anomaly-detection/issues/241
         'org.opensearch.ad.task.ADBatchTaskRunner',
         'org.opensearch.ad.task.ADTaskManager',
-        
+
         //TODO: custom result index caused coverage drop
         'org.opensearch.ad.indices.AnomalyDetectionIndices',
-        'org.opensearch.ad.transport.handler.AnomalyResultBulkIndexHandler'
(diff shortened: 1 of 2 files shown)


owaiskazi19:
@kaituo is this change required because of JDK8? Does that mean we don't want this change on `main` then?

jackiehanyang (approved):

kaituo:
> @kaituo is this change required because of JDK8? Does that mean we don't want this change on `main` then?

The test environment windows + JDK 8 fails the test. But this change is also required on main since the test itself has problems (see git commit description).

owaiskazi19:
> @owaiskazi19 The test environment windows + JDK 8 fails the test. But this change is also required on main since the test itself has problems (see git commit description).

I see! Do we have a PR for the `main` branch as well?

kaituo:
> > @owaiskazi19 The test environment windows + JDK 8 fails the test. But this change is also required on main since the test itself has problems (see git commit description).
> 
> I see! Do we have a PR for the `main` branch as well?

will send one after this PR is merged.

owaiskazi19 (approved):

PR #1110, opened 2023-12-06
Add an AD transport client

ohltyler:
### Description
This PR adds an AD transport client for the following transport actions:
- search anomaly detectors
- search anomaly results

This follows the same approach as the ML commons client - it is constructed using the caller's node client, and uses it to execute various transport actions.

For these particular methods, we don't do any special pre- or post-processing. These actions are mostly simple pass-through actions to query the internal system indices, and take a general SearchRequest and return a general SearchResponse. We do the same with the client methods that are added.

Note that the unit tests are intended to ensure the underlying transport actions of the client methods are being exercised and returning expected results, covering some of the basic use cases (no system index, empty results, populated results). The exhaustive set of transport action scenarios are within the respective transport action test suites themselves. We do not want/need to unnecessarily duplicate all of those tests here, as it would add require more places that would need to be updated, should the action interfaces change.


ohltyler:
Not backporting. I have a manual branch `ad-client` I will open against 2.x when this is merged. They need to be maintained separately because of the package refactoring done on main.

joshpalis (approved):
Seems the checks are failing do to a spotless error, `./gradlew spotlessApply` should resolve these. Otherwise lgtm!

ohltyler:
@joshpalis CIs are passing on Java 17+ which is what the current codebase is compiled and formatted on. We should be fine to merge and potentially update CIs when minimum Java version is defined for 3.x. Regardless, prefer to leave out any logic changing that as part of this PR which is unrelated.

joshpalis (approved):

dbwiddis (approved):
This is pretty cool functionality.

PR #1119, opened 2023-12-21
Add profile transport action to AD client

ohltyler:
### Description
Adds a profile transport action method to the AD client. Note this method only takes in a `detectorId` string vs. the full `ADTaskProfileRequest` as required by the transport action. This is to simplify the interface and requirements from caller's side, as generating the request is not trivial and requires some AD-specific logic, including the constructed `DiscoveryNodeFilterer` used to determine the eligible AD nodes for the transport action to distribute the `ADTaskProfileRequest` to.

Additionally, this updates the `ADTaskProfileResponse` to add a helper method to serialize from a generic `ActionResponse`. This is needed to prevent classloader issues and ClassCastExceptions when this is constructed from external plugins using the AD client. 

Adds relevant tests and some minor cleanup on existing tests.

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
+++ b/src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
@@ -6,11 +6,9 @@
 package org.opensearch.ad.client;
 
 import java.util.function.Function;
-import java.util.function.Predicate;
 
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
-import org.opensearch.ad.constant.ADCommonName;
 import org.opensearch.ad.transport.ADTaskProfileAction;
 import org.opensearch.ad.transport.ADTaskProfileRequest;
 import org.opensearch.ad.transport.ADTaskProfileResponse;
--- a/src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
+++ b/src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
@@ -7,9 +7,9 @@ package org.opensearch.ad.client;
 
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
-import org.opensearch.ad.transport.ProfileAction;
-import org.opensearch.ad.transport.ProfileRequest;
-import org.opensearch.ad.transport.ProfileResponse;
+import org.opensearch.ad.transport.ADTaskProfileAction;
+import org.opensearch.ad.transport.ADTaskProfileRequest;
+import org.opensearch.ad.transport.ADTaskProfileResponse;
 import org.opensearch.ad.transport.SearchAnomalyDetectorAction;
 import org.opensearch.ad.transport.SearchAnomalyResultAction;
 import org.opensearch.client.Client;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
```diff
@@ -32,4 +35,11 @@ public void searchAnomalyResults(SearchRequest searchRequest, ActionListener<Sea
             listener.onResponse(searchResponse);
         }, listener::onFailure));
     }
+
+    @Override
+    public void getDetectorProfile(ADTaskProfileRequest profileRequest, ActionListener<ADTaskProfileResponse> listener) {
+        this.client.execute(ADTaskProfileAction.INSTANCE, profileRequest, ActionListener.wrap(profileResponse -> {
+            listener.onResponse(profileResponse);
```
dbwiddis, line 42: When using MLClient class transport responses in Flow Framework, we experienced Class Cast Exceptions on these responses because the two different plugins load the same class using different classloaders.

I believe that will be the case here as well.  See the MLClient for implementations that use serialization to generate a new copy of the `profileResponse` by the calling thread.

These are not needed for response types that are pulled from OpenSearch itself.

ohltyler:
Explanation of failing CI: same as previous PR - see https://github.com/opensearch-project/anomaly-detection/pull/1110#issuecomment-1850694995

dbwiddis (commented):
LGTM in general but a concern about class loading. Not sure if you've run this calling from the tools repo?

src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
ohltyler: Thanks for the insight! I'll check the ML client. I haven't tested the integration with the skills repo yet. I suppose I didn't run into the issue yet since the 2 transport actions already added are just `SearchRequest`/`SearchResponse` inputs/outputs.

ohltyler: I've updated to follow the same pattern as ML client. Also, changed the input to simply be string; the client will generate the request. This prevents any potential classloader issues on request-side, and simplifies the interface and requirements from caller's side, as generating the request is not trivial and requires some AD-specific logic.

dbwiddis (approved):
LGTM!

dbwiddis (approved):

jackiehanyang:
Have you tested this on a cluster? If so, could you paste the new API request and response example here?

ohltyler:
> Have you tested this on a cluster? If so, could you paste the new API request and response example here?

Basic scenarios covered in the tests confirm the expected outputs. There is no new API request/response, it just takes in a `String` and returns a `ADTaskProfileResponse` defined in the method signature in `AnomalyDetectionClient`. The overall usage is the same as that of the ML client, for example. An external plugin initializes the client with its own node client, and uses the interfaced methods to execute various AD-related transport actions.

There will likely be slight changes made to these methods as they get more extensively tested in the context of AD-related tools in the [skills plugin](https://github.com/opensearch-project/skills) and the integration test framework being built there.

jackiehanyang (approved):

PR #1164, opened 2024-02-21
Inject NamedWriteableRegistry in AD node client

ohltyler:
### Description
Currently, when serializing the `GetAnomalyDetectorResponse` via `fromActionResponse` (triggered by executing via the node client from a separate plugin) it is not passing a named-writeable-aware stream input, which causes serialization to fail for certain sub-fields within `GetAnomalyDetectorResponse` that are trying to read named writeables via `readNamedWriteable()`. Error seen:
```
Caused by: java.lang.UnsupportedOperationException: can't read named writeable from StreamInput
```

This fixes that by changing the constructor of the node client to pass in a `NamedWriteableRegistry`, to finally pass down to `fromActionResponse` to construct a valid `NamedWriteableAwareStreamInput` to successfully serialize and return a `GetAnomalyDetectorResponse`.

Note `fromActionResponse()` is only used in the node client so the blast radius is small. Also note this serialization is only needed for this response type since the other response types currently in the node client are generic `SearchResponse`s which don't require the ser/deser that AD-implemented objs do to get around the `ClassLoaderException`s caused by plugin-to-plugin communication.

Testing:
- added unit test to cover `fromActionResponse` with a mock registry
- tested end-to-end with a patched version of the skills repo (see PR here: https://github.com/opensearch-project/skills/pull/235) that passes down that plugin's `NamedWriteableRegistry` to the client. Was able to reproduce the issue, and confirmed it works as expected after this fix.

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
+++ b/src/main/java/org/opensearch/ad/client/AnomalyDetectionNodeClient.java
@@ -17,12 +17,15 @@ import org.opensearch.ad.transport.SearchAnomalyResultAction;
 import org.opensearch.client.Client;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.action.ActionResponse;
+import org.opensearch.core.common.io.stream.NamedWriteableRegistry;
 
 public class AnomalyDetectionNodeClient implements AnomalyDetectionClient {
     private final Client client;
+    private final NamedWriteableRegistry namedWriteableRegistry;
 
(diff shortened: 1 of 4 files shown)


ohltyler:
BWC failures unrelated - see them from previous PR merges

ohltyler:
UT failures unrelated - all changed files and added UT is passing.

amitgalitz (approved):
Good find!

owaiskazi19 (approved):

PR #1180, opened 2024-03-27
[backport to 2.x] rebase main

kaituo:
### Description

This PR rebases the current branch onto the latest version of `main`. This integration brings in all the recent refactoring changes from the `main` branch, making it easier to release new enhancements.

Note the PR includes https://github.com/opensearch-project/anomaly-detection/commit/f0ed43b1aa3c5d80dc6f23a5dd4465bb03ef1739 where we renamed the plugin name from opendistro_anomaly_detector to opensearch_time_series_analytics.

This PR aligns the 2.x branch with its existing dependencies and reverts breaking changes in OpenSearch 3.0, ensuring smooth continuation of development and release processes in 2.x.

* due to https://github.com/opensearch-project/OpenSearch/issues/3351 and https://github.com/opensearch-project/OpenSearch/pull/4459, there are breaking changes in 3.0 such that we migrated client transports to Apache HttpClient / Core 5.x. Since 2.x has no such changes, I reverted HTTP and content type classes from `org.apache.hc.core5` to `org.apache.http` to align with libraries present in 2.x.

```
-org.apache.hc.core5.http.HttpEntity;
+org.apache.http.HttpEntity;

-org.apache.hc.core5.http.ContentType;
+org.apache.http.entity.ContentType;

-org.apache.hc.core5.http.Header;
+org.apache.http.Header;

-org.apache.hc.core5.http.HttpEntity;
+org.apache.http.HttpEntity;

-org.apache.hc.core5.http.io.entity.StringEntity;
+org.apache.http.entity.StringEntity;

-org.apache.hc.core5.http.HttpHeaders
+org.apache.http.HttpHeaders

-org.apache.hc.core5.http.message.BasicHeader
+org.apache.http.message.BasicHeader

-org.apache.hc.core5.http.HttpHost
+org.apache.http.HttpHost

-org.apache.hc.core5.http.io.entity.EntityUtils
+org.apache.http.util.EntityUtils
```

* Adjusted broadcast request handling to remain with `BaseNodeRequest` instead of migrating to `TransportRequest`, maintaining 2.x consistency.

```
-org.opensearch.transport.TransportRequest;
+org.opensearch.action.support.nodes.BaseNodeRequest
```

* In ODFERestTestCase
  * configureHttpsClient: conf.setSocketTimeout takes int instead of Timeout as a parameter.
  * configureHttpsClient:
    * use current code as there is no Apache HttpClient / Core 5.x library in 2.x.
    * MediaType.fromMediaType takes a string instead of Header as a parameter.

* 2.x won't throw ParseException during deserialzing XContent. Thus, I removed unnecessary `ParseException` declarations in test cases
  * HistoricalAnalysisRestApiIT
  * HistoricalAnalysisRestTestCase
  * ADRestTestUtils

* Maintained version 1.3 in `.github/labeler.yml`

Testing Summary:
- CI passed


.github/workflows/test_security.yml
```diff
@@ -57,6 +55,7 @@ jobs:
           then
             echo "FROM opensearchstaging/opensearch:$docker_version" >> Dockerfile
             echo "RUN if [ -d /usr/share/opensearch/plugins/opensearch-anomaly-detection ]; then /usr/share/opensearch/bin/opensearch-plugin remove opensearch-anomaly-detection; fi" >> Dockerfile
+            echo "RUN if [ -d /usr/share/opensearch/plugins/opensearch-time-series-analytics ]; then /usr/share/opensearch/bin/opensearch-plugin remove opensearch-time-series-analytics; fi" >> Dockerfile
```
ohltyler, line 58: This implies a plugin name change that I imagine could have lots of other impacts, like infra builds / scripts or frontend plugin builds / scripts that rely on this. We should consider this a breaking change I don't believe we want in 2.x, lmk what you think.

I also see lots of changes related to cluster settings and REST APIs. Could you confirm those are all internal facing and nothing changes from an external/naming perspective? If the latter, same concern as above.

kaituo: > We should consider this a breaking change I don't believe we want in 2.x, lmk what you think.
I have already synced with infra team and will work with them after 2.13 release.

> Could you confirm those are all internal facing and nothing changes from an external/naming perspective? 
Those are internal facing.

ohltyler: Sounds good. Let's call out the only change is related to the plugin name, rest of the functionality remains the same. Sounds like infra will be handled, can you handle tracking of changes on frontend plugin CIs?

ohltyler (approved):
LGTM! Thanks for all of the details. This will indeed make development on main and backporting -> 2.x simpler with all of the refactoring changes now included in 2.x :)

dbwiddis (approved):

Issue #1213, opened 2024-05-28
[FEATURE] AD 2.15 & 2.17 Improvements


jackiehanyang:
### Memory Circuit Breaker Threshold Adjustment  (_targeting 2.15 release_)

**_Problem Statement:_**
Currently, the memory circuit breaker threshold is hardcoded to 85%. There is a need for a dynamic setting that allows adjustment of this threshold. The system should dynamically read the open memory circuit breaker threshold on each node, particularly since the customer occasionally experiences memory usage spikes up to 88%.

**_Action Items:_**
- [x] Update default threshold to 95%. 
- [x] Update upper threshold limit to 98%.
- [x] Make `JVM_HEAP_USAGE_THRESHOLD` a dynamic setting, where customers can adjust it dynamically.
- [x] e2e testing

-----
### Result Index Lifecycle Management _(targeting 2.15 release)_
**_Problem Statement:_**
Currently, customers must use ISM to manage their custom result indices. We aim to implement lifecycle management for result indices within AD for customers, which includes creating daily result indices and setting a retention period for the data.

**_Action Items:_**
- [x] On AD-Dashboard detector creation page, add three result index lifecycle management conditions: `max_index_age_condition`, `max_index_size_condition`, and `index_cleanup_intervals`
- [x] On AD-Dashboard detector details page, detector configurations tab, display these three conditions after Custom result index.
- [x] Update AD detector [config](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/timeseries/model/Config.java) to read these three conditions at backend
- [x] Update [the current index management logic in AD](https://github.com/opensearch-project/anomaly-detection/blob/8141c27b3c2039f4fe239f9d6480b923c441f23f/src/main/java/org/opensearch/ad/indices/ADIndexManagement.java#L228) to include managing the lifecycle of custom result indices based on three newly added conditions.
- [x] Update all related UTs and ITs
----
### Flattening result index Nested/array Values for Dashboard Viewing _(targeting 2.17 release)_
_**Problem Statement:**_
Many values are not flattened, making it difficult to view them on the dashboard. For instance, entity values are nested objects, and features are arrays. The requirement is to reference a feature by name and apply conditions like `f1 > 3`. Additionally, there is a need to perform terms aggregation on categorical fields. This will require adjustments to the mapping and the addition of new fields in the result index.

**_Solution:_**
[...]

PR #1215, opened 2024-05-31
Add custom result index lifecycle management condition fields to config

jackiehanyang:
### Description
Add custom result index lifecycle management condition fields to config

### Issues Resolved
https://github.com/opensearch-project/anomaly-detection/issues/1213

### DIFF
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -147,6 +147,9 @@ public class AnomalyDetector extends Config {
      * @param seasonIntervals seasonality in terms of intervals
      * @param historyIntervals history intervals we look back during cold start
      * @param rules custom rules to filter out AD results
+     * @param customResultIndexMinSize custom result index lifecycle management min size condition
+     * @param customResultIndexMinAge custom result index lifecycle management min age condition
+     * @param customResultIndexTTL custom result index lifecycle management ttl
      */
     public AnomalyDetector(
         String detectorId,
@@ -170,7 +173,10 @@ public class AnomalyDetector extends Config {
         Integer recencyEmphasis,
         Integer seasonIntervals,
         Integer historyIntervals,
-        List<Rule> rules
+        List<Rule> rules,
+        Integer customResultIndexMinSize,
+        Integer customResultIndexMinAge,
+        Integer customResultIndexTTL
     ) {
         super(
             detectorId,
@@ -194,7 +200,10 @@ public class AnomalyDetector extends Config {
             recencyEmphasis,
             seasonIntervals,
             new ADShingleGetter(seasonIntervals),
-            historyIntervals
+            historyIntervals,
+            customResultIndexMinSize,
+            customResultIndexMinAge,
+            customResultIndexTTL
         );
 
         checkAndThrowValidationErrors(ValidationAspect.DETECTOR);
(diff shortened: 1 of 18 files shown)


src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -294,6 +306,21 @@ public Config(StreamInput input) throws IOException {
         this.recencyEmphasis = input.readInt();
         this.seasonIntervals = input.readInt();
         this.historyIntervals = input.readInt();
+        if (input.readBoolean()) {
+            this.customResultIndexMinSize = input.readInt();
+        } else {
+            customResultIndexMinSize = null;
```
kaituo, line 312: can you add this. for all 3 fields to make the style consistent?

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -343,6 +370,21 @@ public void writeTo(StreamOutput output) throws IOException {
         output.writeInt(recencyEmphasis);
         output.writeInt(seasonIntervals);
         output.writeInt(historyIntervals);
+        if (customResultIndexMinSize != null) {
```
kaituo, line 373: Can you just to use writeOptionalInt and readOptionalInt that would writeBoolean true/false based on null check?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -330,6 +348,21 @@ public void writeTo(StreamOutput output) throws IOException {
         } else {
             output.writeBoolean(false);
         }
+        if (customResultIndexMinSize != null) {
+            output.writeOptionalInt(customResultIndexMinSize);
```
kaituo, line 352: You can remove null check as writeOptinalInt covers that.

src/main/java/org/opensearch/timeseries/model/Config.java
jackiehanyang: sure, will update it

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
jackiehanyang: using `writeOptinalInt` in both AD and Config file

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -268,14 +277,23 @@ public AnomalyDetector(StreamInput input) throws IOException {
         if (input.readBoolean()) {
             this.rules = input.readList(Rule::new);
         }
+        if (input.readBoolean()) {
```
kaituo, line 280: You can use readOptionalInt now.

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -294,6 +306,21 @@ public Config(StreamInput input) throws IOException {
         this.recencyEmphasis = input.readInt();
         this.seasonIntervals = input.readInt();
         this.historyIntervals = input.readInt();
+        if (input.readBoolean()) {
```
kaituo, line 309: You can use readOptionalInt now.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
jackiehanyang: good catch! forgot to update this part

src/main/java/org/opensearch/timeseries/model/Config.java
jackiehanyang: Updated

kaituo (approved):

kaituo (approved):

PR #1225, opened 2024-06-06
Make Custom Result Index Name an Alias

kaituo:
### Description
Previously, we used configured custom result index names as the index name, creating the index only when a detector was created. This approach caused several problems for index management:

1. **Index State Management (ISM) Incompatibility**: ISM requires an alias for rolling over an index, but we only had the index name.
2. **Index Recognition Issue**: Even if we managed to roll over an index, AD could not recognize it because the index would not be recreated unless the detector was recreated.
3. **Result History and Top Anomaly Results**: The result history on the dashboard and top anomaly results were reading from a single index instead of an index pattern. Thus, after rolling over an index, the result history and top anomaly results would be lost.

This PR addresses these issues:

1. **Custom Result Index Creation**: We now create a custom result index with the name `<custom-name-history-{now/d}-1>` and use the alias `custom-name` to point to it.
2. **Index Recreation**: We recreate an index when failing to find a result index, updating the configured alias to point to the new index. This ensures continuity when an index is rolled over and new indices do not exist.
3. **Query Index Pattern**: The top anomaly result API now queries an index pattern instead of a single index. The result history on the dashboard follows the same approach. The dashboard code will be posted in a separate PR.

Additionally, this PR updates the custom result index mapping when the mapping is outdated, similar to existing logic on the default result index.

**Testing Done**:
* Successfully updated the mapping of the custom result index when the mapping is outdated.
* Verified that the frontend can still see old and new results after a rollover.
* Confirmed that the backend can still write to new indices after a rollover.
* Ensured all existing tests pass.

More tests will be added in the following PRs.


jackiehanyang:
Synced offline about this pr. LGTM

jackiehanyang (approved):

PR #1262, opened 2024-07-08
Update PULL_REQUEST_TEMPLATE to include an API spec change in the checklist.

dblock:
Coming from https://github.com/opensearch-project/opensearch-api-specification/issues/387, update PR template to include a checkbox that requests that all API changes be documented in the OpenAPI spec used to generate clients.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,16 @@
+### Description
+[Describe what this change achieves]
+
+### Related Issues
+Resolves #[Issue number to be closed when this PR is merged]
+<!-- List any other related issues here -->
+
+### Check List
+- [ ] New functionality includes testing.
+- [ ] New functionality has been documented.
+- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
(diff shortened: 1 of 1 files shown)


owaiskazi19 (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+[Describe what this change achieves]
+
+### Related Issues
+Resolves #[Issue number to be closed when this PR is merged]
+<!-- List any other related issues here -->
+
+### Check List
+- [ ] New functionality includes testing.
+- [ ] New functionality has been documented.
+- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
```
vamshin, line 11: To me its not coming out clear from messaging side. Should we say something like

- [ ] New/Modify Existing APIs. API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).

dblock: I propose to keep it consistent with other repos (x20 PRs), this is the language that was chosen in core.

dblock:
@owaiskazi19 can we merge?

PR #1287, opened 2024-08-30
Prevent resetting latest flag of real-time when starting historical analysis

kaituo:
### Description
This PR addresses a bug where starting a historical analysis after a real-time analysis on the same detector caused the real-time task’s latest flag to be incorrectly reset to false by the historical run.

The fix ensures that only the latest flags of the same analysis type are reset:
* Real-time analysis will only reset the latest flag of previous real-time analyses.
* Historical analysis will only reset the latest flag of previous historical analyses.

Additional Changes:
* Set a minimum recencyEmphasis value of 2 to align with RCF requirements.
* Updated imputation logic to execute only when the entity maps to the local node ID in the hash ring, reducing errors when models of the same entity exist on multiple nodes.

Testing:
- Added an integration test to reproduce the bug and verified the fix.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/transport/ADHCImputeTransportAction.java
+++ b/src/main/java/org/opensearch/ad/transport/ADHCImputeTransportAction.java
@@ -19,6 +19,7 @@ import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.nodes.TransportNodesAction;
 import org.opensearch.ad.caching.ADCacheProvider;
 import org.opensearch.ad.ml.ADRealTimeInferencer;
+import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.core.common.io.stream.StreamInput;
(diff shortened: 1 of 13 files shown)


src/main/java/org/opensearch/ad/transport/ADHCImputeTransportAction.java
```diff
@@ -156,4 +159,46 @@ protected ADHCImputeNodeResponse nodeOperation(ADHCImputeNodeRequest nodeRequest
...
+    /**
+     * Determines whether the model state should be processed based on various conditions.
+     *
+     * Conditions checked:
+     * - The model's last seen execution end time is not the minimum Instant value.
+     * - The current execution end time is greater than or equal to the model's last seen execution end time,
+     *   indicating that the model state was updated in previous intervals.
+     * - The entity associated with the model state is present.
+     * - The owning node for real-time processing of the entity, with the same local version, is present in the hash ring.
+     * - The owning node for real-time processing matches the current local node.
```
amitgalitz, line 171: before we could have occasions where two nodes have the same model for the same entity, so two nodes at the same time will find if there is an anomaly on the interval?

kaituo: yes.

amitgalitz: would this mean we would write result twice or we handled it later to only be written once?

amitgalitz (approved):

src/main/java/org/opensearch/ad/transport/ADHCImputeTransportAction.java
kaituo: We would only write it once (not guaranteed) as here we only process imputation on a node the entity is mapped to.

PR #1296, opened 2024-09-05
Passing user to getLatestDataTime call

amitgalitz:
### Description
Changed getLatestDataTime so we pass the correct user so we have the right user info when querying remote indices, otherwise we get a failure like: 
```
nested: RemoteTransportException[[35692e6c255f][172.22.0.4:9300][indices:data/read/search]]; nested: OpenSearchSecurityException[No user found for indices:data/read/search];
```


jackiehanyang (approved):

kaituo:
Can you add a test to cover the bug? If it is too time consuming, we can do it after cut off.

amitgalitz:
> Can you add a test to reproduce the bug? If it is too time consuming, we can do it after cut off.

It only occurs on remote index, not 100% sure on best way to have integration with remote clusters, might take me longer then cut off

PR #1351, opened 2024-10-23
Fix race condition in PageListener

kaituo:
### Description
This PR
- Introduced an `AtomicInteger` called `pagesInFlight` to track the number of pages currently being processed. 
- Incremented `pagesInFlight` before processing each page and decremented it after processing is complete
- Adjusted the condition in `scheduleImputeHCTask` to check both `pagesInFlight.get() == 0` (all pages have been processed) and `sentOutPages.get() == receivedPages.get()` (all responses have been received) before scheduling the `imputeHC` task. 
- Removed the previous final check in `onResponse` that decided when to schedule `imputeHC`, relying instead on the updated counters for accurate synchronization.

These changes address the race condition where `sentOutPages` might not have been incremented in time before checking whether to schedule the `imputeHC` task. By accurately tracking the number of in-flight pages and sent pages, we ensure that `imputeHC` is executed only after all pages have been fully processed and all responses have been received.

Testing done:
1. Reproduced the race condition by starting two detectors with imputation. This causes an out of order illegal argument exception from RCF due to this race condition. Also verified the change fixed the problem.
2. added an IT for the above scenario.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
+++ b/src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
@@ -229,15 +229,15 @@ public abstract class ResultProcessor<TransportResultRequestType extends ResultR
 
         @Override
         public void onResponse(CompositeRetriever.Page entityFeatures) {
+            // Increment pagesInFlight to track the processing of this page
+            pagesInFlight.incrementAndGet();
+
             // start processing next page after sending out features for previous page
             if (pageIterator.hasNext()) {
                 pageIterator.next(this);
             } else if (config.getImputationOption() != null) {
                 scheduleImputeHCTask();
             }
-
-            // Increment pagesInFlight to track the processing of this page
-            pagesInFlight.incrementAndGet();
 
             if (entityFeatures != null && false == entityFeatures.isEmpty()) {
                 LOG
--- a/src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
+++ b/src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
@@ -210,6 +210,10 @@ public abstract class ResultProcessor<TransportResultRequestType extends ResultR
         private String taskId;
         private AtomicInteger receivedPages;
         private AtomicInteger sentOutPages;
+        // By introducing pagesInFlight and incrementing it in the main thread before asynchronous processing begins,
+        // we ensure that the count of in-flight pages is accurate at all times. This allows us to reliably determine
+        // when all pages have been processed.
+        private AtomicInteger pagesInFlight;
 
         PageListener(PageIterator pageIterator, Config config, long dataStartTime, long dataEndTime, String taskId) {
             this.pageIterator = pageIterator;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
```diff
@@ -220,14 +224,21 @@ class PageListener implements ActionListener<CompositeRetriever.Page> {
...
+            this.pagesInFlight = new AtomicInteger();
         }
 
         @Override
         public void onResponse(CompositeRetriever.Page entityFeatures) {
             // start processing next page after sending out features for previous page
             if (pageIterator.hasNext()) {
                 pageIterator.next(this);
+            } else if (config.getImputationOption() != null) {
+                scheduleImputeHCTask();
```
amitgalitz, line 236: Why are we first going inside here and then incrementing the pages inFlight, shouldn't we first increment?

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
```diff
@@ -309,19 +320,15 @@ public void onResponse(CompositeRetriever.Page entityFeatures) {
...
-                    scheduleImputeHCTask();
-                } else {
-                    // no data in current interval. Send out impute request right away.
-                    imputeHC(dataStartTime, dataEndTime, configId, taskId);
-                }
-
+            } else {
+                // No entity features to process
+                // Decrement pagesInFlight immediately
+                pagesInFlight.decrementAndGet();
```
amitgalitz, line 331: Just to confirm, here every page is from the feature aggregation results, we want to make sure we received every page and then sending it out to processing (means sending the aggregated feature data to the correct model and doing .process())? then after we check each entity if data was received and send impute call to place the imputed value?

amitgalitz:
Overall were we processing results of detector 1 still but then detector 2 already finished and went to scheduleImputeHCTask?

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
kaituo: yeah, changed

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
kaituo: yes

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
amitgalitz: I am not 100% sure on this actually, then this `pagesInFlight.get() == 0` will never be reached? I was just thinking of first case also if its 0 it might pass right away

kaituo:
> Overall were we processing results of detector 1 still but then detector 2 already finished and went to scheduleImputeHCTask?

scheduleImputeHCTask is detector specific. So when detector 2 is finished, scheduleImputeHCTask will start regardless of detector 1.

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
kaituo: When we finished processing all of the inflight requests, pagesInFlight.get() == 0, right?

amitgalitz (approved):

PR #1358, opened 2024-10-31
Adding feature direction rules

amitgalitz:
### Description
This PR expands on the custom supression rules that we have added in 2.17. Users now have the ability to configures rules which have x number of conditions in them regarding the feature. 

Rules can look like this:
```
rules": [
            {
              "action": "IGNORE_ANOMALY",
              "conditions": [
                {
                  "feature_name": "logVolume",
                  "threshold_type": "
ACTUAL_OVER_EXPECTED_RATIO",
                  "value": 0.3,
                  "operator": LTE
                }
              ]
            }
    ]
```
This rule means that we will ignore anomalies if the actual value is less than or equal too 30% higher than the expected.
For example: "Ignore anomalies for feature logVolume when the actual value is no more than 30% above the expected value."

These first sets of threshold are based on either ration or value margin between the actual and expected value compared to a given value by the user. These rules are given to the RCF model and RCF code takes care of applying these. 

In this PR we are adding two more generic rules that are simply checking if the actual value is below or above the expected value. These rules wont need a value or operator as they are more absolute comparison. Currently RCF doesn’t have the logic to handles such rules so we decided to go with the approach of overriding the anomaly grade in the Anomaly Result. This takes care of both historical, real time and preview results. 

Newly accepted payloads:
```
rules": [
            {
              "action": "IGNORE_ANOMALY",
              "conditions": [
                {
                  "feature_name": "sum_http_4xx",
                  "threshold_type": "ACTUAL_IS_OVER_EXPECTED",
                  "value": null,
                  "operator": null
                }
              ]
            }
    ]
```
or

```
"conditions": [
                {
                  "feature_name": "sum_http_4xx",
                  "threshold_type": "ACTUAL_IS_OVER_EXPECTED"
                }
              ]
 ```             
Edge cases:

* I first consider past values for comparison with expected value if they are present
* We only care about the feature with the highest attribution when looking at the rules, if they are equal then we look at both. 

### Related Issues
#616 

### Check List
- [x] New functionality includes testing.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java
+++ b/src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java
@@ -54,7 +54,7 @@ public class IgnoreSimilarExtractor {
             for (Rule rule : rules) {
                 for (Condition condition : rule.getConditions()) {
                     if (condition.getThresholdType() != ThresholdType.ACTUAL_IS_BELOW_EXPECTED
-                        || condition.getThresholdType() != ThresholdType.ACTUAL_IS_OVER_EXPECTED) {
+                        && condition.getThresholdType() != ThresholdType.ACTUAL_IS_OVER_EXPECTED) {
                         processCondition(
                             condition,
                             featureNames,
--- a/src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java
+++ b/src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java
@@ -53,15 +53,18 @@ public class IgnoreSimilarExtractor {
         if (rules != null) {
             for (Rule rule : rules) {
                 for (Condition condition : rule.getConditions()) {
-                    processCondition(
-                        condition,
-                        featureNames,
-                        baseDimension,
-                        ignoreSimilarFromAbove,
-                        ignoreSimilarFromBelow,
-                        ignoreSimilarFromAboveByRatio,
-                        ignoreSimilarFromBelowByRatio
-                    );
+                    if (condition.getThresholdType() != ThresholdType.ACTUAL_IS_BELOW_EXPECTED
+                        || condition.getThresholdType() != ThresholdType.ACTUAL_IS_OVER_EXPECTED) {
+                        processCondition(
+                            condition,
+                            featureNames,
+                            baseDimension,
+                            ignoreSimilarFromAbove,
+                            ignoreSimilarFromBelow,
+                            ignoreSimilarFromAboveByRatio,
+                            ignoreSimilarFromBelowByRatio
+                        );
+                    }
                 }
             }
         }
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java [hunk 1]
```diff
@@ -53,15 +53,18 @@ public static ThresholdArrays processDetectorRules(AnomalyDetector detector) {
...
-                        condition,
-                        featureNames,
-                        baseDimension,
-                        ignoreSimilarFromAbove,
-                        ignoreSimilarFromBelow,
-                        ignoreSimilarFromAboveByRatio,
-                        ignoreSimilarFromBelowByRatio
-                    );
+                    if (condition.getThresholdType() != ThresholdType.ACTUAL_IS_BELOW_EXPECTED
+                        || condition.getThresholdType() != ThresholdType.ACTUAL_IS_OVER_EXPECTED) {
```
ohltyler, line 57: should this be AND instead of OR?

src/main/java/org/opensearch/ad/model/Condition.java
```diff
@@ -70,11 +70,19 @@ public static Condition parse(XContentParser parser) throws IOException {
...
+                case "operator":
+                    if (parser.currentToken() == XContentParser.Token.VALUE_NULL) {
+                        operator = null; // Set operator to null if the field is missing
+                    } else {
+                        operator = Operator.valueOf(parser.text().toUpperCase(Locale.ROOT));
+                    }
                     break;
                 case VALUE_FIELD:
-                    value = parser.doubleValue();
+                    if (parser.currentToken() == XContentParser.Token.VALUE_NULL) {
```
ohltyler, line 81: super nit: add same comment as line 75?

src/main/java/org/opensearch/timeseries/transport/ResultResponse.java
```diff
@@ -89,7 +90,7 @@ public boolean shouldSave() {
     }
 
     public abstract List<IndexableResultType> toIndexableResults(
-        String configId,
+        Config configId,
```
ohltyler, line 93: nit - `Config config`

src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java
jackiehanyang: +1, this will always evaluate to true

src/main/java/org/opensearch/ad/model/Condition.java
```diff
@@ -70,11 +70,19 @@ public static Condition parse(XContentParser parser) throws IOException {
                 case THRESHOLD_TYPE_FIELD:
                     thresholdType = ThresholdType.valueOf(parser.text().toUpperCase(Locale.ROOT));
                     break;
-                case OPERATOR_FIELD:
-                    operator = Operator.valueOf(parser.text().toUpperCase(Locale.ROOT));
+                case "operator":
```
jackiehanyang, line 73: make it an enum to keep the coding style consistent

src/main/java/org/opensearch/ad/ml/IgnoreSimilarExtractor.java
amitgalitz: yes good catch

src/main/java/org/opensearch/ad/model/Condition.java
amitgalitz: added

ohltyler (approved):
LGTM!

jackiehanyang (approved):

PR #1366, opened 2024-11-14
Not blocking detector creation on unknown feature validation error

### LINKED ISSUE #1365: [BUG] Feature validation blocks detector creation on all errors
**What is the bug?**
Currently we have validation when creating a detector to check if the configurations given are correct. For the feature fields we check that we need to have at least 1 feature and it needs to have proper syntax. Additionally we also check if the given aggregation type and the given field will produce any errors when aggregating against the source index. We currently fail if there is any exception, however there could be cases where the query momentarily times out or we have some other issue unrelated to AD that can cause a failure, in such a a case we shouldn't block AD creation completely.

This is a little difficult to reproduce but one way to do this is to create an underscaled cluster and create detector when CPU is extremly high on a very large data source index. IT can also be replicated by changing the search backpressure setting or timeout setting to make timeouts more likely

amitgalitz:
### Description
Currently during detector creation we check for a few different things in relation to the configured feature. If aggregation is valid, if we have at least one feature, that we don't aggregate over a non numerical field with something like average.
However we currently fail on any seen exception, we want to change this that if there is an unknown exception that we don't completely block detector creation. This is due to the fact that we might get things like timeout or search cancellation that are related to the cluster, not cause the configuration is wrong. Users should be able to still create the detector and handle any cluster or settings changes they need to later on if it effects the actual detector creation.

### Related Issues
Resolves #1365

### Check List

### DIFF
--- a/src/main/java/org/opensearch/timeseries/rest/handler/AbstractTimeSeriesActionHandler.java
+++ b/src/main/java/org/opensearch/timeseries/rest/handler/AbstractTimeSeriesActionHandler.java
@@ -890,6 +890,7 @@ public abstract class AbstractTimeSeriesActionHandler<T extends ActionResponse, 
                 feature.getId()
             );
             ssb.aggregation(internalAgg.getAggregatorFactories().iterator().next());
+            ssb.trackTotalHits(false);
             SearchRequest searchRequest = new SearchRequest().indices(config.getIndices().toArray(new String[0])).source(ssb);
             ActionListener<SearchResponse> searchResponseListener = ActionListener.wrap(response -> {
                 Optional<double[]> aggFeatureResult = searchFeatureDao.parseResponse(response, Arrays.asList(feature.getId()), false);
@@ -907,11 +908,17 @@ public abstract class AbstractTimeSeriesActionHandler<T extends ActionResponse, 
                 String errorMessage;
                 if (isExceptionCausedByInvalidQuery(e)) {
                     errorMessage = CommonMessages.FEATURE_WITH_INVALID_QUERY_MSG + feature.getName();
+                    logger.error(errorMessage, e);
+                    multiFeatureQueriesResponseListener.onFailure(new OpenSearchStatusException(errorMessage, RestStatus.BAD_REQUEST, e));
                 } else {
                     errorMessage = CommonMessages.UNKNOWN_SEARCH_QUERY_EXCEPTION_MSG + feature.getName();
+                    logger.error(errorMessage, e);
+                    // If we see an unexpected error such as timeout or some task cancellation cause of search backpressure
+                    // we don't want to block detector creation as this is unlikely an error due to wrong configs
+                    // but we want to record what error was seen
+                    multiFeatureQueriesResponseListener
+                        .onResponse(new MergeableList<>(new ArrayList<>(List.of(Optional.of(new double[] { 1.0 })))));
                 }
-                logger.error(errorMessage, e);
-                multiFeatureQueriesResponseListener.onFailure(new OpenSearchStatusException(errorMessage, RestStatus.BAD_REQUEST, e));
             });
             clientUtil.asyncRequestWithInjectedSecurity(searchRequest, client::search, user, client, context, searchResponseListener);
         }


src/main/java/org/opensearch/timeseries/rest/handler/AbstractTimeSeriesActionHandler.java
```diff
@@ -907,11 +908,17 @@ protected void validateConfigFeatures(String id, boolean indexingDryRun, ActionL
...
                 if (isExceptionCausedByInvalidQuery(e)) {
                     errorMessage = CommonMessages.FEATURE_WITH_INVALID_QUERY_MSG + feature.getName();
+                    logger.error(errorMessage, e);
+                    multiFeatureQueriesResponseListener.onFailure(new OpenSearchStatusException(errorMessage, RestStatus.BAD_REQUEST, e));
                 } else {
                     errorMessage = CommonMessages.UNKNOWN_SEARCH_QUERY_EXCEPTION_MSG + feature.getName();
+                    logger.error(errorMessage, e);
+                    // If we see an unexpected error such as timeout or some task cancellation cause of search backpressure
+                    // we don't want to block detector creation as this is unlikely an error due to wrong configs
+                    // but we want to record what error was seen
```
owaiskazi19, line 918: 1. Are we sure the error in this block is related to timeout on the query on the OS cluster side?
2. It will be difficult to write IT for this change but we can still have UTs?
3. Why `(List.of(Optional.of(new double[] { 1.0 }` and not `Collections.singletonList(Optional.empty())` below?

amitgalitz: 1. Yes we have tested this in production with both timeout and search backpressure blocks
2. I wrote a UT
3. CHanged to empty. either works here, its more that some results exists not the content

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
```diff
@@ -250,4 +251,63 @@ public void testValidateMoreThanTenMultiEntityDetectorsLimit() throws IOExceptio
...
+        when(userIndexResponse.getHits()).thenReturn(TestHelpers.createSearchHits(userIndexHits));
+        AnomalyDetector singleEntityDetector = TestHelpers.randomAnomalyDetector(TestHelpers.randomUiMetadata(), null, true);
+
+        SearchResponse configInputIndicesResponse = mock(SearchResponse.class);
+        when(configInputIndicesResponse.getHits()).thenReturn(TestHelpers.createSearchHits(2));
+
+        // extend NodeClient since its execute method is final and mockito does not allow to mock final methods
+        // we can also use spy to overstep the final methods
+        NodeClient client = IndexAnomalyDetectorActionHandlerTests
+            .getCustomNodeClient(detectorResponse, userIndexResponse, configInputIndicesResponse, singleEntityDetector, threadPool);
```
owaiskazi19, line 275: Can't we just mock the client here?

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
```diff
@@ -250,4 +251,63 @@ public void testValidateMoreThanTenMultiEntityDetectorsLimit() throws IOExceptio
...
+            method,
+            xContentRegistry(),
+            null,
+            searchFeatureDao,
+            ValidationAspect.DETECTOR.getName(),
+            clock,
+            settings
+        );
+
+        final CountDownLatch inProgressLatch = new CountDownLatch(1);
```
owaiskazi19, line 301: Instead of latch we can use ActionFuture here

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -405,10 +405,13 @@ public void testCreateAnomalyDetectorWithCustomResultIndex() throws IOException
         AnomalyDetector detector = cloneDetector(anomalyDetector, resultIndex);
         // User goat has no permission to create index
         Exception exception = expectThrows(IOException.class, () -> { createAnomalyDetector(detector, true, goatClient); });
+        System.out.println("error message: " + exception.getMessage());
```
owaiskazi19: Debug logs?

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
amitgalitz: I tried mocking the client here: https://github.com/opensearch-project/anomaly-detection/blob/main/src/test/java/org/opensearch/ad/transport/ValidateAnomalyDetectorTransportActionTests.java#L374
But needed to mock a lot more requests, I changed to use the way we have in this file where we needed to create a customNodeClient where based on the comment it was needed cause we can't mock final methods. I tried also if we just use `client = mock(Client.class);` we can't use spy:
```
Please don't pass mock here. Spy is not allowed on mock.
java.lang.IllegalArgumentException: Please don't pass mock here. Spy is not allowed on mock.
```

owaiskazi19: Yeah that's because we can't spy a mocked object.

I see there's a method already to fetch `getCustomNodeClient` and the only difference here is the featureConfig check. Why not to combine both the method together to reduce code duplication?

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
```diff
@@ -250,4 +251,63 @@ public void testValidateMoreThanTenMultiEntityDetectorsLimit() throws IOExceptio
         assertTrue(inProgressLatch.await(100, TimeUnit.SECONDS));
         verify(clientSpy, never()).execute(eq(GetMappingsAction.INSTANCE), any(), any());
     }
+
+    // This test also validates that if we get a non timeseries exception or not an invalid query that we will not completely block
+    // detector creation, this is applicable like things when we get timeout not cause of AD configuration errors but because cluster
+    // is momentarily under utilized.
+    public void testValidateMoreThanTenMultiEntityDetectorsLimitDuplicateNameFailure() throws IOException, InterruptedException {
+        SearchResponse mockResponse = mock(SearchResponse.class);
+        int totalHits = 1;
```
owaiskazi19, line 260: Nit: We don't need this variable since used just once

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
```diff
@@ -250,4 +251,63 @@ public void testValidateMoreThanTenMultiEntityDetectorsLimit() throws IOExceptio
...
+        when(mockResponse.getHits()).thenReturn(TestHelpers.createSearchHits(totalHits));
+        SearchResponse detectorResponse = mock(SearchResponse.class);
+        when(detectorResponse.getHits()).thenReturn(TestHelpers.createSearchHits(totalHits));
+        SearchResponse userIndexResponse = mock(SearchResponse.class);
+        int userIndexHits = 0;
```
owaiskazi19, line 265: Same as above

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
amitgalitz: CountDownLatch seems to do the job here since I don't need to do anything special and I only have one call.
Do you mean something like:
```
PlainActionFuture<ValidateConfigResponse> future = PlainActionFuture.newFuture();
        handler.start(future);
        try {
            future.actionGet(100, TimeUnit.SECONDS);
            fail("should not reach here");
        } catch (ValidationException e) {
            String errorMsg = String.format(
                    Locale.ROOT,
                    IndexAnomalyDetectorActionHandler.EXCEEDED_MAX_HC_DETECTORS_PREFIX_MSG,
                    maxMultiEntityAnomalyDetectors
            );
            assertTrue(e.getMessage().contains(errorMsg));
        }
        ```

owaiskazi19: Yes, the above looks good

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
amitgalitz: on accident, removed this

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
amitgalitz: used twice

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
amitgalitz: removed

src/test/java/org/opensearch/action/admin/indices/mapping/get/ValidateAnomalyDetectorActionHandlerTests.java
amitgalitz: combined the methods

owaiskazi19 (approved):
Looks much cleaner! Thanks for addressing the comments

PR #1473, opened 2025-05-09
Switch guava deps from compileOnly to implementation

cwperks:
### Description

AD has its guava dependencies set at compileOnly level which means they are not packaged into the AD assembly and available at runtime. 

AD uses these deps at runtime, however, the way this works is because of the extending relationship between AD and JS plugin where the deps of the extended plugin are available to the extending plugin at runtime.

In https://github.com/opensearch-project/job-scheduler/pull/773, guava was removed from JS which means these deps are no longer available to AD at runtime. I'm creating this PR to change the guava deps from compileOnly to implementation to ensure they are included in the assembly and available at runtime.

### Check List
- [ ] New functionality includes testing.
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,8 +137,8 @@ dependencies {
     compileOnly "org.opensearch:opensearch-job-scheduler-spi:${job_scheduler_version}"
     implementation "org.opensearch:common-utils:${common_utils_version}"
     implementation "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
-    compileOnly group: 'com.google.guava', name: 'guava', version:'32.1.3-jre'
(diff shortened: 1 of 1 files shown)


owaiskazi19:
Looks like this change is needed for all the plugins having JS as a dependency and using guava as a compile dependency?

cwperks:
> Looks like this change is needed for all the plugins having JS as a dependency and using guava as a compile dependency?

yes, it would affect any plugin with similar configuration.

kaituo:
@cwperks @jngz-es Do you think we should revert job scheduler changes? I know a few other plugins are using job scheduler as well. Can ml-commons change to compileOnly?

kaituo (approved):

PR #1513, opened 2025-07-07
Support >1 hr intervals

kaituo:
### Description
This commit adds support for configuration intervals exceeding one hour.

Key changes:

* For intervals greater than 1 hour, models are no longer loaded directly into cache. Instead, they are sent to the cold entity queue and checkpoints are reloaded at each interval. Additionally, cold entity processing priority for long-interval configs is elevated from 'LOW' to 'MEDIUM' to ensure timely processing.
* Improved Suggest and Validate APIs: Replaced the previous median-based interval detection method with a robust adaptive "zoom-in/zoom-out" algorithm. The new method employs progressively refined date histograms to accurately determine optimal intervals. This enhancement enables validation and suggestions for intervals longer than 1 hour.

Testing:

* Conducted multi-day manual tests to verify daily interval functionality.
* Added ForecastRestApiIT.testDailyInterval integration test to validate the full forecasting workflow (interval suggestion, forecaster creation, execution, and stats verification) for daily interval data.

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List

### DIFF
--- a/src/main/java/org/opensearch/timeseries/rest/handler/IntervalCalculation.java
+++ b/src/main/java/org/opensearch/timeseries/rest/handler/IntervalCalculation.java
@@ -53,7 +53,7 @@ public class IntervalCalculation {
     private final Logger logger = LogManager.getLogger(IntervalCalculation.class);
     // keep ≤256 buckets when searching for minimum interval
     private static final int BUCKET_CAP = 256;
-    public static int MAX_SPLIT_DEPTH = 10;
+    private static final int MAX_SPLIT_DEPTH = 10;
 
     private final AggregationPrep aggregationPrep;
     private final Client client;
@@ -329,6 +329,8 @@ public class IntervalCalculation {
                 || maxAgg == null
                 || Double.isInfinite(minAgg.getValue())
                 || Double.isInfinite(maxAgg.getValue())
+                || Double.isNaN(minAgg.getValue())
+                || Double.isNaN(maxAgg.getValue())
                 || minAgg.getValue() == maxAgg.getValue()) {
                 listener.onResponse(null); // < 2 docs
                 return;
--- a/src/test/java/org/opensearch/timeseries/indices/rest/handler/IntervalCalculationTests.java
+++ b/src/test/java/org/opensearch/timeseries/indices/rest/handler/IntervalCalculationTests.java
@@ -25,7 +25,6 @@ import java.util.Arrays;
 import java.util.Map;
 
 import org.apache.lucene.search.TotalHits;
-import org.junit.After;
 import org.junit.Before;
 import org.mockito.ArgumentCaptor;
 import org.opensearch.action.search.SearchResponse;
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/timeseries/transport/EntityResultProcessor.java
```diff
@@ -165,6 +165,8 @@ public ActionListener<Optional<? extends Config>> onGetConfig(
                     );
             }
 
+            // cold entities has LOW priority, but long interval entities have normal priorities
+            RequestPriority coldEntityPriority = config.isLongInterval() ? RequestPriority.MEDIUM : RequestPriority.LOW;
```
amitgalitz, line 169: I know you commented on this in description as well but just want to understand it better:

> Additionally, cold entity processing priority for long-interval configs is elevated from 'LOW' to 'MEDIUM' to ensure timely processing.

For this change, we are always treating an entity with a long interval as a cold entity, never caching it but because we don't want to miss processing because it is low priority we put in medium priority? The other entities in cold entity queue are just entities we haven't seen for a while right? so only downside is maybe that we would be putting entities with long intervals that we haven't seen for a while still in medium queue or is that wrong for the code being executed here?

I thought for the hot entities(not the long interval ones) we wouldn't be putting them in a checkpoint read since they are in cache as well?

src/main/java/org/opensearch/timeseries/rest/handler/IntervalCalculation.java
```diff
@@ -216,109 +228,535 @@ public void onFailure(Exception e) {
...
+        SearchRequest boundsReq = new SearchRequest(config.getIndices().toArray(new String[0])).source(boundsSrc);
+
+        client.search(boundsReq, ActionListener.wrap(r -> {
+
+            Min minAgg = r.getAggregations().get("min_ts");
+            Max maxAgg = r.getAggregations().get("max_ts");
+
+            if (minAgg == null
+                || maxAgg == null
+                || Double.isInfinite(minAgg.getValue())
```
amitgalitz, line 330: should we add NaN check here also, not sure if aggregation can ever return NaN here or only if we are dividing by zero which doesn't happen here.

src/main/java/org/opensearch/timeseries/rest/handler/IntervalCalculation.java
```diff
@@ -35,6 +51,9 @@
 
 public class IntervalCalculation {
     private final Logger logger = LogManager.getLogger(IntervalCalculation.class);
+    // keep ≤256 buckets when searching for minimum interval
+    private static final int BUCKET_CAP = 256;
+    public static int MAX_SPLIT_DEPTH = 10;
```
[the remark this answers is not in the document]
kaituo: fixed

src/main/java/org/opensearch/timeseries/rest/handler/IntervalCalculation.java
kaituo: added

PR #1533, opened 2025-07-29
Onboards to centralized resource access control for detectors and forecasters

DarshitChanpura:
### Description
This PR refactors changes introduced in #1400 to utilize auto-evaluation offered by security plugin.
This skips reliance on AD plugin to call verifyAccess

NOTE: Pipelines will fail until https://github.com/opensearch-project/security/pull/5408 is merged.

### Check List
- [ ] New functionality includes testing.


src/main/java/org/opensearch/timeseries/rest/handler/AbstractTimeSeriesActionHandler.java
```diff
@@ -1092,29 +1086,8 @@ public void onResponse(IndexResponse indexResponse) {
...
-                if (shouldUseResourceAuthz) {
-                    // Share with user's backend_roles here before sending response
-                    ResourceSharingClient resourceSharingClient = ResourceSharingClientAccessor.getInstance().getResourceSharingClient();
-
-                    String configId = indexResponse.getId();
-                    String configIndex = indexResponse.getIndex();
-                    Map<Recipient, Set<String>> recipients = Map.of(Recipient.BACKEND_ROLES, Set.copyOf(user.getBackendRoles()));
-                    ShareWith shareWith = new ShareWith(Map.of(PLACE_HOLDER, new Recipients(recipients)));
-
-                    resourceSharingClient.share(configId, configIndex, shareWith, ActionListener.wrap(resourceSharing -> {
```
DarshitChanpura, line 1105: we will no longer automatically share with backend roles upon config creation. Instead, users will now have to explicitly share the config with entities at certain access-levels. This will be done via explicit REST api calls or via dashboard component. Share API is being introduced here: https://github.com/opensearch-project/security/pull/5459

src/main/java/org/opensearch/timeseries/transport/JobRequest.java
```diff
@@ -15,13 +15,15 @@
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.action.DocRequest;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.timeseries.model.DateRange;
 
-public class JobRequest extends ActionRequest {
+public class JobRequest extends ActionRequest implements DocRequest {
```
DarshitChanpura, line 23: this is how resource access requests are filtered for evaluation. Check out implementation here: https://github.com/opensearch-project/security/pull/5408/files#diff-5892c7b9dfd08da869c293e469b09654309be9994205ca939f0ce9e336d529de

src/main/java/org/opensearch/timeseries/util/ParseUtils.java
```diff
@@ -715,44 +712,22 @@ public static boolean shouldUseResourceAuthz(Settings settings) {
...
         Consumer<Object[]> onSuccess,
         Object[] successArgs,
         Consumer<Object[]> fallbackOn501,
         Object[] fallbackArgs
     ) {
-        // TODO: Remove this feature flag check once feature is GA, as it will be enabled by default
-        // detectorId will be null when this is a create request and so we don't need resource authz check
-        if (shouldEvaluateWithNewAuthz && !Strings.isNullOrEmpty(configId)) {
-            ResourceSharingClient resourceSharingClient = ResourceSharingClientAccessor.getInstance().getResourceSharingClient();
-            resourceSharingClient.verifyAccess(configId, configIndex, ActionListener.wrap(isAuthorized -> {
```
DarshitChanpura, line 743: with security plugin's lates change, access will be auto evaluated. Thus, we no longer need to explicitly call verifyAccess.

src/test/java/org/opensearch/action/admin/indices/mapping/get/IndexForecasterActionHandlerTests.java
```diff
@@ -1427,182 +1409,4 @@ public void testCreateMappingException() throws InterruptedException, IOExceptio
         }));
         assertTrue(inProgressLatch.await(10, TimeUnit.SECONDS));
     }
-
```
amitgalitz, line 1430: you removed this test, are we checking this somewhere else?

DarshitChanpura: this branch is no longer being tested here since AD no longer has knowledge of feature flag OPENSEARCH_RESOURCE_SHARING_ENABLED. It was added here specifically to cover that scenario. 

if you look index IndexForecasterActionHandler.java, you will see that I've remove relevant code as well. We no longer automatically share the resource with backend_roles upon creation, with the new authz scheme. Users will be required to explicitly share the resource. 
A share API is en-route to be added to Security plugin.

AD also has an option to add its own /share api however security is the ideal place as it avoids multiple implementations by different plugins to achieve same result.

src/main/java/org/opensearch/timeseries/util/ParseUtils.java
```diff
@@ -708,51 +704,28 @@ public static String checkFilterByBackendRoles(User requestedUser) {
      */
     public static boolean shouldUseResourceAuthz(Settings settings) {
         boolean filterByEnabled = AnomalyDetectorSettings.AD_FILTER_BY_BACKEND_ROLES.get(settings);
```
DarshitChanpura, line 706: the new authz feature is still toggleable by the old setting of `filter_by_backend_role`. However, the ask from AD maintainers is whether this should still be kept as a switch for turning on the new authz scheme?

src/main/java/org/opensearch/timeseries/rest/handler/AbstractTimeSeriesActionHandler.java
amitgalitz: Few questions:
1. Does this mean backend role filtering enabled setting doesn't do anything?
2. If this is enabled we now no longer save check the backend role when doing operations? 
3. After creating detector/config they will need to call the share api if they want anyone but themselves to have access?
4. will admin still have access to all configs/detectors

src/main/java/org/opensearch/timeseries/util/ParseUtils.java
amitgalitz: Okay so from my understanding from offline conversation:
verifyAccess will look for a few things currently: 

- global flag must be turned on along with ad flag
- request must be of DocRequest type
- any existing sharing info must have been migrated to the new model using /migrate  API offered by security plugin

src/main/java/org/opensearch/timeseries/util/ParseUtils.java
amitgalitz: yes I wanted to understand if we should be deprecating this switch already or not. From my understanding is for now:
global flag + ad flag -> new authz route
only ad flag -> current ad authz
no flag -> no sharing concept, default behavior where users can see other configs

src/main/java/org/opensearch/timeseries/rest/handler/AbstractTimeSeriesActionHandler.java
amitgalitz: After talking more offline, this is my current understanding:
1. We are still gonna have some support of our backend role filtering but this new authz will be used if both flags are enabled, see: https://github.com/opensearch-project/anomaly-detection/pull/1533#discussion_r2255341389
2. we have the backend role as part of the user information but that isn't what dictates access to the detector, the access to the resource will be dictated based on permission granted in action group as part of security plugin
3. yes for example: after creating a detector even if I have my own backend role of "HR" this isn't applied to the detector, if I want everyone with that backend role to be able to view the detector I need to call on share api to add backend role to the config ID for example for READ_ONLY or READ_WRITE
4. Only super admin will have access, just getting * (all access admin) isn't enough to see all resources that are not shared.

amitgalitz (approved):

PR #1546, opened 2025-08-12
Updates search handler to consume resource authz and updates resource authz related tests

DarshitChanpura:
### Description
Updates search handler to consume PluginClient to allow DLS style filtering for protected resources, also adds tests for resource-sharing feature enabled scenario

### Check List
- [x] New functionality includes testing.


src/main/java/org/opensearch/timeseries/transport/handler/SearchHandler.java
```diff
@@ -35,25 +36,32 @@ public class SearchHandler {
...
      * Validate user role, add backend role filter if filter enabled
      * and execute search.
      *
-     * @param request search request
+     * @param request        search request
+     * @param pair          paid of index name and id field
      * @param actionListener action listerner
      */
-    public void search(SearchRequest request, ActionListener<SearchResponse> actionListener) {
+    public void search(SearchRequest request, Pair<String, String> pair, ActionListener<SearchResponse> actionListener) {
```
DarshitChanpura, line 56: please review this. This change currently modifies all `searchHandler.search` calls to pass the index name and id field of the config. However, the search calls that are made from tasks and results indices are not covered by resource-authz. Only AD and forecast config index are, since those two are registered as protected indices.

src/main/java/org/opensearch/timeseries/util/ParseUtils.java
```diff
@@ -473,6 +474,39 @@ public static SearchSourceBuilder addUserBackendRolesFilter(User user, SearchSou
...
+    public static void addAccessibleConfigsFilterAndSearch(
+        Client client,
+        Pair<String, String> pair,
+        SearchRequest searchRequest,
+        ActionListener<SearchResponse> listener
+    ) {
+        logger.debug("Filtering result by accessible resources");
+        ResourceSharingClient resourceSharingClient = ResourceSharingClientAccessor.getInstance().getResourceSharingClient();
+        SearchSourceBuilder searchSourceBuilder = searchRequest.source();
+        resourceSharingClient.getAccessibleResourceIds(pair.getLeft(), ActionListener.wrap(configIds -> {
```
DarshitChanpura, line 486: makes as call to get accessible Id and then adds a filter to current search query to only search from those ids.

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
```diff
@@ -484,6 +487,22 @@ public void testUpdateAnomalyDetectorFlattenResultIndexField() throws Exception
             detector.getLastBreakingUIChangeTime()
         );
 
+        if (isResourceSharingFeatureEnabled()) {
+            Awaitility.await().atMost(Duration.ofSeconds(30)).pollInterval(Duration.ofMillis(200)).until(() -> {
```
DarshitChanpura, line 491: this is required as adding resource-sharing is handled by index listener in security plugin and requires some time to index resource-sharing record.

src/test/java/org/opensearch/forecast/rest/SecureForecastRestIT.java
```diff
@@ -711,20 +791,39 @@ public void testAuthApi() throws IOException, InterruptedException {
...
-                readClient,
-                "GET",
-                String.format(Locale.ROOT, GET_FORECASTER, forecasterId),
-                ImmutableMap.of(),
-                (HttpEntity) null,
-                null
+        // case 5: given a forecaster Id, read access user can
+        if (isResourceSharingFeatureEnabled()) {
+            // not look up forecaster configuration for full-client's forecaster
+            exception = expectThrows(
```
DarshitChanpura, line 796: user won't have access by default.

src/test/java/org/opensearch/forecast/rest/SecureForecastRestIT.java
```diff
@@ -1471,6 +1627,309 @@ public void testFilterBy() throws IOException {
         }
     }
 
+    public void testResourceSharing() throws IOException {
```
DarshitChanpura, line 1629: adds test specifically for resource sharing feature

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -60,14 +76,57 @@ public class SecureADRestIT extends AnomalyDetectorRestTestCase {
...
+                if (isForbidden(e)) {
+                    return null;
+                }
+                // Anything else is unexpected: fail fast
+                throw e;
+            }
+        }, notNullValue());
+    }
+
+    private void waitForRevokeNonVisibility(String detId, RestClient client) {
```
DarshitChanpura, line 103: to test that resource is no longer visible.

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -60,14 +76,57 @@ public class SecureADRestIT extends AnomalyDetectorRestTestCase {
...
     private String indexSearchAccessRole = "index_all_search";
+
+    private static final String READ_ONLY_AG = "anomaly_read_only_ag";
+    private static final String FULL_ACCESS_AG = "anomaly_full_access_ag";
+    private static final String SHARE_CONFIG_ROLE = "share_config_role";
+
     String oceanUser = "ocean";
     RestClient oceanClient;
 
+    private void waitForSharingVisibility(String detId, RestClient client) {
```
DarshitChanpura, line 87: to test that resource is visible.

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -169,132 +230,416 @@ private void confirmingClientIsAdmin() throws IOException {
...
-
-        enableFilterBy();
-        // User Dog has AD full access, but has no backend role
-        // When filter by is enabled, we block creating Detectors
-        Exception exception = expectThrows(IOException.class, () -> { createRandomAnomalyDetector(false, false, dogClient); });
-        Assert
-            .assertTrue(
-                exception.getMessage().contains("Filter by backend roles is enabled and User dog does not have backend roles configured")
-            );
+        if (isResourceSharingFeatureEnabled()) {
```
DarshitChanpura, line 233: covers resource sharing feature enabled scenario.

DarshitChanpura:
```
| -> Failed installing file:/__w/anomaly-detection/anomaly-detection/build/distributions/opensearch-anomaly-detection-3.3.0.0-SNAPSHOT.zip
| -> Rolling back opensearch-job-scheduler
| -> Rolled back opensearch-job-scheduler
| -> Rolling back file:/__w/anomaly-detection/anomaly-detection/build/distributions/opensearch-anomaly-detection-3.3.0.0-SNAPSHOT.zip
| -> Rolled back file:/__w/anomaly-detection/anomaly-detection/build/distributions/opensearch-anomaly-detection-3.3.0.0-SNAPSHOT.zip
| Exception in thread "main" java.lang.IllegalStateException: failed to load plugin opensearch-anomaly-detection due to jar hell
| 	at org.opensearch.plugins.PluginsService.checkBundleJarHell(PluginsService.java:763)
| 	at org.opensearch.plugins.PluginsService.checkJarHellForPlugin(PluginsService.java:404)
| 	at org.opensearch.tools.cli.plugin.InstallPluginCommand.jarHellCheck(InstallPluginCommand.java:834)
| 	at org.opensearch.tools.cli.plugin.InstallPluginCommand.loadPluginInfo(InstallPluginCommand.java:811)
| 	at org.opensearch.tools.cli.plugin.InstallPluginCommand.installPlugin(InstallPluginCommand.java:846)
| 	at org.opensearch.tools.cli.plugin.InstallPluginCommand.execute(InstallPluginCommand.java:277)
| 	at org.opensearch.tools.cli.plugin.InstallPluginCommand.execute(InstallPluginCommand.java:251)
| 	at org.opensearch.common.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:110)
| 	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
| 	at org.opensearch.cli.MultiCommand.execute(MultiCommand.java:104)
| 	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
| 	at org.opensearch.cli.Command.main(Command.java:101)
| 	at org.opensearch.tools.cli.plugin.PluginCli.main(PluginCli.java:66)
| Caused by: java.lang.IllegalStateException: jar hell!
| class: org.apache.commons.codec.BinaryDecoder
| jar1: /__w/anomaly-detection/anomaly-detection/build/testclusters/integTest-0/distro/3.3.0-ARCHIVE/plugins/.installing-17176381523107832485/commons-codec-1.18.0.jar
| jar2: /__w/anomaly-detection/anomaly-detection/build/testclusters/integTest-0/distro/3.3.0-ARCHIVE/plugins/opensearch-job-scheduler/commons-codec-1.18.0.jar
| 	at org.opensearch.common.bootstrap.JarHell.checkClass(JarHell.java:316)
```

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
```diff
@@ -336,7 +336,8 @@ public void onResponse(CompositeRetriever.Page entityFeatures) {
 
         @Override
         public void onFailure(Exception e) {
-            LOG.error("Unexpetected exception", e);
+            // TODO : {@link MultiEntityResultTests#testQueryErrorEndRunNotNow} fails if below line is uncommented
+            // LOG.error("Unexpetected exception", e);
```
kaituo, line 340: why would log error causes test to fail?

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
```diff
@@ -499,7 +500,8 @@ private void executeAnalysis(
             Optional<Exception> previousException = nodeStateManager.fetchExceptionAndClear(configID);
             if (previousException.isPresent()) {
                 Exception exception = previousException.get();
-                LOG.error(new ParameterizedMessage("Previous exception of [{}]", configID), exception);
+                // TODO : {@link MultiEntityResultTests#testQueryErrorEndRunNotNow} fails if below line is uncommented
```
kaituo, line 503: What changes make the test fail?

src/main/resources/plugin-additional-permissions.yml
```diff
@@ -0,0 +1,21 @@
+cluster_permissions:
```
kaituo, line 1: How will we use this file? Is it related to https://github.com/opensearch-project/opensearch-plugins/issues/238?

src/main/resources/plugin-additional-permissions.yml
```diff
@@ -0,0 +1,21 @@
...
+  - "cluster:admin/opendistro/ad*"
+  - "cluster:admin/plugin/forecast*"
+index_permissions:
+  - index_patterns:
+      - ".opendistro-anomaly-detectors"
+      - ".opendistro-anomaly-results*"
+      - ".opendistro-anomaly-checkpoints"
+      - ".opendistro-anomaly-detection-state"
+      - ".opendistro-anomaly-detector-jobs"
+      - "opensearch-ad-plugin-result*"
```
kaituo, line 13: opensearch-ad-plugin-result* is not protected index.

src/main/resources/plugin-additional-permissions.yml
```diff
@@ -0,0 +1,21 @@
...
+      - ".opensearch-forecasters"
+      - ".opensearch-forecast-checkpoints"
+      - ".opensearch-forecast-state"
+      - "opensearch-forecast-result*"
```
kaituo, line 17: opensearch-forecast-result* is not protected index.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -465,6 +465,11 @@ public static AnomalyDetector parse(
             parser.nextToken();
 
             switch (fieldName) {
+                case ID_FIELD:
```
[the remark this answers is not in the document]
DarshitChanpura: The way resource access authorization works is that it requires ID and type field to be implemented for DocRequest inheritors. DocRequest is how resource authz framework differentiates from regular requests.

Without an id resource authorization will be ignored and normal authorization route will follow:
https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java#L104

src/test/java/org/opensearch/ad/rest/AnomalyDetectorRestApiIT.java
DarshitChanpura: Behind the scenes, security plugin calls an update operation to create/update the resource sharing record. Since tests run much faster consecutively then real world use-case, we force the tests to wait until resource-sharing record is updated.

It is used in all places after create/update AD or forecaster config is successful.

src/main/java/org/opensearch/ad/transport/DeleteAnomalyResultsTransportAction.java
```diff
@@ -70,19 +79,46 @@ public void delete(DeleteByQueryRequest request, ActionListener<BulkByScrollResp
...
             // Case 1: user == null when 1. Security is disabled. 2. When user is super-admin
             // Case 2: If Security is enabled and filter is disabled, proceed with search as
             // user is already authenticated to hit this API.
             client.execute(DeleteByQueryAction.INSTANCE, request, listener);
         } else {
-            // Security is enabled and backend role filter is enabled
             try {
-                ParseUtils.addUserBackendRolesFilter(user, request.getSearchRequest().source());
+                // Security is enabled and resource sharing access control is enabled
+                if (shouldUseResourceAuthz) {
```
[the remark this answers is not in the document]
DarshitChanpura: We have a migrate API already in place:
https://github.com/opensearch-project/security/blob/main/RESOURCE_SHARING_AND_ACCESS_CONTROL.md#3-migration-api-cluster-admins

Once the feature flag is turned on (which requires cluster restart), cluster-admins must run the migration script to migrate any existing info to the new format.

The feature is not intended to be used on dual mode. At present we do not stop storage of user object in ad or forecaster index, that will be done later. 

yes, newer config will not have access to by default except owners, they will have to be shared with for them to have resource access. Owners will have full access by default.

As for migrated resources, all existing backend_roles will be shared with so existing behavior is preserved. Migrate API takes backend_roles_path and default_access_level as input, and this is what the backend_roles will have access at. if cluster-admins want all backend_roles associated with all docs to be granted full access in the new model, they can supply `ad_full_access` or `forecast_full_access` as mentioned in `resource_actions_groups.yml`

src/main/resources/resource-action-groups.yml
```diff
@@ -0,0 +1,31 @@
+# no internal actions are included here
+resource_types:
```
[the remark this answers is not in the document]
DarshitChanpura: sure, this is exactly what I need from AD maintainers. Since I'm not aware of the workflow completely, I would need AD's help to get this file right.

I will update all permissions based on config/roles.

src/test/java/org/opensearch/forecast/rest/SecureForecastRestIT.java
```diff
@@ -1471,6 +1620,309 @@ public void testFilterBy() throws IOException {
...
+                devOpsClient,
+                "GET",
+                String.format(Locale.ROOT, SEARCH_FORECASTER),
+                ImmutableMap.of(),
+                TestHelpers.toHttpEntity(searchForecasterRequest),
+                null
+            );
+        searchResponse = SearchResponse.fromXContent(createParser(JsonXContent.jsonXContent, response.getEntity().getContent()));
+        total = searchResponse.getHits().getTotalHits().value();
+        assertTrue("got: " + total, total == 1);
```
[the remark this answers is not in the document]
DarshitChanpura: sure, i will add.

src/main/java/org/opensearch/timeseries/TimeSeriesResourceSharingExtension.java
```diff
@@ -27,8 +27,8 @@ public class TimeSeriesResourceSharingExtension implements ResourceSharingExtens
     public Set<ResourceProvider> getResourceProviders() {
         return Set
             .of(
```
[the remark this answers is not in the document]
DarshitChanpura: This is not an exhaustive list. I need input from AD maintainers to help decide which indices should have resource-sharing protection. Please let me know which indices and I will add them here.

src/main/java/org/opensearch/timeseries/transport/ResultProcessor.java
DarshitChanpura: https://github.com/opensearch-project/OpenSearch/pull/19430

the original PR that above PR reverted was the cause, but since revert has been merged, i will uncomment this.

DarshitChanpura:
Thanks @kaituo. i've addressed all the comments and left couple of clarifying questions around which indices should be marked as resource-sharing protected, and the permissions in resource-action-groups.yml

src/main/resources/plugin-additional-permissions.yml
DarshitChanpura: There is work in flight to address this: https://github.com/opensearch-project/security/pull/5656. And this file will be updated once we merge.

src/main/java/org/opensearch/timeseries/TimeSeriesResourceSharingExtension.java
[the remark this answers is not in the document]
DarshitChanpura: are we marking all of these resources as protected resources?

src/main/java/org/opensearch/ad/transport/DeleteAnomalyResultsTransportAction.java
```diff
@@ -70,19 +79,46 @@ public void delete(DeleteByQueryRequest request, ActionListener<BulkByScrollResp
...
+                    // verify if new authz should be added on result index, if soo replace the null
+                    addAccessibleConfigsFilterAndDelete(ADIndex.RESULT.getIndexName(), request.getSearchRequest(), listener);
+                    return;
```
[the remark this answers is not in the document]
DarshitChanpura: This is slightly confusing to me. All search requests to protected resources currently have DLS style filtering applied by default. For this specific use-case I had originally used the 2-step approach since we need to filter the DBQ by ids that are accessible. Filter-by-backend works here because a copy of user is stored in the document itself, but since moving all info to security index, we need to fetch the accessible ids first and then apply the filter.

We do store `all_shared_principals` now in the config document. Can we perhaps use that here?
like constructing a terms query to for `all_shared_principals` which will act as filter.
https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java#L414

Thoughts?

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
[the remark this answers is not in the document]
DarshitChanpura: The one from `_id` field and only if detectorId is null.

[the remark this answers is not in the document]
DarshitChanpura: Is there anything specific that your are concerned about?

src/main/java/org/opensearch/timeseries/TimeSeriesResourceSharingExtension.java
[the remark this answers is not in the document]
cwperks: FYI sharable resource indices would be a subset of system indices. Just like the jobs indices from job scheduler. Not every index stores job metadata for a plugin, but jobs indices are system indices. In some cases, a system index could both store job metadata and be a "sharable resource" index simultaneously.

Indices that should be marked "sharable resource" indices would be system indices that contain metadata for dashboards objects that are sharable.

For instance, think of reporting definitions or reporting instances. I can share the report definitions with different users at different levels. i.e. I give permission to Darshit at read level (where he cannot make edits), but then share with Kaituo at write level (allowing Kaituo to make edits). This style access control is not possible with `filter_by_backend_role` where everything is through the cluster administrator and done with `cluster_permissions`. This model of access control empowers the creator of the dashboards resource to specify how they'd like to share (think Quip Share button).

Only include the indices in here that have metadata for sharable resources.

src/main/resources/plugin-additional-permissions.yml
```diff
@@ -0,0 +1,21 @@
+cluster_permissions:
+  - "indices:data/read*"
+  - "indices:data/write/index*"
+  - "cluster:admin/opendistro/ad*"
+  - "cluster:admin/plugin/forecast*"
+index_permissions:
+  - index_patterns:
+      - ".opendistro-anomaly-detectors"
```
cwperks, line 8: System indices a plugin declares do not need to be in `plugin-additional-permissions.yml`. These are additional permissions because they are in _addition_ to system index access by default. The `PluginClient` introduced in this PR let's AD access its own declared system indices without allowing AD to make additional calls. `PluginClient` is a direct replacement to `ThreadContext.stashContext` which allows any plugin to make any system calls without authz and in the long term OpenSearch is moving away from that model towards a model where plugins are better sandboxed. 

Plugins get access to their system indices by default. What belongs in this file is anything additional that it needs.

The example I always use for this is that the security plugin needs permissions to create and write to an opensearch index when using `internal_opensearch` as the audit log destination. In that case, the audit log index is a regular index (not a system index) and security needs additional permissions to be able to create this and write to it using system calls.

DarshitChanpura: ack.. can we clean this up in a follow-up PR?

cwperks: One last thing to add to this thread. This PR added PluginClient and only replaced a one usage of stashContext. In the future, we would want to replace all usages of `ThreadContext.stashContext`, but out of scope of this particular change. The advantage of using PluginClient to make system calls is that it injects information into the ThreadContext about the plugin that is making the system call unlike `ThreadContext.stashContext` which temporarily erases everything and bypasses any sort of authz. See https://github.com/opensearch-project/geospatial/issues/792 for an example of what an error message would look like if a plugin tries to make an unauthorized system call.

In the context of this PR, it was important to introduce it when searching on the resource index because that is when security will automatically apply a filter behind the scenes. From a resource plugins perspective you simply do a search on the resource index (without having to compute the BackendRole filter you do today) and trust that security does the filtration behind-the-scenes and only returns resources _visible_ to the authenticated user.

Visible means 1 of 3 things:

- The authenticated user is the creator
- The resource has been shared with the authenticated user either by username, role or backend role
- The resource is publicly visible

src/main/resources/resource-action-groups.yml
```diff
@@ -5,17 +5,15 @@ resource_types:
     ad_read_only:
       allowed_actions:
         - 'cluster:admin/opendistro/ad/detector/info'
-        - 'cluster:admin/opendistro/ad/detector/search'
         - 'cluster:admin/opendistro/ad/detector/validate'
```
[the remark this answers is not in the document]
DarshitChanpura: search is not for individual resource. Search encompasses multiple resources and must be granted through role.

whether resource is visible in the search result will depend on whether the resource is shared. but the search permission itself must be granted through role mapping.

src/main/resources/resource-action-groups.yml
```diff
@@ -35,17 +34,17 @@ resource_types:
         - 'cluster:admin/plugin/forecast/forecaster/validate'
         - 'cluster:admin/plugin/forecast/forecasters/get'
         - 'cluster:admin/plugin/forecast/forecasters/info'
-        - 'cluster:admin/plugin/forecast/forecasters/search'
```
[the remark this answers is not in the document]
DarshitChanpura: search is not for individual resource. Search encompasses multiple resources and must be granted through role.

whether resource is visible in the search result will depend on whether the resource is shared. but the search permission itself must be granted through role mapping.

kaituo (approved):

PR #1641, opened 2026-01-06
Correlating Anomalies via Temporal Overlap Similarity

kaituo:
### Description
OpenSearch anomalies such as service degradation, job delays, and incident bursts are represented as time intervals, not isolated points. If two detectors fire on the same incident, their anomaly intervals will substantially overlap in time (might with a little timestamp jitter due to different interval, detector start time, and causal relationship). Our similarity therefore measures:

* how much the time windows overlap (after a small tolerance δ to account for jitter),
* optionally, whether the duration is consistent.

This PR implements threshold-graph + connected components based on similarity.

Major algorithm:
- De-dupe input anomalies by id (stable insertion order).
- For every pair (i,j):
  - Dilate both time intervals by ±delta to tolerate bucket alignment drift.
  - Require dilated overlap >= minOverlap (cheap early filter).
  - Compute temporal overlap:
    - IoU (Jaccard over time) on dilated intervals
    - Overlap coefficient (overlap / min(lenA,lenB)) for containment cases
  - Detect strong containment (ovl >= tauContain and duration ratio <= rhoMax).
  - Pick temporal term by mode:
    - IOU: use IoU
    - OVL: use overlap coefficient
    - HYBRID: if strong containment, blend ((1-lam)*IoU + lam*OVL); else use IoU
  - Compute duration penalty exp(-|durA-durB|/kappa).
    - If strong containment, relax the penalty via pow(basePen, containmentRelax) (or disable penalty entirely when containmentRelax == 0).
  - Similarity = temporalTerm * penalty; add an undirected edge if similarity >= alpha.
- Run DFS connected-components on the threshold graph to form clusters.
- Output deterministically: sort members in each cluster by anomaly id.
- Attach an event window per cluster as [min(start), max(end)] across its members.

Testing done:
1. UT
2. Tests on real world data

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List

### DIFF
--- a/src/main/java/org/opensearch/ad/correlation/Anomaly.java
+++ b/src/main/java/org/opensearch/ad/correlation/Anomaly.java
@@ -20,7 +20,7 @@ import java.util.Objects;
  */
 public final class Anomaly {
     // This uniquely identifies the emitting source of the anomaly (e.g., model id).
-    private final String id;
+    private final String modelId;
     // The id of the detector.
     private final String configId;
     // The start time of the anomaly.
@@ -28,8 +28,8 @@ public final class Anomaly {
     // The end time of the anomaly.
     private final Instant dataEndTime;
 
-    public Anomaly(String id, String configId, Instant dataStartTime, Instant dataEndTime) {
-        this.id = Objects.requireNonNull(id, "id");
+    public Anomaly(String modelId, String configId, Instant dataStartTime, Instant dataEndTime) {
+        this.modelId = Objects.requireNonNull(modelId, "modelId");
         this.configId = Objects.requireNonNull(configId, "configId");
         this.dataStartTime = Objects.requireNonNull(dataStartTime, "dataStartTime");
         this.dataEndTime = Objects.requireNonNull(dataEndTime, "dataEndTime");
@@ -39,8 +39,8 @@ public final class Anomaly {
         }
     }
 
-    public String getId() {
-        return id;
+    public String getModelId() {
+        return modelId;
     }
 
     public Instant getDataStartTime() {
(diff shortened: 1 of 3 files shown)


kaituo:
> should we not allow grouping anomalies from the same detector if they overlap or are adjacent?

We don't allow group anomalies from the same entities (e.g., model ids).  But grouping anomalies from the same detector is allowed since we may have high cardinality detectors. Also for single-stream detectors, if anomalies are adjacent, it may make sense to combine them.

kaituo:
> > > should we not allow grouping anomalies from the same detector if they overlap or are adjacent?
> > 
> > 
> > We don't allow group anomalies from the same entities (e.g., model ids). But grouping anomalies from the same detector is allowed since we may have high cardinality detectors. Also for single-stream detectors, if anomalies are adjacent, it may make sense to combine them.
> 
> I mean more like excluding groups that only have one entity in it. Results that only have one entity in it are not presentable on dashboard to customers. Should we stop generating those?

added a parameter to exclude groups that only have one entity in it: https://github.com/opensearch-project/anomaly-detection/compare/82a03a2bf142310a23a8b9a1f662e739e3c35cd5..31ed352793d65e8bbcbe219da48bbe76ddce76f2

kaituo:
> After running the algorithm on different data, I noticed the clustering results tend to be fragmented - multiple clusters with a small number (usually 3 - 5) of entities in it. The time gap in between different clusters are usually 30-60 minutes. Is there any improvement we can done to bridge across short quiet periods?

add more dilation in the start of an anomaly so that if two anomalies don’t overlap (even after dilation), they can still have a chance to be correlated. Please test and let me know.

PR #1715, opened 2026-05-12
Add runtime support for Prometheus and PPL-backed detectors

jackiehanyang:
## Summary
This PR combines the Prometheus direct-query detector work and the PPL runtime detector work into one backend change set.

It adds direct-query detector runtime support for two non-DSL data-source paths:
- Prometheus-backed detectors and preview execution
- PPL-backed detectors that execute PPL at runtime rather than translating detector execution into source DSL

The PPL path is intentionally scoped to detector-friendly single-stream queries. It persists the original `ppl_source` query, derives detector metadata from it at create time, and executes PPL directly for preview/runtime feature fetches.

## What changed
- add Prometheus direct-query detector execution and preview plumbing
- add `source_type = PPL` with persisted `ppl_source`
- add `PPLSource` parsing/compilation for detector-friendly PPL shapes
- add `PPLDirectQueryExecutor` for runtime PPL execution
- support `count()`, `count(*)`, multiple metrics, and pre-`stats` stages like `where`/`eval`
- derive `detection_interval` from `span(...)` when omitted for PPL detectors
- add REST and model test coverage for preview/create/start/profile flows
- fix preview CI regressions for empty sample ranges
- install `opensearch-sql` in integ test clusters so PPL runtime tests execute against the real plugin
- align the local PPL transport request wire format with the upstream SQL plugin
- add secure inline-preview coverage for PPL detectors

## Verification
Ran locally:
- `./gradlew spotlessApply`
- `./gradlew integTest --tests 'org.opensearch.ad.e2e.PreviewMissingSingleFeatureIT'`
- `./gradlew integTest --tests 'org.opensearch.ad.rest.AnomalyDetectorRestApiIT.testPreviewPPLAnomalyDetectorWithoutExplicitDetectionInterval'`
- `./gradlew integTest --tests 'org.opensearch.ad.rest.AnomalyDetectorRestApiIT.testCreateStartAndProfilePPLAnomalyDetectorWithoutExplicitDetectionInterval' --tests 'org.opensearch.ad.rest.AnomalyDetectorRestApiIT.testCreatePPLAnomalyDetectorRejectsUnsupportedPostStatsStage'`
- `./gradlew integTest -Dsecurity=true -Dhttps=true --tests 'org.opensearch.ad.rest.SecureADRestIT.testPreviewInlinePPLAnomalyDetector'`
- `./gradlew integTest -Dsecurity=true -Dhttps=true -Dresource_sharing.enabled=true --tests 'org.opensearch.ad.rest.SecureADRestIT.testPreviewInlinePPLAnomalyDetector'`

## Notes
- This PR intentionally supersedes the scope of #1699 and #1712, while leaving those PRs unchanged.
- The PPL support in this PR is still single-stream only and intentionally limited to detector-friendly query shapes.

### DIFF
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -45,6 +45,7 @@ import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
 import org.opensearch.timeseries.annotation.Generated;
 import org.opensearch.timeseries.common.exception.ValidationException;
+import org.opensearch.timeseries.common.exception.VersionException;
 import org.opensearch.timeseries.constant.CommonMessages;
 import org.opensearch.timeseries.constant.CommonValue;
 import org.opensearch.timeseries.dataprocessor.ImputationOption;
@@ -463,24 +464,26 @@ public class AnomalyDetector extends Config {
             this.frequency = null;
         }
         this.autoCreated = input.readOptionalBoolean();
-        if (input.available() > 0) {
+        if (supportsExternalSourceTransport(input.getVersion())) {
             this.sourceType = normalizeSerializedSourceType(input.readOptionalString());
+            if (this.sourceType == null) {
+                this.sourceType = SOURCE_TYPE_OPENSEARCH;
+            }
+            if (input.readBoolean()) {
+                this.prometheusSource = new PrometheusSource(input);
+                this.sourceType = SOURCE_TYPE_PROMETHEUS;
+            } else {
+                this.prometheusSource = null;
+            }
+            if (input.readBoolean()) {
+                this.pplSource = new PPLSource(input);
+                this.sourceType = SOURCE_TYPE_PPL;
+            } else {
+                this.pplSource = null;
+            }
         } else {
             this.sourceType = SOURCE_TYPE_OPENSEARCH;
-        }
-        if (this.sourceType == null) {
-            this.sourceType = SOURCE_TYPE_OPENSEARCH;
-        }
-        if (input.available() > 0 && input.readBoolean()) {
-            this.prometheusSource = new PrometheusSource(input);
-            this.sourceType = SOURCE_TYPE_PROMETHEUS;
-        } else {
             this.prometheusSource = null;
-        }
-        if (input.available() > 0 && input.readBoolean()) {
-            this.pplSource = new PPLSource(input);
-            this.sourceType = SOURCE_TYPE_PPL;
-        } else {
             this.pplSource = null;
         }
     }
@@ -496,6 +499,19 @@ public class AnomalyDetector extends Config {
      */
     @Override
     public void writeTo(StreamOutput output) throws IOException {
+        if (!supportsExternalSourceTransport(output.getVersion()) && isExternalSourceConfig()) {
+            throw new VersionException(
+                id,
+                String
+                    .format(
+                        Locale.ROOT,
+                        "Cannot serialize [%s] source detector to node before [%s].",
+                        sourceType,
+                        EXTERNAL_SOURCE_TRANSPORT_VERSION
+                    )
+            );
+        }
+
         output.writeOptionalString(id);
         output.writeOptionalLong(version);
         output.writeString(name);
@@ -557,18 +573,20 @@ public class AnomalyDetector extends Config {
             output.writeBoolean(false);
         }
         output.writeOptionalBoolean(autoCreated);
-        output.writeOptionalString(sourceType);
-        if (prometheusSource != null) {
-            output.writeBoolean(true);
-            prometheusSource.writeTo(output);
-        } else {
-            output.writeBoolean(false);
-        }
-        if (pplSource != null) {
-            output.writeBoolean(true);
-            pplSource.writeTo(output);
-        } else {
-            output.writeBoolean(false);
+        if (supportsExternalSourceTransport(output.getVersion())) {
+            output.writeOptionalString(sourceType);
+            if (prometheusSource != null) {
+                output.writeBoolean(true);
+                prometheusSource.writeTo(output);
+            } else {
+                output.writeBoolean(false);
+            }
+            if (pplSource != null) {
+                output.writeBoolean(true);
+                pplSource.writeTo(output);
+            } else {
+                output.writeBoolean(false);
+            }
         }
     }
 
@@ -931,6 +949,14 @@ public class AnomalyDetector extends Config {
         }
         String normalized = sourceType.trim();
         return normalized.isEmpty() ? null : normalized.toUpperCase(Locale.ROOT);
+    }
+
+    private boolean isExternalSourceConfig() {
+        String normalizedSourceType = normalizeSerializedSourceType(sourceType);
+        return SOURCE_TYPE_PROMETHEUS.equals(normalizedSourceType)
+            || SOURCE_TYPE_PPL.equals(normalizedSourceType)
+            || prometheusSource != null
+            || pplSource != null;
     }
 
     public String getDetectorType() {
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -889,11 +889,7 @@ public class AnomalyDetector extends Config {
         return detector;
     }
 
-    private static TimeConfiguration resolveDetectionInterval(
-        TimeConfiguration detectionInterval,
-        String sourceType,
-        PPLSource pplSource
-    ) {
+    private static TimeConfiguration resolveDetectionInterval(TimeConfiguration detectionInterval, String sourceType, PPLSource pplSource) {
         if (detectionInterval != null || !isPPLSourceConfig(sourceType, pplSource)) {
             return detectionInterval;
         }
--- a/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/org/opensearch/ad/model/AnomalyDetector.java
@@ -51,6 +51,7 @@ import org.opensearch.timeseries.model.Config;
 import org.opensearch.timeseries.model.DateRange;
 import org.opensearch.timeseries.model.Feature;
 import org.opensearch.timeseries.model.IntervalTimeConfiguration;
+import org.opensearch.timeseries.model.PrometheusSource;
 import org.opensearch.timeseries.model.ShingleGetter;
 import org.opensearch.timeseries.model.TimeConfiguration;
 import org.opensearch.timeseries.model.ValidationAspect;
@@ -189,6 +190,74 @@ public class AnomalyDetector extends Config {
         TimeConfiguration frequency,
         Boolean autoCreated
     ) {
+        this(
+            detectorId,
+            version,
+            name,
+            description,
+            timeField,
+            indices,
+            features,
+            filterQuery,
+            detectionInterval,
+            windowDelay,
+            shingleSize,
+            uiMetadata,
+            schemaVersion,
+            lastUpdateTime,
+            categoryFields,
+            user,
+            resultIndex,
+            imputationOption,
+            recencyEmphasis,
+            seasonIntervals,
+            historyIntervals,
+            rules,
+            customResultIndexMinSize,
+            customResultIndexMinAge,
+            customResultIndexTTL,
+            flattenResultIndexMapping,
+            lastBreakingUIChangeTime,
+            frequency,
+            autoCreated,
+            null,
+            null
+        );
+    }
+
+    public AnomalyDetector(
+        String detectorId,
+        Long version,
+        String name,
+        String description,
+        String timeField,
+        List<String> indices,
+        List<Feature> features,
+        QueryBuilder filterQuery,
+        TimeConfiguration detectionInterval,
+        TimeConfiguration windowDelay,
+        Integer shingleSize,
+        Map<String, Object> uiMetadata,
+        Integer schemaVersion,
+        Instant lastUpdateTime,
+        List<String> categoryFields,
+        User user,
+        String resultIndex,
+        ImputationOption imputationOption,
+        Integer recencyEmphasis,
+        Integer seasonIntervals,
+        Integer historyIntervals,
+        List<Rule> rules,
+        Integer customResultIndexMinSize,
+        Integer customResultIndexMinAge,
+        Integer customResultIndexTTL,
+        Boolean flattenResultIndexMapping,
+        Instant lastBreakingUIChangeTime,
+        TimeConfiguration frequency,
+        Boolean autoCreated,
+        String sourceType,
+        PrometheusSource prometheusSource
+    ) {
         super(
             detectorId,
             version,
@@ -218,7 +287,9 @@ public class AnomalyDetector extends Config {
             flattenResultIndexMapping,
             lastBreakingUIChangeTime,
             frequency,
-            autoCreated
+            autoCreated,
+            sourceType,
+            prometheusSource
         );
 
         checkAndThrowValidationErrors(ValidationAspect.DETECTOR);
@@ -236,7 +307,6 @@ public class AnomalyDetector extends Config {
             errorMessage = CommonMessages.getTooManyCategoricalFieldErr(maxCategoryFields);
             issueType = ValidationIssueType.CATEGORY;
         }
-
         validateRules(features, rules);
 
         checkAndThrowValidationErrors(ValidationAspect.DETECTOR);
@@ -305,6 +375,17 @@ public class AnomalyDetector extends Config {
             this.frequency = null;
         }
         this.autoCreated = input.readOptionalBoolean();
+        if (input.available() > 0) {
+            this.sourceType = input.readOptionalString();
+        } else {
+            this.sourceType = SOURCE_TYPE_OPENSEARCH;
+        }
+        if (input.available() > 0 && input.readBoolean()) {
+            this.prometheusSource = new PrometheusSource(input);
+            this.sourceType = SOURCE_TYPE_PROMETHEUS;
+        } else {
+            this.prometheusSource = null;
+        }
     }
 
     public XContentBuilder toXContent(XContentBuilder builder) throws IOException {
@@ -379,6 +460,13 @@ public class AnomalyDetector extends Config {
             output.writeBoolean(false);
         }
         output.writeOptionalBoolean(autoCreated);
+        output.writeOptionalString(sourceType);
+        if (prometheusSource != null) {
+            output.writeBoolean(true);
+            prometheusSource.writeTo(output);
+        } else {
+            output.writeBoolean(false);
+        }
     }
 
     @Override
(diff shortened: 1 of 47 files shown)


jackiehanyang:
somehow the CI is not starting and can't complete, checking why...

src/main/java/org/opensearch/forecast/transport/PreviewForecasterTransportAction.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        String forecasterId = request.getForecasterId();
+        if (StringUtils.isBlank(forecasterId)) {
+            listener.onFailure(new OpenSearchStatusException(MISSING_REQUEST_MESSAGE, RestStatus.BAD_REQUEST));
+            return;
+        }
+
+        GetRequest getRequest = new GetRequest(ForecastCommonName.CONFIG_INDEX).id(forecasterId);
+        client.get(getRequest, onGetForecasterResponse(request, listener));
```
kaituo, line 74: with security or resource sharing enabled this could let a user preview another user’s forecaster.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -305,6 +463,26 @@ public AnomalyDetector(StreamInput input) throws IOException {
             this.frequency = null;
         }
         this.autoCreated = input.readOptionalBoolean();
+        if (input.available() > 0) {
```
kaituo: We should use input.readOptionalString() corresponds to output.writeOptionalString(sourceType). available() means bytes left in the whole enclosing stream, not bytes left for this object.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -206,7 +210,145 @@ public AnomalyDetector(
...
+            lastBreakingUIChangeTime,
+            frequency,
+            autoCreated,
+            null,
+            null,
+            null
+        );
+    }
+
+    public AnomalyDetector(
```
kaituo, line 232: Don't create multiple constructors. Use one or two constructor and let callers pass in null. We don't want to create a new constructor when adding a new field.

src/main/java/org/opensearch/ad/model/AnomalyDetector.java
```diff
@@ -218,15 +360,32 @@ public AnomalyDetector(
...
+            autoCreated,
+            sourceType,
+            prometheusSource,
+            pplSource
         );
 
         checkAndThrowValidationErrors(ValidationAspect.DETECTOR);
 
-        if (detectionInterval == null) {
+        if (isPPLSourceConfig(this.sourceType, this.pplSource) && detectionInterval != null) {
```
kaituo, line 372: Do we need similar changes in Forecaster?

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -166,21 +178,296 @@ protected Config(
...
+            lastBreakingUIChangeTime,
+            frequency,
+            autoCreated,
+            sourceType,
+            prometheusSource,
+            null
+        );
+    }
+
+    protected Config(
```
kaituo, line 289: don't add new constructors for new fields. See my comments in AnomalyDetector.

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -77,6 +77,7 @@ public class SecureADRestIT extends AnomalyDetectorRestTestCase {
     private static final String READ_ONLY_AG = "ad_read_only";
     private static final String READ_WRITE_AG = "ad_read_write";
     private static final String FULL_ACCESS_AG = "ad_full_access";
+    private static final String PPL_ACCESS_ROLE = "ppl_access";
```
kaituo, line 80: don't forget to add ppl permission and forecaster preview in security package for read only and full access users and also AD package's resource-access-levels.yml.

src/test/java/org/opensearch/ad/rest/SecureADRestIT.java
```diff
@@ -951,6 +1029,39 @@ public void testPreviewAnomalyDetector() throws IOException {
...
+        Assert.assertEquals("doc_count", featureData.get(0).get("feature_name"));
+        Assert.assertEquals("avg_total_metric", featureData.get(1).get("feature_name"));
+
+        Response catResponse = previewAnomalyDetector(catClient, requestBody, 1);
+        Assert.assertEquals(RestStatus.OK, TestHelpers.restStatus(catResponse));
+
+        Response goatResponse = previewAnomalyDetector(goatClient, requestBody, 1);
+        Assert.assertEquals(RestStatus.OK, TestHelpers.restStatus(goatResponse));
+
+        String noPermsMessage = "no permissions for [cluster:admin/opendistro/ad/detector/preview]";
```
kaituo, line 1057: Can you add tests for the following two scenarios?

* Testing "has AD preview but no PPL"
* Testing "has AD+PPL but no source index read"

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+        ResolvedPrometheusDataSource cached = dataSourceConfigCache.get(dataConnectionId);
+        if (cached != null) {
+            listener.onResponse(cached);
+            return;
+        }
+
+        GetRequest getRequest = new GetRequest(DATASOURCE_INDEX, dataConnectionId);
+        GetResponse response;
+        try (ThreadContext.StoredContext ignored = client.threadPool().getThreadContext().stashContext()) {
+            response = client.get(getRequest).actionGet();
```
kaituo: don't use synchronous call

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+            || credentialValue.trim().isEmpty()
+            || dataSourceEncryptionMasterKey == null
+            || dataSourceEncryptionMasterKey.isEmpty()) {
+            return credentialValue;
+        }
+
+        try {
+            return decryptCredential(credentialValue);
+        } catch (Exception e) {
+            return credentialValue;
```
kaituo, line 520: We should not use encrypted value as credentials. Better return failure.

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+    private static final DateTimeFormatter AWS_DATE_STAMP_FORMATTER = DateTimeFormatter
+        .ofPattern("yyyyMMdd", Locale.ROOT)
+        .withZone(ZoneOffset.UTC);
+
+    private static final int DEFAULT_TIMEOUT_SECONDS = 30;
+    private static final int ERROR_BODY_LIMIT = 512;
+
+    private final Client client;
+    private final HttpClient httpClient;
+    private final ConcurrentHashMap<String, ResolvedPrometheusDataSource> dataSourceConfigCache;
```
kaituo, line 94: User can change credentials or change password. Better have a invalidation on 403.

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+                        .onFailure(
+                            new IllegalStateException(
+                                "Failed to query Prometheus datasource [" + source.getDataConnectionId() + "].",
+                                throwable
+                            )
+                        );
+                    return;
+                }
+
+                if (response.statusCode() < 200 || response.statusCode() >= 300) {
```
kaituo, line 228: When we get 404, the endpoint is gone (user deleted it). Better send EndRunException and stop the detector.

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Executes Prometheus range queries for Prometheus source detectors.
+ *
+ * Instead of depending on SQL direct-query transport classes (which are not visible across
+ * plugin classloaders), this executor resolves datasource metadata from SQL datasource storage
+ * and sends Prometheus query_range HTTP requests directly.
+ */
+public class PrometheusDirectQueryExecutor {
```
lezzago: SQL plugin has direct query apis that can call prometheus. It would be better to use that especially if we are depending on the SQL plugin. Though I believe there is some work to be done to expose the SQL plugin direct query apis. 
ref: https://github.com/opensearch-project/sql/tree/main/direct-query/src/main/java/org/opensearch/sql/directquery/transport

src/main/java/org/opensearch/forecast/rest/RestPreviewForecasterAction.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        ForecasterExecutionInput input;
+        try {
+            input = getForecasterExecutionInput(request);
+        } catch (IOException e) {
+            return channel -> channel.sendResponse(new BytesRestResponse(RestStatus.BAD_REQUEST, e.getMessage()));
+        }
+
+        String error = validateForecasterExecutionInput(input);
```
kaituo, line 72: Can we use currently lacks the min_preview_size parameter similar to preview detector? This number would be at least 40. The higher the better.

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -699,6 +1040,18 @@ public List<String> getIndices() {
         return indices;
     }
 
+    public String getSourceType() {
```
kaituo, line 1043: did you add new fields to toString()?

src/main/java/org/opensearch/timeseries/model/Config.java
```diff
@@ -361,6 +645,38 @@ protected Config(
...
         this.frequency = frequency;
         this.autoCreated = autoCreated != null ? autoCreated : false;
+        this.sourceType = normalizedSourceType;
+        this.prometheusSource = prometheusSource;
+        this.pplSource = pplSource;
+    }
+
+    private String normalizeSourceType(String sourceType, PrometheusSource prometheusSource, PPLSource pplSource) {
+        if (Strings.isBlank(sourceType)) {
+            if (prometheusSource != null) {
```
kaituo: We should return error when both prometheusSource and pplSource are not null.

src/main/java/org/opensearch/timeseries/feature/PPLDirectQueryExecutor.java
```diff
@@ -0,0 +1,345 @@
...
+            this.queryId = null;
+        }
+
+        @Override
+        public ActionRequestValidationException validate() {
+            return null;
+        }
+
+        @Override
+        public void writeTo(StreamOutput out) throws IOException {
```
kaituo, line 293: Where is your readFrom method?

src/main/java/org/opensearch/timeseries/feature/PPLDirectQueryExecutor.java
```diff
@@ -0,0 +1,345 @@
...
+            }
+            if (valueNode.isNumber()) {
+                values[i] = valueNode.doubleValue();
+                continue;
+            }
+            if (valueNode.isTextual() && valueNode.textValue() != null && !valueNode.textValue().isBlank()) {
+                values[i] = Double.parseDouble(valueNode.textValue());
+                continue;
+            }
+            return Optional.empty();
```
kaituo, line 209: Should we throw IllegalStateException as we did in line 223 or at least log a warning?

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+                    null,
+                    new SecretKeySpec(dataSourceEncryptionMasterKey.getBytes(StandardCharsets.UTF_8), "AES"),
+                    "Custom",
+                    "opensearch.config.master.key",
+                    "AES/GCM/NoPadding"
+                );
+            Method decryptDataMethod = null;
+            for (Method method : crypto.getClass().getMethods()) {
+                if ("decryptData".equals(method.getName()) && method.getParameterCount() == 2) {
+                    decryptDataMethod = method;
```
kaituo, line 549: this method is recreated in every decryption. Should be created once and reused later.

src/main/java/org/opensearch/forecast/ForecasterRunner.java
```diff
@@ -0,0 +1,116 @@
...
+import org.opensearch.forecast.model.Forecaster;
+import org.opensearch.timeseries.AnalysisType;
+import org.opensearch.timeseries.feature.FeatureManager;
+import org.opensearch.timeseries.feature.Features;
+import org.opensearch.timeseries.ml.Sample;
+
+/**
+ * Runner to trigger a forecasting preview request.
+ */
+public final class ForecasterRunner {
```
kaituo, line 37: It looks like the preview logic in ForecasterRunner currently duplicates an older version of AnomalyDetectorRunner line-for-line. Because of this duplication, the Forecaster preview is missing two features that AD preview already supports:

* User Propagation: executeForecaster does not accept or restore ThreadContext.StoredContext (like AD does on line 86). Without restoring the user context in the background thread, internal queries may fail due to security plugin permission errors.
* HCAD: Currently, ForecasterRunner immediately rejects requests with categoryFields. The AD runner actively supports this by fetching preview entities and fetching features specifically for each entity.

Instead of maintaining two separate but highly identical preview runners, we should consider refactoring the shared logic into an AbstractPreviewRunner base class. This would give the forecast preview full feature parity with AD for free and eliminate duplicated code.

src/main/java/org/opensearch/forecast/rest/RestPreviewForecasterAction.java
```diff
@@ -0,0 +1,105 @@
...
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+        if (!ForecastEnabledSetting.isForecastEnabled()) {
+            throw new IllegalStateException(ForecastCommonMessages.DISABLED_ERR_MSG);
+        }
+
+        ForecasterExecutionInput input;
+        try {
+            input = getForecasterExecutionInput(request);
+        } catch (IOException e) {
+            return channel -> channel.sendResponse(new BytesRestResponse(RestStatus.BAD_REQUEST, e.getMessage()));
```
kaituo, line 69: We should throw exception instead of returning raw bytes to the caller. They are expecting a json response not a string.

src/main/java/org/opensearch/forecast/transport/PreviewForecasterTransportAction.java
kaituo, line 74: Do we need to use resolveUserAndExecute()? This means that if User B somehow guesses or knows the forecasterId of a forecaster created by User A, User B can run a preview against it. This circumvents the resource-sharing and backend-role restrictions that are supposed to isolate tenants and users from each other.

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1005 @@
...
+
+        ResolvedPrometheusDataSource cached = dataSourceConfigCache.get(dataConnectionId);
+        if (cached != null) {
+            listener.onResponse(cached);
+            return;
+        }
+
+        GetRequest getRequest = new GetRequest(DATASOURCE_INDEX, dataConnectionId);
+        GetResponse response;
+        try (ThreadContext.StoredContext ignored = client.threadPool().getThreadContext().stashContext()) {
```
kaituo, line 305: You need to do RBAC check by checking if current user's roles are one of the datasource's allowedRoles.

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
jackiehanyang: Agree this is the better long term direction. For this PR I’m keeping direct HTTP because SQL direct query APIs are not yet exposed/stable enough for AD to depend on. Also given the timeline, once SQL exposes a stable transport/client API, we can replace this executor without changing the AD detector API.

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1036 @@
...
+        for (Map.Entry<Long, double[]> entry : timestampAccumulator.entrySet()) {
+            double[] sumAndCount = entry.getValue();
+            if (sumAndCount[1] > 0d) {
+                averaged.put(entry.getKey(), sumAndCount[0] / sumAndCount[1]);
+            }
+        }
+        return averaged;
+    }
+
+    Map<Map<String, String>, NavigableMap<Long, Double>> parsePrometheusResponseBySeries(String body, Map<String, String> seriesFilter)
```
kaituo, line 732: Is it possible users have:

feature: rate(http_requests_total[5m])
category field = host

this method returns:

{
  {
    "__name__": "http_requests_total",
    "host": "serverA",
    "method": "GET",
    "status": "200"
  } -> {
    1710000000000L -> 50.0
  },

  {
    "__name__": "http_requests_total",
    "host": "serverA",
    "method": "POST",
    "status": "500"
  } -> {
    1710000000000L -> 10.0
  },

  {
    "__name__": "http_requests_total",
    "host": "serverB",
    "method": "GET",
    "status": "200"
  } -> {
    1710000000000L -> 100.0
  }
}


Thus, we have

{host="serverA", method="GET",  status="200"}  -> 50
{host="serverA", method="POST", status="500"} -> 10

Both map to the same AD entity:

Entity{host=serverA}

Then CompositeRetriever.analyzePrometheusSeries() builds an AD entity using only configured categorical fields like host

then we have

{host="serverA"} -> 10

because CompositeRetriever.analyzePrometheusSeries uses 

results.put(entity.get(), new double[] { value.get() }); 

to overwrite the same entity's value.

Is this possible? Would promql aggregate values to the detector’s categorical fields?

src/main/java/org/opensearch/timeseries/feature/PrometheusDirectQueryExecutor.java
```diff
@@ -0,0 +1,1036 @@
...
+                        .onFailure(
+                            new IllegalStateException(
+                                "Prometheus query_range failed with HTTP " + response.statusCode() + ". Body: " + truncate(response.body())
+                            )
+                        );
+                    return;
+                }
+
+                try {
+                    listener.onResponse(responseParser.parse(response.body()));
```
kaituo, line 239: Do you want to paginate the query? A raw Prometheus query like rate(http_requests_total[5m]) can produce a huge number of series.

kaituo (commented):
Reviewed until https://github.com/opensearch-project/anomaly-detection/pull/1715/commits/0f6f80f6c3df64ded11f55998dd75628df64ca0f

PR #1775, opened 2026-09-11
Advance to OpenSearch 3.9 and fix build and integration compatibility

jackiehanyang:
Advances the plugin to OpenSearch 3.9.0-SNAPSHOT and fixes the compatibility failures exposed by that upgrade. This replaces #1760 because the automated version branch overwrites follow-up fixes when it refreshes.

- Restore the Jackson 2 core dependency required by PPL response parsing and Random Cut Forest serialization alongside OpenSearch’s Jackson 3 dependencies.
- Match SQL 3.9’s PPL request wire format, including the analyze and partial-result fields. Remove the unused inbound request adapter and verify the outbound bytes directly.
- Close both regular and certificate-authenticated admin HTTPS connection pools before the test framework shuts down HTTP reactors, preventing selector assertions and leaked threads.
- Exercise historical PPL analysis through the real SQL plugin and assert the returned feature names and values. This covers 23 previously missed executor lines, including date-range and grouped-query parsing.
- Give the null/empty-feature job tests a valid timestamp mapping so they reach the intended feature validation.
- Handle an unmapped detector-type aggregation in the stats API without returning HTTP 500. A regression test exercises both an empty config index and a document without a detector type.

Validation: the full `./gradlew build` passed locally on Java 21, including the complete integration and unit suites, packaging, Spotless, and coverage verification. The unmapped detector-type regression first reproduced CI's exact aggregation cast error; all five stats transport tests pass with the fix. The stats and historical PPL REST tests also passed over HTTPS with resource sharing enabled using CI's failing seed, locale, and time zone. Earlier focused validation covered both HTTPS configurations and historical PPL execution on a three-node Java 25 cluster. The existing CI matrix and coverage thresholds are unchanged.

Supersedes #1760.

Dependency and TLS review evidence:

- The resolved artifact is `com.fasterxml.jackson.core:jackson-core:2.22.2`, published by FasterXML. Its cached JAR matches the checksum published by Maven Central. The version comes from OpenSearch’s existing `versions.jackson` setting; Jackson 2 databind, annotations, and Random Cut Forest serialization already require this API.
- The admin HTTPS setup uses the same `TrustStore` certificate and client keystore as `SecureRestClientBuilder`, retains the default hostname verifier and TLS protocol/cipher settings, and preserves its connection/request timeouts. Its connection manager is retained so the test fixture can close active TLS connections before reactor shutdown.
- No test, coverage threshold, or security assertion has been disabled.

### DIFF
--- a/src/main/java/org/opensearch/timeseries/feature/PPLDirectQueryExecutor.java
+++ b/src/main/java/org/opensearch/timeseries/feature/PPLDirectQueryExecutor.java
@@ -342,6 +342,7 @@ public class PPLDirectQueryExecutor {
         COMPACT
     }
 
+    // Keep the wire format aligned with SQL's TransportPPLQueryRequest.
     private static final class PPLTransportRequest extends ActionRequest {
         private final String query;
         private final String format;
@@ -350,7 +351,9 @@ public class PPLDirectQueryExecutor {
         private final String path;
         private final boolean sanitize;
         private final boolean profile;
+        private final boolean analyze;
         private final String queryId;
+        private final Boolean partialResult;
 
         private PPLTransportRequest(String query, String format, String path) {
             this.query = query;
@@ -360,7 +363,9 @@ public class PPLDirectQueryExecutor {
             this.path = path;
             this.sanitize = true;
             this.profile = false;
+            this.analyze = false;
             this.queryId = null;
+            this.partialResult = null;
         }
 
         private PPLTransportRequest(StreamInput in) throws IOException {
@@ -373,7 +378,9 @@ public class PPLDirectQueryExecutor {
             this.sanitize = in.readBoolean();
             in.readEnum(PPLJsonStyle.class);
             this.profile = in.readBoolean();
+            this.analyze = in.readBoolean();
             this.queryId = in.readOptionalString();
+            this.partialResult = in.readOptionalBoolean();
         }
 
         @Override
@@ -392,7 +399,9 @@ public class PPLDirectQueryExecutor {
             out.writeBoolean(sanitize);
             out.writeEnum(PPLJsonStyle.COMPACT);
             out.writeBoolean(profile);
+            out.writeBoolean(analyze);
             out.writeOptionalString(queryId);
+            out.writeOptionalBoolean(partialResult);
         }
 
     }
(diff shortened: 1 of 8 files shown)


reta:
@jackiehanyang could you please incorporate https://github.com/opensearch-project/anomaly-detection/pull/1776/changes#diff-7414e9228e5f40d1fd13be3a0c75ac7a12e7ed7095a79ca650a2423cb1d03c0a ? thank you

jackiehanyang:
> @jackiehanyang could you please incorporate https://github.com/opensearch-project/anomaly-detection/pull/1776/changes#diff-7414e9228e5f40d1fd13be3a0c75ac7a12e7ed7095a79ca650a2423cb1d03c0a ? thank you

@reta thanks for the change and looks like we're doing the same thing here :) but #1776 CI failed, I think some test classes need to be tweaked a little, which is what I'm doing in this pr. I can take it from here, will make sure the version bump change can be merged in by end of day today

gaiksaya:
Thank you @jackiehanyang @reta

src/main/java/org/opensearch/timeseries/feature/PPLDirectQueryExecutor.java
```diff
@@ -360,20 +364,9 @@ private PPLTransportRequest(String query, String format, String path) {
             this.path = path;
             this.sanitize = true;
             this.profile = false;
+            this.analyze = false;
             this.queryId = null;
-        }
-
-        private PPLTransportRequest(StreamInput in) throws IOException {
```
kaituo, line 366: why do you remove this?

jackiehanyang: I removed it because `PPLTransportRequest` is an outbound only adapter: AD constructs it and calls `writeTo()`, while SQL deserializes the bytes using its own `TransportPPLQueryRequest(StreamInput)` constructor. The removed constructor had no production callers， only the unit test used it.

I replaced that test with explicit assertions on the serialized field order, including `analyze` and `partialResult`. This checks the expected SQL wire format directly and avoids a round-trip test passing when our reader and writer share the same mistake.

The removal is cleanup, adding those missing fields to `writeTo()` is the actual compatibility fix.

kaituo (approved):

PR #1782, opened 2026-09-19
Prevent historical HCAD analyses from remaining stuck in RUNNING

kaituo:
### Description

Prevent historical high-cardinality anomaly detection (HCAD) analyses from remaining indefinitely in `RUNNING` after task-slot scale-down, and fix the nullable rule serialization failure that triggered the reported incident.

### RCA summary

The incident required two defects:

1. **Remote rule serialization failed.** A valid `ACTUAL_IS_OVER_EXPECTED` or `ACTUAL_IS_BELOW_EXPECTED` condition has no operator or value. `Condition.writeTo` nevertheless called `writeEnum(operator)`, which dereferenced the null operator while forwarding an entity task to another node.
2. **The last execution lane was abandoned.** After an entity failed and the number of unfinished entities decreased, the coordinator released idle task slots. The old `runNextEntityForHCADHistorical` returned unconditionally whenever `scaleDelta < 0`, even when pending entities remained and no other entity was active.

The first defect triggered entity dispatch failures. The second defect converted those failures into a permanently stuck parent task.

### Slots, lanes, and entity states

A historical HCAD analysis has one parent task and one child task for each category-field value or combination.

The coordinator tracks child entities as:

- **Pending:** waiting to be dispatched.
- **Temp:** removed from pending and being dispatched, but not yet accepted by a worker.
- **Running:** accepted by a worker and executing.

`activeEntityCount` is `temp + running`.

A **task slot** is reserved concurrency capacity. Four assigned slots mean the detector may run up to four entity tasks; slots do not start work themselves.

An **execution lane** is the callback chain that keeps polling work. When one entity completes or fails, the lane can dispatch the next pending entity. A lane is control flow, not a persisted task or thread.

The important distinction is that a detector can have available slots but no lane left to use them. This is equivalent to having open checkout stations but no cashier serving the waiting customers.

### What `scaleDelta` means

The coordinator calculates:

```text
desired limit = min(
    unfinished entities,
    cluster batch-task capacity,
    per-detector running-entity limit
)

scaleDelta = desired limit - assigned task slots
```

- Positive: reserve more slots.
- Zero: retain the current slots.
- Negative: release idle slots because fewer are now needed.

Releasing idle slots does not necessarily mean the current lane should stop. If pending work exists and `activeEntityCount < assignedTaskSlots` after scale-down, the remaining slots still have unused capacity and the current lane must continue.

### Failure sequence

1. The parent task discovered its HCAD entities.
2. Some local child entities completed normally.
3. Remote child dispatch attempted to serialize a valid absolute-threshold rule whose operator was null.
4. `writeEnum(null)` threw before the remote worker accepted the task.
5. The failed child was removed from temp/running tracking and its callback attempted to advance the lane.
6. Because fewer unfinished entities remained, `scaleDelta` became negative and idle slots were released.
7. The old code stopped the current lane solely because `scaleDelta < 0`.
8. Eventually the coordinator reached a state such as:

   ```text
   pending = 4
   temp = 0
   running = 0
   assigned slots = 4
   ```

   Capacity remained, but no callback lane existed to poll another entity.
9. The parent finalization path requires no pending, temp, or running entities. Because the pending queue remained nonempty, progress stayed below `1.0` and the persisted state remained `RUNNING`.

In the reported reproduction, 4 of 280 entities remained pending:

```text
progress = 1 - 4 / 280 = 98.57%
```

The UI may show a nearby rounded or previously sampled value such as 97%, but it cannot reach 100% while pending entities remain. The stale `RUNNING` state also prevents normal detector deletion.

#### Preserve scheduler liveness after scale-down

After scaling down, `runNextEntityForHCADHistorical` now stops the current lane only when:

```text
pendingEntityCount == 0
    OR
activeEntityCount >= assignedTaskSlots
```

Otherwise, pending work exists and at least one remaining slot is unused, so the current lane dispatches another entity.

For the stuck tail state:

```text
pending = 4
active = 0
slots = 4
```

the lane continues because `pending > 0` and `active < slots`.

#### Serialize nullable rule operators

`Condition` transport serialization now writes an operator-presence boolean:

```text
operator present = false  -> operator is null
operator present = true   -> enum value follows
```

The reader consumes the matching presence-bit format. Historical task requests are routed through version-specific hash rings, so coordinator and worker nodes run the same AD version and use the same wire layout. An old plugin version therefore continues to read and write its old layout, while the new plugin version reads and writes the nullable layout.

### Test coverage

Focused serialization tests verify:

- A null operator round-trips on the current transport version.
- A non-null operator round-trips on the current version.

Focused task-manager tests verify:

- Negative `scaleDelta`, pending entities, zero active entities, and available assigned slots cause the current lane to continue.
- The current lane stops when all remaining slots are occupied.
- The current lane stops when no pending entities remain.

The three-node REST IT creates nine HCAD entities and a valid absolute-threshold rule with a null operator. It verifies that the rule crosses the remote transport boundary, task-slot scale-down occurs, and the parent reaches a terminal state with progress `1.0` and no pending or running entities.

Once nullable serialization is fixed, successful remote dispatch activates the other entity lanes, so the REST IT no longer creates the exact zero-active-lane state. The focused task-manager test deterministically reproduces that scheduler state.

### Local verification

```bash
./gradlew spotlessJavaCheck --console=plain

./gradlew test --console=plain

./gradlew integTest -PnumNodes=3 \
  --tests org.opensearch.ad.rest.HistoricalAnalysisRestApiIT.testScaleDownWithNullableRuleCompletes \
  --console=plain
```

Results:

- Full unit suite passed.
- `ConditionTests`: 2 tests, 0 skipped, 0 failures.
- `ADTaskManagerTests`: 51 tests, 0 skipped, 0 failures.
- Three-node REST IT: 1 test, 0 skipped, 0 failures.
- Spotless checks passed.

### Check List

- [x] Bug fix includes focused unit tests.
- [x] Bug fix includes a three-node REST integration test.
- [x] Nullable transport encoding follows same-version historical task routing.
- [x] Changelog entry added.
- [x] No public API change.
- [x] Commit is signed off per the DCO.

For more information on the Developer Certificate of Origin, see [CONTRIBUTING.md](https://github.com/opensearch-project/anomaly-detection/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


jackiehanyang:
thanks for the change, lgtm

jackiehanyang (approved):

